PR #12792, opened 2020-01-15
Fix: id-blacklist should ignore ObjectPatterns (fixes #12787)

### LINKED ISSUE #12787: Rule id-blacklist incorrectly flags object property reassignments

* **ESLint Version:** 6.8.0
* **Node Version:** 10.15.3
* **npm Version:** 6.11.3

babel-eslint

<details>

```yaml
extends: airbnb
parser: babel-eslint
rules:
  space-before-function-paren: 2
  brace-style: [2, 'stroustrup']
  arrow-body-style: 0
  no-underscore-dangle: 0
  class-methods-use-this: 1
  consistent-return: 0
  no-unused-vars: 1
  no-bitwise: 0
  no-plusplus: 0
  linebreak-style: 0
  arrow-parens: [2, 'always']
  no-useless-escape: 0
  curly: [2, 'all']
  no-lonely-if: 1
  semi: 2
  max-len: [1, { code: 120, ignoreComments: true }]
  prefer-destructuring: 1
  lines-between-class-members: 0
  object-curly-newline: 0
  indent: 1
  jsx-a11y/anchor-is-valid: [ 'error', {
    components: [ 'Link' ],
    specialLink: [ 'to' ]
  }]
  jsx-a11y/click-events-have-key-events: 0
  jsx-a11y/no-static-element-interactions: 0
  jsx-a11y/no-noninteractive-tabindex: 0
  react/prefer-stateless-function: 1
  react/no-unused-state: 1
  react/state-in-constructor: [1, "never"]
  react/jsx-one-expression-per-line: 0
  react/prop-types: 0 # This should be enabled at some point
  react/jsx-indent: 1
  import/no-extraneous-dependencies: ["error", {"devDependencies": ["test/**/*.js"]}]
  jsx-a11y/label-has-associated-control: [ 2, {
    depth: 1,
  }]
  react/forbid-component-props: ["error", {"forbid": ["style", "className"]}]
  react/forbid-dom-props: ["error", {"forbid": ["style", "className"]}]
  react/jsx-sort-props: ["error", {"callbacksLast": true, "reservedFirst": true, "shorthandFirst": true, "noSortAlphabetically": true}]
  id-blacklist: [1, "data", "err", "e", "cb", "callback"]
env:
  browser: true
  node: true
  es6: true
```

</details>

Highlight:
```yaml
id-blacklist: [1, "data", "err", "e", "cb", "callback"]
```

```js
const { data: user } = await axios.post('/api/auth', loginCredentials);
if (!user) {
  throw new Error('Error fetching authentication data');
}
```

```bash
npx eslint --ext .js --ext .jsx src/
```

The `data` property shouldn't be flagged here because it's coming _from_ the axios request and there's nothing I can do about it. I'm converting the value `response.data` to the variable `user`. However it highlights it as being as if I were assigning the data variable.

```
45:15  warning  Identifier 'data' is blacklisted  id-blacklist
```

I would but I'm not sure how as I've never delved into the eslint source!

jpramassini:

[ ] Documentation update
[ ] Add autofixing to a rule
[ ] Add a CLI option
[ ] Add something to the core
[ ] Other, please explain:

https://github.com/eslint/eslint/issues/12787

I modified the should report logic to check that the `effectiveParent`'s `parent` is a block statement and added tests for the case.

Thoughts on any potentially cleaner ways to accomplish this? This is my first PR to ESLint so I want to make sure it's up to snuff.

### DIFF
--- a/lib/rules/id-blacklist.js
+++ b/lib/rules/id-blacklist.js
@@ -101,7 +101,7 @@ module.exports = {
                             report(node);
                         }
 
-                        // Report AssignmentExpressions only if they are the left side of the assignment
+                    // Report AssignmentExpressions only if they are the left side of the assignment
                     } else if (effectiveParent.type === "AssignmentExpression" &&
                         (effectiveParent.right.type !== "MemberExpression" ||
                             effectiveParent.left.type === "MemberExpression" &&
@@ -111,7 +111,7 @@ module.exports = {
                         }
                     }
 
-                    // Properties have their own rules
+                // Properties have their own rules
                 } else if (node.parent.type === "Property") {
 
                     if (shouldReport(effectiveParent, name)) {
--- a/lib/rules/id-blacklist.js
+++ b/lib/rules/id-blacklist.js
@@ -88,8 +88,6 @@ module.exports = {
         return {
 
             Identifier(node) {
-
-                // if (node.name === 'food') console.log(node)
                 const name = node.name,
                     effectiveParent = (node.parent.type === "MemberExpression") ? node.parent.parent : node.parent;
 
--- a/lib/rules/id-blacklist.js
+++ b/lib/rules/id-blacklist.js
@@ -61,9 +61,12 @@ module.exports = {
          * @returns {boolean} whether an error should be reported or not
          */
         function shouldReport(effectiveParent, name) {
-            return effectiveParent.type !== "CallExpression" &&
+            return (
+                effectiveParent.type !== "CallExpression" &&
                 effectiveParent.type !== "NewExpression" &&
-                isInvalid(name);
+                effectiveParent.parent.type !== "BlockStatement" &&
+                isInvalid(name)
+            );
         }
 
         /**
@@ -101,21 +104,21 @@ module.exports = {
                         // Report AssignmentExpressions only if they are the left side of the assignment
                     } else if (effectiveParent.type === "AssignmentExpression" &&
                         (effectiveParent.right.type !== "MemberExpression" ||
-                        effectiveParent.left.type === "MemberExpression" &&
-                        effectiveParent.left.property.name === node.name)) {
+                         effectiveParent.left.type === "MemberExpression" &&
+                         effectiveParent.left.property.name === node.name)) {
                         if (isInvalid(name)) {
                             report(node);
                         }
                     }
 
-                // Properties have their own rules
+                    // Properties have their own rules
                 } else if (node.parent.type === "Property") {
 
                     if (shouldReport(effectiveParent, name)) {
                         report(node);
                     }
 
-                // Report anything that is a match and not a CallExpression
+                    // Report anything that is a match and not a CallExpression
                 } else if (shouldReport(effectiveParent, name)) {
                     report(node);
                 }
--- a/lib/rules/id-blacklist.js
+++ b/lib/rules/id-blacklist.js
@@ -64,7 +64,7 @@ module.exports = {
             return (
                 effectiveParent.type !== "CallExpression" &&
                 effectiveParent.type !== "NewExpression" &&
-                effectiveParent.parent.type !== "BlockStatement" &&
+                effectiveParent.parent.type !== "ObjectPattern" &&
                 isInvalid(name)
             );
         }
@@ -88,6 +88,8 @@ module.exports = {
         return {
 
             Identifier(node) {
+
+                // if (node.name === 'food') console.log(node)
                 const name = node.name,
                     effectiveParent = (node.parent.type === "MemberExpression") ? node.parent.parent : node.parent;
 
@@ -104,8 +106,8 @@ module.exports = {
                         // Report AssignmentExpressions only if they are the left side of the assignment
                     } else if (effectiveParent.type === "AssignmentExpression" &&
                         (effectiveParent.right.type !== "MemberExpression" ||
-                         effectiveParent.left.type === "MemberExpression" &&
-                         effectiveParent.left.property.name === node.name)) {
+                            effectiveParent.left.type === "MemberExpression" &&
+                            effectiveParent.left.property.name === node.name)) {
                         if (isInvalid(name)) {
                             report(node);
                         }
--- a/tests/lib/rules/id-blacklist.js
+++ b/tests/lib/rules/id-blacklist.js
@@ -66,8 +66,9 @@ ruleTester.run("id-blacklist", rule, {
             options: ["f", "fo", "fooo", "b", "ba", "barr", "bazz", "bingg"]
         },
         {
-            code: "{foo: bar}",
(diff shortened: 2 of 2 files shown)


yeonjuan:
@jpramassini 
Hello.
I'm not an eslint member but leave some comments.
Issue #12787 is related to the [Destructuring assignment](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment), while the codes in your test cases are [Labeled statement](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/label).
https://github.com/eslint/eslint/blob/ec3983c9d1fbc65185065d235f627a43d8cf9a1e/tests/lib/rules/id-blacklist.js#L69

jpramassini:
@yeonjuan This is a very good point, dumb mistake on my part. I'll update my test cases later today.

kaicataldo:
I also want to note that the fix should be to not warn on destructured assignments at all. From the [rule docs](https://eslint.org/docs/rules/id-blacklist#rule-details) (emphasis my own):

> It will not catch blacklisted identifiers that are:
>
> * function calls (so you can still use functions you do not have control over)
> * **object properties (so you can still use objects you do not have control over)**

Ideally, I do think the rule should warn when a property is renamed to a blacklisted identifier name in destructuring assignment, but I also think that can (and probably should) be done in a separate PR as a separate bug fix.

jpramassini:
@kaicataldo Thanks for the help on the AST! I'll take another crack at this.

jpramassini:
@kaicataldo Updated to ignore in Object Pattern nodes, and removed the error test case as we want to ignore it for this PR.

kaicataldo:
You got it! Feel free to reach out with any questions you have :)

lib/rules/id-blacklist.js
```diff
@@ -85,6 +88,8 @@ module.exports = {
         return {
 
             Identifier(node) {
+
+                // if (node.name === 'food') console.log(node)
```
kaicataldo: Looks like this might have been left in accidentally 😄

tests/lib/rules/id-blacklist.js
```diff
@@ -65,6 +65,11 @@ ruleTester.run("id-blacklist", rule, {
             code: "var obj = { key: foo.bar };",
             options: ["f", "fo", "fooo", "b", "ba", "barr", "bazz", "bingg"]
         },
+        {
+            code: "const {foo: bar} = baz",
```
kaicataldo: Can we also add tests for nested destructuring assignments? Example:

```js
const { foo: { bar: baz } } = baz;
```

kaicataldo: Would also be good to add tests for destructured function parameters (including nested `ObjectPattern`s):

```js
function foo({ bar: baz }) {}
```

kaicataldo:
We're getting closer! Thanks for working on this.

jpramassini:
Do we want a failing test case for `function foo({ bar: baz }) {}`? I.e. testing to make sure we're still getting the function name if `foo` is illegal?

kaicataldo:
Sorry, that wasn't a great example. I think a test for the destructured parameters is sufficient. In the example above, that would be having `bar` blacklisted and not warning.

kaicataldo:
That being said, additional tests are always welcome :)

jpramassini:
I think between the existing function name tests and the added nested and destructured arguments we should be covered, so I left it out.

kaicataldo:
One last thing from me: do you mind updating the PR title to more accurately reflect the current iteration of this PR?

lib/rules/id-blacklist.js
```diff
@@ -101,21 +104,21 @@ module.exports = {
...
-                        effectiveParent.left.property.name === node.name)) {
+                            effectiveParent.left.type === "MemberExpression" &&
+                            effectiveParent.left.property.name === node.name)) {
                         if (isInvalid(name)) {
                             report(node);
                         }
                     }
 
-                // Properties have their own rules
+                    // Properties have their own rules
```
kaicataldo: Mind undoing this? This is referring to the follow block and I think it's clearer to have it indented as it was.

lib/rules/id-blacklist.js
```diff
@@ -101,21 +104,21 @@ module.exports = {
...
                 } else if (node.parent.type === "Property") {
 
                     if (shouldReport(effectiveParent, name)) {
                         report(node);
                     }
 
-                // Report anything that is a match and not a CallExpression
+                    // Report anything that is a match and not a CallExpression
```
kaicataldo: Mind undoing this? This is referring to the follow block and I think it's clearer to have it indented as it was.

kaicataldo:
Re the commit message check that's failing now: How about changing it to `Fix: id-blacklist should ignore ObjectPatterns (fixes #12787)`?

kaicataldo:
You're doing great. We really appreciate the contribution!

lib/rules/id-blacklist.js
kaicataldo: Yeah, I see that. Strange :/ Seems like something on GitHub's end?

jpramassini: Seems like it. This is the first time I've had this happen. 🤷‍♂

jpramassini:
Awesome, thanks again for all of your help!

tests/lib/rules/id-blacklist.js
```diff
@@ -65,6 +65,26 @@ ruleTester.run("id-blacklist", rule, {
...
+            options: ["foo", "bar"],
```
yeonjuan: @jpramassini 
Hi, In my opinion, this case should be included in `invalid` because it makes a new identifier `bar,` which is in the id-blacklist.

jpramassini: Yeah, this should definitely be invalid at some point due to the spirit of the rule. I added that to try and illustrate for now that that's intended behavior for now. Should I make a separate issue to make sure that gets tracked?

yeonjuan: > Should I make a separate issue to make sure that gets tracked?    

Maybe it would be welcomed, but it is up to you. :)

kaicataldo: An issue would be great!

jpramassini:
@kaicataldo Unsure of where to put this, this seems like the best place to follow up. For the issue to track the reassignment identifier warning for this rule, should that issue be categorized as a bug fix, or is it a rule change proposal?

kaicataldo:
I think it should be categorized as a semver-minor bug fix (reported as a bug fix).

jpramassini:
@kaicataldo New issue here: https://github.com/eslint/eslint/issues/12807

PR #12316, opened 2019-09-26
Fix: preserve whitespace in multiline-comment-style (fixes #12312)

kaicataldo:
**What is the purpose of this pull request? (put an "X" next to item)**

[ ] Documentation update
[x] Bug fix ([template](https://raw.githubusercontent.com/eslint/eslint/master/templates/bug-report.md))
[ ] New rule ([template](https://raw.githubusercontent.com/eslint/eslint/master/templates/rule-proposal.md))
[ ] Changes an existing rule ([template](https://raw.githubusercontent.com/eslint/eslint/master/templates/rule-change-proposal.md))
[ ] Add autofixing to a rule
[ ] Add a CLI option
[ ] Add something to the core
[ ] Other, please explain:

fixes #12312

**What changes did you make? (Give an overview)**

~~This PR preserves leading whitespace when converting a block comment to a "starred" block comment in the `multiline-comment-style` rule's autofixer. Since this rule doesn't know about indentation, I decided to take the strategy of checking if the leading whitespace matches the expected whitespace (i.e. the whitespace before the opening delimiter for the block comment) and, if it does, to use any remaining whitespace after the asterisk. I think this will work for the vast majority of cases (though it could get weird if folks mix tabs and spaces). If they don't match, it falls back to the current behavior and adds one space.~~

Apologies for the large diff - I ended up doing a lot of refactoring to try to make the code easier to follow. All the calculation of how much whitespace should be used to offset each line now gets calculated in `getCommentLines()`, which I think is a clearer separation of concerns between this module's internal functions.

I believe the autofix should be consistent across formats now. In the case of an easy to use delimiter (`*` for `starred-block` or `//` for `separate-lines`), calculating the whitespace is pretty straightforward. For `bare-block`, the check now iterates through the lines of the comment and checks if any lines are indented less than the opening line, and offsets all the lines by the difference. If it's indented more, it removes the appropriate amount of whitespace. Otherwise, if it's correct, it returns the line unchanged.

**Is there anything you'd like reviewers to focus on?**

1. ~~This only preserves whitespace if the the leading whitespace on the line matches the leading whitespace before the opening delimiter of the block comment (`/*`). Does this seem like a good way to go about this? In other cases, I'm not sure how you would know exactly how much to indent something (particularly if there's a mix of tabs and spaces).~~
1. ~~Not directly related to this change, but I think it's a bit strange that we don't push a space after the asterisk when we convert line comments to "starred" block comments. I have added a test to document this behavior.~~
[...]

PR #12823, opened 2020-01-22
Fix: multiline-comment-style rule add extra space after * (fixes #12785)

### LINKED ISSUE #12785: [multline-comment-style] fix breaks code

* **ESLint Version:** 6.7.2
* **Node Version:** 12.14.1
* **npm Version:** 6.13.4

babel-eslint@10.0.3

<details>

```js

```

</details>

When autofixing the following code it resulted in non-parseable js code. This is a minimal reproducible example.

```js
/* eslint-enable multiline-comment-style */
/*
  // a line comment
  some.code();
*/
```

I expect it to not yield non-parseable code.

It yielded non-parseable code

karthikkp:
#### Prerequisites checklist

[ ] Documentation update
[ ] Add autofixing to a rule
[ ] Add a CLI option
[ ] Add something to the core
[ ] Other, please explain:

fixes #12785 

Added the extra white space which was removed here https://github.com/eslint/eslint/pull/12316/files#diff-e8cce2004dd5e0386021b0e80370c8ebL190

No

### DIFF
--- a/lib/rules/multiline-comment-style.js
+++ b/lib/rules/multiline-comment-style.js
@@ -315,7 +315,7 @@ module.exports = {
                                         const lineTextToAlignWith = sourceCode.lines[firstComment.loc.start.line - 1 + idx];
                                         const [, prefix = "", initialOffset = ""] = lineTextToAlignWith.match(/^(\s*(?:\/?\*)?(\s*))/u) || [];
 
-                                        offset = `${commentTextPrefix.slice(prefix.length)}${initialOffset}`;
+                                        offset = `${commentTextPrefix.slice(prefix.length)}${initialOffset} `;
                                         break;
                                     }
 
--- a/lib/rules/multiline-comment-style.js
+++ b/lib/rules/multiline-comment-style.js
@@ -315,10 +315,10 @@ module.exports = {
                                         const lineTextToAlignWith = sourceCode.lines[firstComment.loc.start.line - 1 + idx];
                                         const [, prefix = "", initialOffset = ""] = lineTextToAlignWith.match(/^(\s*(?:\/?\*)?(\s*))/u) || [];
 
-                                        if (/^\s*\//u.test(lineText)) {
+                                        offset = `${commentTextPrefix.slice(prefix.length)}${initialOffset}`;
+
+                                        if (/^\s*\//u.test(lineText) && offset.length === 0) {
                                             offset = `${commentTextPrefix.slice(prefix.length)}${initialOffset} `;
-                                        } else {
-                                            offset = `${commentTextPrefix.slice(prefix.length)}${initialOffset}`;
                                         }
                                         break;
                                     }
--- a/lib/rules/multiline-comment-style.js
+++ b/lib/rules/multiline-comment-style.js
@@ -315,7 +315,11 @@ module.exports = {
                                         const lineTextToAlignWith = sourceCode.lines[firstComment.loc.start.line - 1 + idx];
                                         const [, prefix = "", initialOffset = ""] = lineTextToAlignWith.match(/^(\s*(?:\/?\*)?(\s*))/u) || [];
 
-                                        offset = `${commentTextPrefix.slice(prefix.length)}${initialOffset} `;
+                                        if (/^\s*\//u.test(lineText)) {
+                                            offset = `${commentTextPrefix.slice(prefix.length)}${initialOffset} `;
+                                        } else {
+                                            offset = `${commentTextPrefix.slice(prefix.length)}${initialOffset}`;
+                                        }
                                         break;
                                     }
 
--- a/lib/rules/multiline-comment-style.js
+++ b/lib/rules/multiline-comment-style.js
@@ -318,7 +318,7 @@ module.exports = {
                                         offset = `${commentTextPrefix.slice(prefix.length)}${initialOffset}`;
 
                                         if (/^\s*\//u.test(lineText) && offset.length === 0) {
-                                            offset = `${commentTextPrefix.slice(prefix.length)}${initialOffset} `;
+                                            offset += " ";
                                         }
                                         break;
                                     }
--- a/tests/lib/rules/multiline-comment-style.js
+++ b/tests/lib/rules/multiline-comment-style.js
@@ -1379,6 +1379,25 @@ ${"                   "}
                 { messageId: "missingStar", line: 4 },
                 { messageId: "endNewline", line: 4 }
             ]
+        },
+        {
+            code: `
+                /*
+                 // a line comment
+                 some.code();
+                 */
+            `,
+            output: `
+                /*
+                 * // a line comment
+                 *some.code();
+                 */
+            `,
+            options: ["starred-block"],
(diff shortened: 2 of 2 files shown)


kaicataldo:
Additionally, we require tests for all changes. Feel free to start with the case reported in the issue!

karthikkp:
> Thanks for working on this! This looks like a regression I introduced in #12316. As you can see [from the tests](https://github.com/eslint/eslint/pull/12316/files#diff-60dc9bbd70dbd85be2088ec3350e6065R1342-R1382), the current behavior is intentional. If the original comment didn't have a leading offset space, it currently tries to preserve that formatting when it autofixes.
> 
> I think the correct fix here would be to check if _any_ line starts with a `/` character and, if there isn't leading whitespace being added, set the offset to a single space character to all lines. This would allow the rule to function as it does now for the majority of cases and handle this edge case gracefully.

Actually the problem is with the regular expression here https://github.com/eslint/eslint/pull/12316/files#diff-e8cce2004dd5e0386021b0e80370c8ebR316 and not the diff in the description. If we split the regular expression into two parts, one for the prefix before * and one after, I can get it to work. Do you want me to do that instead of "If the original comment didn't have a leading offset space, it currently tries to preserve that formatting". That way it code will be generic and we may not have to just check that one edge case specifically.

Mean time I will try this and let you know

kaicataldo:
To be clear, the behavior before the PR mentioned in the description was very different how it is now. 

Can you share an example of what you're thinking? I'm not sure I'm following. I think it's working correctly for other cases besides this one edge case, and we should keep the current behavior the same (which should be easy to verify since this rule is tested fairly well).

karthikkp:
So what I am saying is, in the diff pasted, this problem occurs when the preceeding line also does not have * and initial offset hence is an empty string. So I can just add a check to see if the initial offset is an empty string and if it is just set initialoffset to a string with a space.

kaicataldo:
@karthikkp If I'm understanding correctly, that would cause the tests I linked above to fail. The rule tries to preserve the leading whitespace (or lack thereof) when it's converting the rule, so it's sometimes correct for the initial offset to be an empty string. I think this is the correct behavior (and was intentional when I made that change). Are there any other cases where malformed code can occur other than when there's no leading whitespace and the first character of a comment line is a `/` character?

karthikkp:
Oh okay got it. Then I think handling the edge case alone will be the way to go. Will do that

tests/lib/rules/multiline-comment-style.js
```diff
@@ -1379,6 +1379,25 @@ ${"                   "}
                 { messageId: "missingStar", line: 4 },
                 { messageId: "endNewline", line: 4 }
             ]
+        },
```
kaicataldo: Can we make this test suite a bit more robust by adding other kinds of formatted comments being formatted to `starred-block`?

Some examples:
```js
////This comment is in
//`separate-lines` format.

// // This comment is in
// `separate-lines` format.


/*
  * // a line comment
 *some.code();
 */


/*
// a line comment
 * some.code();
 */
```

lib/rules/multiline-comment-style.js
```diff
@@ -315,7 +315,11 @@ module.exports = {
                                         const lineTextToAlignWith = sourceCode.lines[firstComment.loc.start.line - 1 + idx];
                                         const [, prefix = "", initialOffset = ""] = lineTextToAlignWith.match(/^(\s*(?:\/?\*)?(\s*))/u) || [];
 
-                                        offset = `${commentTextPrefix.slice(prefix.length)}${initialOffset}`;
+                                        if (/^\s*\//u.test(lineText)) {
```
yeonjuan: Because the regex(`/^\s*\//u`) checks all leading whitespace characters(`^\s*`) before the slash(`/`), In some cases, unnecessary  `space`  can be placed.

For example, this comment is not cracked by a fixer in the actual version because of the tab character(`/t`). - [demo](https://eslint.org/demo#eyJ0ZXh0IjoiLyogZXNsaW50IG11bHRpbGluZS1jb21tZW50LXN0eWxlOiBlcnJvciAqL1xuLypcbntcblxcdFwiZm9vXCI6IDEsXG5cXHQvL1wiYmFyXCI6IDJcbn1cbiovIiwib3B0aW9ucyI6eyJwYXJzZXJPcHRpb25zIjp7ImVjbWFWZXJzaW9uIjoxMSwic291cmNlVHlwZSI6Im1vZHVsZSIsImVjbWFGZWF0dXJlcyI6eyJqc3giOnRydWV9fSwicnVsZXMiOnt9LCJlbnYiOnsiYnJvd3NlciI6ZmFsc2V9fX0=)
* before
```js
/*
{
\t"foo": 1,
\t//"bar": 2
}
*/
```
* the actual version - after fix
```js
/*
*{
*\t"foo": 1,
*\t//"bar": 2
*}
*/
```

But in this PR, the fixer will place an unnessasary space before `/t`.
* this PR - after fix
```js
/*
*{
*\t"foo": 1,
* \t//"bar": 2
*}
*/
```
Can we make it insert a `space` only if its fixed result can be cracking?

karthikkp: I will make a local build with my changes and test the cases you have presented here. Thank you

tests/lib/rules/multiline-comment-style.js
karthikkp: The first two are not converted to starred block comments in the current version. So the fix that I am working should not change that behavior right?

lib/rules/multiline-comment-style.js
karthikkp: This has been handled now

yeonjuan: @karthikkp Thanks 👍 
Can we add tests for ensuring it? it is a example.
* code
```js
/*
{
\t"foo": 1,
\t// a line comment
}
*/
```
* output
```js
/*
*{
*\t"foo": 1,
*\t// a line comment
*}
*/
```

lib/rules/multiline-comment-style.js
```diff
@@ -316,6 +316,10 @@ module.exports = {
                                         const [, prefix = "", initialOffset = ""] = lineTextToAlignWith.match(/^(\s*(?:\/?\*)?(\s*))/u) || [];
 
                                         offset = `${commentTextPrefix.slice(prefix.length)}${initialOffset}`;
+
+                                        if (/^\s*\//u.test(lineText) && offset.length === 0) {
+                                            offset = `${commentTextPrefix.slice(prefix.length)}${initialOffset} `;
```
yeonjuan: In my opinion, we can simplify this by appending a space character.
```js
 offset += " ";
```

tests/lib/rules/multiline-comment-style.js
kaicataldo: Oh, interesting. Yes, that makes sense to me. Thanks for adding these tests.

karthikkp:
@yeonjuan The changes you had requested have been made 👍

kaicataldo:
Thanks for contributing to ESLint!

PR #12887, opened 2020-02-08
Breaking: change relative paths with --config (refs eslint/rfcs#37)

### LINKED ISSUE #11558: Override files globbing logic breaks with --config

* **ESLint Version:** 5.15.3
* **Node Version:** 10.15.0
* **npm Version:** 6.9.0

<details>

```js
{
    "overrides": [{
        "files": ["**/blah.js"],
        "rules": {
            "prefer-const": 0
        }
    }],
    "parserOptions": {
        "ecmaVersion": 2018
    },
    "rules": {
        "prefer-const": 2
    }
}
```

</details>

In this case, I have a configuration that is inside of my node_modules, or another subdirectory. Using the `--config` flag, I use that file directly rather than having a config in the root of my project. In the file, as above, I've got a glob that should match some files. However, that glob is not matched correctly. In particular, in `getConfigVector` in `config`, we use `config.baseDirectory` to compute a relative path. In this case, because the config is in a sibling directory, that relative path starts with something like `../`, which makes the glob always fail.

```js
N/A
```

```bash
eslint --config ./configs/.eslintrc.json .
```

It seems likely that if someone is using `--config`, they probably want the globs to be computed relative to `cwd` rather than wherever that config happens to be on disk.

Globs are computed relative to the config's location on disk, which makes them not work super well.

mysticatea:
#### Prerequisites checklist

[X] Other, please explain: change some existing core features: fixes #11558.

This PR implements RFC37 -- changes the relative paths of `ignorePatterns`, `overrides[].files`, and `overrides[].excludedFiles` as being relative to the current working directory if the config is provided by the `--config` option.

I'm sorry for these large differences. Most of the changes are to replace the pair of a name and a file path in parameters to a context object. Because most internal methods in `ConfigArrayFactory` had the pair in its parameter, the differences went huge. The context object has four properties; `type`, `name`, `filePath`, and `matchBasePath`. The `matchBasePath` is the new stuff to specify the base path of `ignorePatterns`, `overrides[].files`, and `overrides[].excludedFiles`. (the `type` was moved from config data because it's a context data rather than a config data.)

On the way, I found an unused option `options.parent` in some methods of `ConfigArrayFactory` and remove those.

Nothing in particular.


lib/cli-engine/config-array-factory.js
```diff
@@ -379,61 +401,91 @@ class ConfigArrayFactory {
...
+                basePath
+            );
+
+            if (fs.existsSync(ctx.filePath)) {
+                let configData;
+
+                try {
+                    configData = loadConfigFile(ctx.filePath);
+                } catch (error) {
+                    if (!error || error.code !== "ESLINT_CONFIG_FIELD_NOT_FOUND") {
```
kaicataldo: Not a blocker for this PR, but an idea I had while reviewing this: maybe we could centralize these errors codes as constants that can be imported?

nzakas: I’d suggest opening an issue for that idea so we can focus the conversation on this PR around what needs to happen to get this merged.

mysticatea: For information, this constant value should appear in the userland only if `--config path/to/a/package.json` is present, the `package.json` exists, and `eslintConfig` field doesn't exist in the `package.json`. (We can distinguish between `ENOENT` and that by code, but it may be not important.)

mysticatea:
Thank you for your review. And I'm sorry for my late response. I have been back from the funeral of my grandma and around procedures.

PR #12896, opened 2020-02-09
Fix: `--no-ignore` does not work with `--stdin-filename` (fixes #12745)

### LINKED ISSUE #12745: `--no-ignore` not work with `--stdin-filename`

* **ESLint Version:** v6.8.0
* **Node Version:** v12.14.0
* **npm Version:** v6.13.4

default

```bash
$ cd ~
$ touch .hello.js
$ eslint .hello.js # ignored, expected
$ eslint --no-ignore .hello.js # not ignored, expected
$ cat .hello.js | eslint --no-ignore --stdin --stdin-filename .hello.js # ignored, not expected
```
I am using Vim plugin [ale](https://github.com/dense-analysis/ale), this plugin always call `eslint` with  `--stdin-filename`, even I append `--no-ignore` to ale option, eslint still ignore the file, so all the files in my $HOME always be ignored when it is opened in Vim.

File shouln't be ignored with below command:

```bash
$ cat .hello.js | eslint --no-ignore --stdin --stdin-filename .hello.js 
```

No

karthikkp:
#### Prerequisites checklist

[ ] Documentation update
[ ] Add autofixing to a rule
[ ] Add a CLI option
[ ] Add something to the core
[ ] Other, please explain:

Added a pattern match check for execute on text  to test if the stdin filename is a dot file just like how it's being tested if eslint is called on files.

Any side effects to core logic.

### DIFF
--- a/lib/cli-engine/cli-engine.js
+++ b/lib/cli-engine/cli-engine.js
@@ -863,7 +863,10 @@ class CLIEngine {
         // Clear the last used config arrays.
         lastConfigArrays.length = 0;
 
-        if (resolvedFilename && this.isPathIgnored(resolvedFilename)) {
+        const dotfilesPattern = /(?:(?:^\.)|(?:[/\\]\.))[^/\\.].*/u;
+        const isDot = dotfilesPattern.test(resolvedFilename);
+
+        if (resolvedFilename && this.isPathIgnored(resolvedFilename, isDot)) {
             if (warnIgnored) {
                 results.push(createIgnoreResult(resolvedFilename, cwd));
             }
@@ -934,9 +937,10 @@ class CLIEngine {
     /**
      * Checks if a given path is ignored by ESLint.
      * @param {string} filePath The path of the file to check.
+     * @param {boolean} isDot is the file a dot file.
      * @returns {boolean} Whether or not the given path is ignored.
      */
-    isPathIgnored(filePath) {
+    isPathIgnored(filePath, isDot) {
         const {
             configArrayFactory,
             defaultIgnores,
@@ -950,10 +954,9 @@ class CLIEngine {
                 .extractConfig(absolutePath);
             const ignores = config.ignores || defaultIgnores;
 
-            return ignores(absolutePath);
+            return ignores(absolutePath, isDot);
         }
-
-        return defaultIgnores(absolutePath);
+        return defaultIgnores(absolutePath, isDot);
     }
 
     /**


lib/cli-engine/cli-engine.js
```diff
@@ -863,7 +863,10 @@ class CLIEngine {
         // Clear the last used config arrays.
         lastConfigArrays.length = 0;
 
-        if (resolvedFilename && this.isPathIgnored(resolvedFilename)) {
+        const dotfilesPattern = /(?:(?:^\.)|(?:[/\\]\.))[^/\\.].*/u;
```
nzakas: Is there a reason this logic can’t go into `isPathIgnored()`?

lib/cli-engine/cli-engine.js
```diff
@@ -934,9 +937,10 @@ class CLIEngine {
     /**
      * Checks if a given path is ignored by ESLint.
      * @param {string} filePath The path of the file to check.
+     * @param {boolean} isDot is the file a dot file.
      * @returns {boolean} Whether or not the given path is ignored.
      */
-    isPathIgnored(filePath) {
+    isPathIgnored(filePath, isDot) {
```
nzakas: This is part of our published API so it would be preferable not to change the function signature.

lib/cli-engine/cli-engine.js
karthikkp: We need the resolved filename to test the dot files pattern. That's why it has to go here.

kaicataldo: I don't think we should add `isDot` to the public API of `isPathIgnored()`, since that's something the method can calculate (as demonstrated here!). `isPathIgnored()` receives the resolved filename as argument, so we should be able to avoid changing the public API by moving this check in that method.

karthikkp: https://github.com/eslint/eslint/blob/master/tests/lib/cli-engine/cli-engine.js#L3849
https://github.com/eslint/eslint/blob/master/tests/lib/cli-engine/cli-engine.js#L3841

These test cases are failing. However they sound counter intuitive. When ignore is false, we want the dot files to be not ignored right?

kaicataldo: My guess is that the intention is to avoid linting directories such as `.git/` when the `--no-ignore` option is set. I think we should scope this PR to the specific bug fix report not change that behavior in this PR (though we always welcome RFCs for core changes if you feel differently).

karthikkp: They don't fail if the check is made within the execute on text function and isDot passed as an extra parameter to isPathIgnored

lib/cli.js
```diff
@@ -68,7 +68,8 @@ function translateOptions(cliOptions) {
         allowInlineConfig: cliOptions.inlineConfig,
         reportUnusedDisableDirectives: cliOptions.reportUnusedDisableDirectives,
         resolvePluginsRelativeTo: cliOptions.resolvePluginsRelativeTo,
-        errorOnUnmatchedPattern: cliOptions.errorOnUnmatchedPattern
+        errorOnUnmatchedPattern: cliOptions.errorOnUnmatchedPattern,
+        stdin: cliOptions.stdin
```
karthikkp: If it's okay to pass to pass this option to cli-engine we can solve the issue. However I am not sure, to solve a problem we pass this option that is required only for one specific use case and useless the rest of the times.

lib/cli-engine/cli-engine.js
karthikkp: @kaicataldo I have modified to code. Let me know if that will work.

lib/cli.js
kaicataldo: This feels like an improvement, but I'm personally wary of adding more options to any of our public APIs just to handle one case like this.

One thought: Is there any way we can use the existing `CLIEngine` API to instantiate it with ignore patterns that negate the defaults?

If we are going to change the public API, I think we should add something more generic and self-explanatory like `noIgnore: boolean`.

mysticatea:
I think that `executeOnText` method should not ignore given text regardless of stdin if `--no-ignore` was present because of consistent with `executeOnFiles` method. 

If `--no-ignore` was present and a path to a file was given,  `executeOnFiles` method doesn't ignore the file even if it's a dotfile. On the other hand, if `--no-ignore` was present and a path to a directory (or a glob) was given,  `executeOnFiles` method ignores dotfiles in the directory (or in matched files).
The `executeOnText` case is similar to the former.

nzakas:
@kaicataldo @mysticatea how do we want to proceed here?

kaicataldo:
I think @mysticatea's rationale makes a lot of sense and support not changing this.

nzakas:
Does that mean we can close this?

Issue #12835, opened 2020-01-27
Deprecate Node-specific core rules


kaicataldo:
**The version of ESLint you are using.**
`master`

**The problem you want to solve.**

The number of rules in core has grown to the point that we have to set a very high bar for new rules. To improve our ability to maintain core, I think it makes sense to treat the current core Node/CommonJS rules similarly to how we handled the deprecation of [core JSDoc rules](https://eslint.org/blog/2018/11/jsdoc-end-of-life). I believe these are the last runtime/library specific rules that we have in core, and think this also makes sense for the sake of consistency.

**Your take on the correct solution to problem.**

As discussed in our TSC meeting (notes [here](https://github.com/eslint/tsc-meetings/blob/master/notes/2019/2019-12-19.md#update-no-path-concat-rule-checks-template-literals-fixes-12323)), this proposal is to deprecate all the Node/CommonJS-specific rules in core and to move all of them into [`eslint-plugin-node`](https://github.com/mysticatea/eslint-plugin-node), maintained by @mysticatea.

**Are you willing to submit a pull request to implement this change?**
Yes!

**Questions**
1. @mysticatea Are you still okay with this? Just wanted to check because it's been a little while since we last discussed this.

mysticatea:
Thank you for this issue.

> Are you still okay with this? Just wanted to check because it's been a little while since we last discussed this.

Yes!

kaicataldo:
Marking as accepted since this was decided on during a TSC meeting and @mysticatea is still on board.

kaicataldo:
I'll work on this :)

kaicataldo:
# Steps to complete this:
- [ ] Port rules over to `eslint-plugin-node` (https://github.com/mysticatea/eslint-plugin-node/pull/206)
- [ ] Deprecate rules in core
- [ ] Write a blog post similar to https://eslint.org/blog/2018/11/jsdoc-end-of-life

ljharb:
I know it might be late, but this seems like a terrible idea. Having rules in core is paramount for discoverability, especially since shared configs and peer deps are still a huge PITA, until shared configs can bring along plugins as regular deps.

PR #12898, opened 2020-02-10
Update: deprecate Node.js & CommonJS rules

kaicataldo:
#### Prerequisites checklist

[ ] Documentation update
[ ] Add autofixing to a rule
[ ] Add a CLI option
[ ] Add something to the core
[ ] Other, please explain:

refs #12835 

As discussed in https://github.com/eslint/eslint/issues/12835, this PR deprecates the Node.js and CommonJS rules we have in core. These rules are generously being adopted by @mysticatea's `eslint-plugin-node` project (PR [here](https://github.com/mysticatea/eslint-plugin-node/pull/206))!

Is there anything I missed?

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -16,7 +16,7 @@ const INTERNAL_FILES = {
  * @param {string} pathOrPattern the path or glob pattern.
  * @returns {string} The resolved path or glob pattern.
  */
-function resolveAbsPath(pathOrPattern) {
+function resolveAbsolutePath(pathOrPattern) {
     return path.resolve(__dirname, pathOrPattern);
 }
 
@@ -32,10 +32,10 @@ function createInternalFilesPatterns(pattern = null) {
             name: [
 
                 // Disallow all children modules.
-                resolveAbsPath(p),
+                resolveAbsolutePath(p),
 
                 // Allow the main `index.js` module.
-                `!${resolveAbsPath(p.replace(/\*\*\/\*$/u, "index.js"))}`
+                `!${resolveAbsolutePath(p.replace(/\*\*\/\*$/u, "index.js"))}`
             ]
         }));
 }
@@ -125,7 +125,7 @@ module.exports = {
             rules: {
                 "node/no-restricted-require": ["error", [
                     ...createInternalFilesPatterns(INTERNAL_FILES.CLI_ENGINE_PATTERN),
-                    resolveAbsPath("lib/init/index.js")
+                    resolveAbsolutePath("lib/init/index.js")
                 ]]
             }
         },
@@ -134,7 +134,7 @@ module.exports = {
             rules: {
                 "node/no-restricted-require": ["error", [
                     ...createInternalFilesPatterns(INTERNAL_FILES.INIT_PATTERN),
-                    resolveAbsPath("lib/rule-tester/index.js")
+                    resolveAbsolutePath("lib/rule-tester/index.js")
                 ]]
             }
         },
@@ -144,9 +144,9 @@ module.exports = {
                 "node/no-restricted-require": ["error", [
                     ...createInternalFilesPatterns(INTERNAL_FILES.LINTER_PATTERN),
                     "fs",
-                    resolveAbsPath("lib/cli-engine/index.js"),
-                    resolveAbsPath("lib/init/index.js"),
-                    resolveAbsPath("lib/rule-tester/index.js")
+                    resolveAbsolutePath("lib/cli-engine/index.js"),
+                    resolveAbsolutePath("lib/init/index.js"),
+                    resolveAbsolutePath("lib/rule-tester/index.js")
                 ]]
             }
         },
@@ -156,11 +156,11 @@ module.exports = {
                 "node/no-restricted-require": ["error", [
                     ...createInternalFilesPatterns(INTERNAL_FILES.RULES_PATTERN),
                     "fs",
-                    resolveAbsPath("lib/cli-engine/index.js"),
-                    resolveAbsPath("lib/init/index.js"),
-                    resolveAbsPath("lib/linter/index.js"),
-                    resolveAbsPath("lib/rule-tester/index.js"),
-                    resolveAbsPath("lib/source-code/index.js")
+                    resolveAbsolutePath("lib/cli-engine/index.js"),
+                    resolveAbsolutePath("lib/init/index.js"),
+                    resolveAbsolutePath("lib/linter/index.js"),
+                    resolveAbsolutePath("lib/rule-tester/index.js"),
+                    resolveAbsolutePath("lib/source-code/index.js")
                 ]]
             }
         },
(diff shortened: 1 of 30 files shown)


kaicataldo:
This is blocked by https://github.com/mysticatea/eslint-plugin-node/pull/206 being released. We'll need to switch over to the plugin internally before tests will pass (since this change makes it so that we're using deprecated rules).

lib/rules/callback-return.js
```diff
@@ -10,6 +10,10 @@
 
 module.exports = {
     meta: {
+        deprecated: true,
+
+        replacedBy: [],
```
nzakas: Would it make sense to include the rule it's replaced by in the plugin?

kaicataldo: I went back and forth on that. Since we won’t be maintaining it, I’m a little worried about it getting out of sync (even though it’s maintained by one of us). Happy to defer to consensus on this, though.

nzakas: I think if we are recommending people use the plugin in our docs and release notes then it makes sense to be part of the console deprecation notice.

kaicataldo: :+1: Definitely not against that. Waiting on `eslint-plugin-node` to be released and will update this PR.

kaicataldo:
Thoughts on leaving the `replacedBy` fields empty for now and then following up with a PR pointing them to the correct URL when `eslint-plugin-node` is released and the docs are generated for the new rules?

nzakas:
Sounds good.

kaicataldo:
After looking closer at the test failures, I do think we should wait for the release of `eslint-plugin-node`, since we utilize `no-restricted-modules` quite heavily in our own config. 

@mysticatea Is there anything I can do to help with that?

mysticatea:
Hi. I apologies that I have not been active while a time. At last, I have released `eslint-plugin-node` with some tweaks: https://github.com/mysticatea/eslint-plugin-node/releases/tag/v11.1.0

kaicataldo:
No need to apologize! We all totally understand. Just wanted to check in to see if there was anything I could do to help.

nzakas:
@kaicataldo does the plugin release unblock this now?

kaicataldo:
Yes! This is now unblocked.

kaicataldo:
@mysticatea I've been trying to update this PR to use the updated plugin, and I've been wrestling with the new `minimatch` globs trying to emulate [this](https://github.com/eslint/eslint/blob/master/.eslintrc.js#L4). We want to be able to allow importing the directory (`index.js`) but not allowing for any subpaths of the directory.

I've tried the following:
```js
"**/lib/cli-engine/**/*"
```
```js
[path.join(__dirname, "lib/cli-engine/**/*"), path.join(__dirname, "!lib/cli-engine")]
```
```js
"lib/cli-engine/{index.js,**/*}"
```
```js
"lib/cli-engine/(!index.js|**)/*}"
```
```js
// I thought this one would work!
[path.join(__dirname, "lib/cli-engine/**/*"), path.join(__dirname, "!lib/cli-engine/index.js")]
```

Do you see anything glaringly wrong here? I'll take a look at the rule to figure it out, but I've spent enough time trying to figure the glob out that I figured I should just ask :) Thanks!

kaicataldo:
Ah, user error - of course, I figured it out right after typing the above :laughing:. Here's the correct answer:

```js
[
    path.join(__dirname, "lib/cli-engine/**/*"),
    `!${path.join(__dirname, "lib/cli-engine/**/*")}`
]
```

kaicataldo:
This should be ready for review (with a few questions).

Tests are currently failing because of https://github.com/mysticatea/eslint-plugin-node/pull/211.

packages/eslint-config-eslint/default.yml
```diff
@@ -106,14 +102,12 @@ rules:
...
     no-new-object: "error"
-    no-new-require: "error"
     no-new-wrappers: "error"
     no-octal: "error"
     no-octal-escape: "error"
     no-param-reassign: "error"
-    no-path-concat: "error"
-    no-process-exit: "error"
     no-proto: "error"
+    no-process-exit: "off"
```
kaicataldo, line 110: This is added due to the bug described [here](https://github.com/mysticatea/eslint-plugin-node/pull/211). Do we want to go ahead and merge this or do we want to wait on the next `eslint-plugin-node` release?

.eslintrc.js
```diff
@@ -84,106 +115,90 @@ module.exports = {
...
             }
         },
         {
-            files: ["lib/cli-engine/**/*"],
+            files: [INTERNAL_FILES.CLI_ENGINE_PATTERN],
             rules: {
-                "no-restricted-modules": ["error", {
-                    patterns: [
-                        ...internalFiles,
-                        "**/init"
```
kaicataldo, line 100: I'm not sure why we specifically disallow these directory imports (`init/index.js`, etc.). Do we want to continue doing this? It feels a bit overkill to me, but maybe I'm missing some context.

lib/rules/callback-return.js
kaicataldo: This has been updated!

packages/eslint-config-eslint/default.yml
mysticatea: Because the update of recommended configs is a breaking change, I have not updated that yet (I think that it's not a problem because deprecation is not removal). I will do the next major release after the ESTree matter is completed.

.eslintrc.js
mysticatea: That's to prevent depending `fs` accidentally and also prevent cyclic depending. The diagram in the [Architecture](https://eslint.org/docs/developer-guide/architecture) page describes the depending directions of each directory.

packages/eslint-config-eslint/default.yml
kaicataldo: Got it, thanks. In that case, I think we should leave this in for now.

.eslintrc.js
kaicataldo: Thanks for the explanation.

.eslintrc.js
```diff
@@ -1,13 +1,44 @@
...
+    RULES_PATTERN: "lib/rules/**/*",
+    SOURCE_CODE_PATTERN: "lib/source-code/**/*"
+};
+
+/**
+ * Resolve an absolute path or glob pattern.
+ * @param {string} pathOrPattern the path or glob pattern.
+ * @returns {string} The resolved path or glob pattern.
+ */
+function resolveAbsPath(pathOrPattern) {
```
nzakas, line 19: Maybe resolveAbsolutePath? We’ve generally not abbreviated words in the codebase.

docs/rules/no-sync.md
```diff
@@ -1,5 +1,7 @@
 # Disallow Synchronous Methods (no-sync)
 
+This rule was **deprecated** in ESLint v7.0.0.
```
nzakas, line 3: Is it worth adding a note about the node plugin as a replacement?

lib/rules/no-sync.js
```diff
@@ -13,6 +13,10 @@
 
 module.exports = {
     meta: {
+        deprecated: true,
+
+        replacedBy: ["https://github.com/mysticatea/eslint-plugin-node/blob/master/docs/rules/no-sync.md"],
```
nzakas, line 18: Should this be the rule name instead of the URL?

kaicataldo: I was wondering that too, and ended up leaning toward this to save folks the step of having to google it. That being said, it does feel weird. Happy to go with consensus on this.

nzakas: It looks like we don’t have other examples using URLs in this spot, so it seems using the rule name would be consistent.

PR #12922, opened 2020-02-16
Breaking: improve plugin resolving (refs eslint/rfcs#47)

mysticatea:
#### Prerequisites checklist

[X] Other, please explain: change existing core features

This PR implements a part of RFC47.

- https://github.com/eslint/eslint/commit/3f092158507090ff94794339701a76e9aa6d24cc changes the location that ESLint loads plugins from. ([link](https://github.com/eslint/rfcs/blob/master/designs/2019-plugin-loading-improvement/README.md#-it-changes-where-eslint-loads-plugins-from))
- ~~https://github.com/eslint/eslint/commit/d29f613c9d8a8106ca0d68bb324415807dca4a20 deprecates `CLIEngine::getRules()` and adds `CLIEngine::getRulesForFile(filePath)`. ([link](https://github.com/eslint/rfcs/blob/master/designs/2019-plugin-loading-improvement/README.md#-it-replaces-the-clienginegetrules-method))~~
- ~~https://github.com/eslint/eslint/commit/0d6afaf6a40792c05659796ec4c26258e5143494 deprecates `metadata.rulesMeta` and adds `metadata.getRuleMeta(ruleId, filePath)`. ([link](https://github.com/eslint/rfcs/blob/master/designs/2019-plugin-loading-improvement/README.md#-it-replaces-the-metadatarulesmeta-property-of-formatters))~~
- ~~https://github.com/eslint/eslint/commit/f3cc32f5c07579aecfc4566201f19918545aca6d deprecates `report.usedDeprecatedRules` and adds `report.results[i].usedDeprecatedRules`. (this is not in RFC47, but is a part of RFC40. I found that `report.usedDeprecatedRules` requires the plugin uniqueness over multiple target files, but RFC47 overlooked it. Fortunately, RFC40 mentions the move of `usedDeprecatedRules`.)~~

Nothing in particular.


mysticatea:
Yes.

1. RFC47 changes the common parts of `CLIEngine` class and `ESLint` class -- the logic of loading plugins.
1. RFC47 deprecates a problematic part of `ESLint` class -- `rulesMeta` depends on `getRules()`, but `ESLint` class doesn't have `getRules()` because it requires mutable state internally. The removal of `rulesMeta` is important for `ESLint` class.
1. Handling `usedDeprecatedRules` in `CLIEngine` class makes `ESLint` class implementation simpler and more correct.

mysticatea:
I updated this PR:

- 1d8beae, 53f6407, and 24130c1 reverted deprecations in order to split PR.
- 3b2790f updated documentation.

kaicataldo:
It looks like tests might be failing due to https://github.com/eslint/eslint/pull/13076.

docs/user-guide/configuring.md
```diff
@@ -321,7 +321,8 @@ And in YAML:
     - eslint-plugin-plugin2
 ```
 
-**Note:** Plugins are resolved relative to the current working directory of the ESLint process. In other words, ESLint will load the same plugin as a user would obtain by running `require('eslint-plugin-pluginname')` in a Node REPL from their project root.
+* **Note1:** Plugins are resolved relative to the config file. In other words, ESLint will load the plugin as a user would obtain by running `require('eslint-plugin-pluginname')` in the config file.
```
kaicataldo: Can we reformat this to something like the following?

```suggestion
**Notes:**

1. Plugins are resolved relative to the config file. In other words, ESLint will load the plugin as a user would obtain by running `require('eslint-plugin-pluginname')` in the config file.

2. Plugins in the base configuration (loaded by `extends` setting) are relative to the derived config file. For example, if `./.eslintrc` has `extends: ["foo"]` and the `eslint-config-foo` has `plugins: ["bar"]`, ESLint finds the `eslint-plugin-bar` from `./node_modules/` (rather than `./node_modules/eslint-config-foo/node_modules/`) or ancestor directories. Thus every plugin in the config file and base configurations is resolved uniquely.
```

kaicataldo:
Thanks for working on this!

bjmiller:
This PR utterly broke an important use case for me.  Can there be a fix/patch that optionally restores the original behavior?

We have an organization-wide config that exists specifically so that people _don't_ need to explicitly import all of the plugins in their local project.  The command line option doesn't really help, but a flag in the local project's root config might.

mysticatea:
You can restore the original behavior by `--resolve-plugins-relative-to .` option.

I'd recommend installing plugins into the same place as the config file that uses it. Otherwise, looks like `package.json` includes plugins but ESLint configuration doesn't use those.

And, it's good practice if your project configuration files include `root: true` in order to not use the configuration that came from outside of the project accidentally.

bjmiller:
That's the problem.  We _need_ to primarily use the central configuration (and plugins), but still allow each project to add their own additional ones if they choose.  And, if we add a plugin to the central configuration, we need that to propagate to all the projects without requiring them to change their local .eslintrc.js file.

Can there be a way to specify _two_ directories in `--resolve-plugins-relative-to`, attempted in order?

mysticatea:
You can use `extends` to extend shared configuration.

> Can there be a way to specify two directories in --resolve-plugins-relative-to, attempted in order?

No. Previously, eslint loads all plugins from cwd. The `--resolve-plugins-relative-to .` is exactly same behavior as eslint 6.

bjmiller:
Thank you for being patient, @mysticatea.  I'll try using the command line option, and see if it works for what I need.

arcanis:
Note that if you use the Yarn workspaces, you can make a workspace package containing your shared configuration, and make everything else depend on it. This way your shared plugins are properly scoped to your conf, and individual consumers can add their own additions if needed.

PR #12939, opened 2020-02-19
Update: Add ESLint API (refs eslint/rfcs#40)

kaicataldo:
#### Prerequisites checklist

[ ] Documentation update
[ ] Add autofixing to a rule
[ ] Add a CLI option
[x] Add something to the core
[ ] Other, please explain:

refs eslint/rfcs#40

This is a WIP. I received some initial feedback [here](https://github.com/eslint/eslint/commit/0195f126b019e0494f0f3d600e046656c69f83b1) and have incorporated that into this PR.

This PR adds the new `ESLint` class, which is an asynchronous wrapper around the `CLIEngine` class.

Still a WIP, but as I get closer to completion: are the options/type defs correct?


lib/eslint/eslint.js
```diff
@@ -0,0 +1,457 @@
...
+        CLIEngine.outputFixes(report);
+    }
+
+    /**
+     * An Array.prototype.sort() compatible compare function to order results by their file path.
+     * @param {LintResult} a The first lint result.
+     * @param {LintResult} b The second lint result.
+     * @returns {number} An integer representing the order in which the two results should occur.
+     */
+    static compareResultsByFilePath(a, b) {
```
kaicataldo: Picking this up again today, and I'm reconsidering whether this method should be part of this PR. Since we're still running the lint job synchronously, this feels a bit premature, and I think it should be deferred until we actually need it (when parallel linting is implemented). This will allow us to alter the design if we need to and avoid breaking changes. Thoughts?

btmills: That makes sense. It's unnecessary right now, and leaving it out decreases the surface area of this PR. We're then free to add it in a semver-minor change later.

nzakas: If it’s not needed, good idea to leave it out. It’s a lot easier to add new APIs later than it is to remove them.

kaicataldo: Thanks for the quick feedback, all. Will defer this addition and remove it from this PR.

kaicataldo:
Closing this for now while I’m out sick, in case someone else can get to it first.

nzakas:
Reopening in case someone wants to work off of this branch (and so it doesn't get removed from the project board).

lib/eslint/eslint.js
```diff
@@ -0,0 +1,471 @@
...
+    envs = [],
+    extensions = null, // TODO: Should we update the CLIEngine check to account for empty arrays?
+    fix = false,
+    fixTypes = [], // TODO: We don't currently set a default value for this. Doing so changes the behavior.
+    globals = [],
+    globInputPaths = true,
+    ignore = true,
+    ignorePath = null,
+    ignorePattern = [],
+    parser = null, // Question: if this is set to a default value it ends up overriding the value set in the config. That seems like unexpected behavior to me.
```
mysticatea: > Question: if this is set to a default value it ends up overriding the value set in the config. That seems like unexpected behavior to me.

All CLI options override the settings in config files. Therefore, if `options.parser` is present, it overrides the value of config files. On the other hand, `options.baseConfig.parser` doesn't override the value of config files.

Maybe we should aggregate CLI options that overrides config files to one such as `cliOptions` because now the class name is not `CLIEngine`.

mysticatea: How about https://github.com/eslint/eslint/commit/c96d18dda71a176ee691f560fa7c8bdad329ff0c?

- It adds `overrideConfig` that is similar to `baseConfig`.
- It renames `configFile` to `overrideConfigFile` to be clear.
    1. The loaded config data from `overrideConfigFile` will be added to the tail of every config array.
    2. The normalized config data from `overrideConfig` will be added to the tail of every config array after `overrideConfigFile` is handled.
- It removes `allowInlineConfig`, `envs`, `globals`, `ignorePattern`, `parser`, `parserOptions`, and `rules` because of preferring the corresponded fields of `overrideConfig`.
- It changes `plugins` to only for `addPlugin`. It doesn't add the `plugins` setting of configurations as same as the current `addPlugin`. The `plugins` of CLIEngine moved to `overrideConfig.plugins`.
- The `ESLint` constructor shows error messages to show the corresponded fields of `overrideConfig` if those removed options are present.

mysticatea: The `overrideConfigFile` may not be needed because we can use  `overrideConfig.extends` instead.

mysticatea:
Maybe I have done:

- rewrite the tests from `cli-engine.js`.
- remove compareResultsByFilePath.
- update option validation stricter.
- update `bin/eslint.js` and `lib/cli.js` to use new `ESLint` class.
- change the `ESLint` constructor options:
    - add `overrideConfig` (similar to `baseConfig`, but is pushed to the last of every config array).
    - remove `envs`, `globals`, `ignorePattern`, `parser`, `parserOptions`, `rules` in favor of `overrideConfig`
    - rename `configFile` to `overrideConfigFile`.
    - split `plugins`:
        - `overrideConfig.plugins` to add plugins into configurations to load the plugins. Corresponds to the `plugins` option of CLIEngine constructor.
        - `plugins` to add plugin definitions to be loaded from configurations. Corresponds to the `CLIEngine#addPlugin()` method. This is an object (key is ID and value is plugin definition). If this is an array, throw an error to suggest `overrideConfig.plugins`.

----

- Is the options' change good?
- Is the `ESLint.getErrorResults()` method needed at this time?

lib/cli.js
```diff
@@ -17,105 +17,176 @@
...
+
+/** @typedef {import("./eslint/eslint").ESLintOptions} ESLintOptions */
+/** @typedef {import("./eslint/eslint").LintMessage} LintMessage */
+/** @typedef {import("./eslint/eslint").LintResult} LintResult */
+
 //------------------------------------------------------------------------------
 // Helpers
 //------------------------------------------------------------------------------
 
+const mkdir = promisify(fs.mkdir);
```
nzakas, line 40: Can we use the fs promises API instead?
https://nodejs.org/api/fs.html#fs_fs_promises_api

lib/eslint/eslint.js
```diff
@@ -0,0 +1,655 @@
...
+ * @typedef {Object} ESLintPrivateMembers
+ * @property {CLIEngine} cliEngine The wrapped CLIEngine instance.
+ * @property {ESLintOptions} options The options used to instantiate the ESLint instance.
+ */
+
+//------------------------------------------------------------------------------
+// Helpers
+//------------------------------------------------------------------------------
+
+const writeFile = promisify(fs.writeFile);
```
nzakas, line 88: Same question re: fs promises API

lib/cli.js
mysticatea: No. We have decided the supported Node.js version is `10.12.0`, but the `fs.promises` needs `10.17.0`. See also https://github.com/eslint/rfcs/pull/44#issuecomment-546804246

nzakas: Ah ok. 👍

mysticatea:
I had forgotten it. I have added the documentation.

lib/eslint/eslint.js
```diff
@@ -0,0 +1,665 @@
+/**
+ * @fileoverview Main API Class
+ * @author Kai Cataldo
```
kaicataldo, line 3: @mysticatea Feel free to change this to your name.

lib/eslint/eslint.js
```diff
@@ -0,0 +1,665 @@
...
+    }
+}
+
+/**
+ * Validates and normalizes options for the wrapped CLIEngine instance.
+ * @param {ESLintOptions} options The options to process.
+ * @returns {ESLintOptions} The normalized options.
+ */
+function processOptions({
+    allowInlineConfig = true, // ← we cannot use `overrideConfig.noInlineConfig` instaed because `allowInlineConfig` has side-effect that suppress warnings that show inline configs are ignored.
```
kaicataldo, line 149:
```suggestion
    allowInlineConfig = true, // ← we cannot use `overrideConfig.noInlineConfig` instead because `allowInlineConfig` has side-effect that suppress warnings that show inline configs are ignored.
```

mysticatea:
I reverted the `source` property change because the deprecation notice is about `LintMessage` type rather than `LintResult` type.

docs/developer-guide/nodejs-api.md
```diff
@@ -17,22 +31,363 @@ While ESLint is designed to be run on the command line, it's possible to use ESL
...
+* [CLIEngine (deprecated)](#cliengine)
 * [RuleTester](#ruletester)
     * [Customizing RuleTester](#customizing-ruletester)
 * [Deprecated APIs](#deprecated-apis)
 
+---
+
+## ESLint class
+
+The `ESLint` class is the facade class to use ESLint from Node.js application.
```
nzakas, line 43:
```suggestion
The `ESLint` class is primary class to use in Node.js applications.
```

docs/developer-guide/nodejs-api.md
```diff
@@ -17,22 +31,363 @@ While ESLint is designed to be run on the command line, it's possible to use ESL
...
+
+This class pretty depends on the file system, so you cannot use it on browsers. If you want to lint code on browsers, use the [Linter](#linter) class instead.
```
nzakas, line 45:
```suggestion
This class depends on the Node.js `fs` module and the file system, so you cannot use it in browsers. If you want to lint code on browsers, use the [Linter](#linter) class instead.
```

docs/developer-guide/nodejs-api.md
```diff
@@ -17,22 +31,363 @@ While ESLint is designed to be run on the command line, it's possible to use ESL
...
+
+For a simple example:
```
nzakas, line 47:
```suggestion
Here's a simple example of using the `ESLint` class:
```

docs/developer-guide/nodejs-api.md
```diff
@@ -17,22 +31,363 @@ While ESLint is designed to be run on the command line, it's possible to use ESL
...
+
+  // 4. Output it.
+  console.log(resultText);
+})().catch((error) => {
+  process.exitCode = 1;
+  console.error(error);
+});
+```
+
+For a simple example with autofix:
```
nzakas, line 71:
```suggestion
And here is an example that autofixes lint problems:
```

docs/developer-guide/nodejs-api.md
```diff
@@ -17,22 +31,363 @@ While ESLint is designed to be run on the command line, it's possible to use ESL
...
+  Default is `true`. If `false` is present, ESLint doesn't load configuration files (`.eslintrc.*` files). Only the configuration of the constructor options is valid.
+
+##### Autofix
+
+* `options.fix` (`boolean | (message: LintMessage) => boolean`)<br>
+  Default is `false`. If `true` is present, the [`eslint.lintFiles()`][eslint-lintfiles] and [`eslint.lintText()`][eslint-linttext] methods work in autofix mode. If a function is present, the methods pass each lint message to the function, then use only the lint messages that the function returned `true`.
+* `options.fixTypes` (`("problem" | "suggestion" | "layout")[] | null`)<br>
+  Default is `null`. The types of the rules that the [`eslint.lintFiles()`][eslint-lintfiles] and [`eslint.lintText()`][eslint-linttext] methods use for autofix.
+
+##### Misc
```
nzakas, line 153:
```suggestion
##### Cache-related
```

docs/developer-guide/nodejs-api.md
```diff
@@ -17,22 +31,363 @@ While ESLint is designed to be run on the command line, it's possible to use ESL
...
+* `options.cacheLocation` (`string`)<br>
+  Default is `.eslintcache`. The [`eslint.lintFiles()`][eslint-lintfiles] method writes caches into this file.
+
+### ◆ eslint.lintFiles(patterns)
+
+```js
+const results = await eslint.lintFiles(patterns);
+```
+
+This method checks the given files then returns the results.
```
nzakas, line 166:
```suggestion
This method lints the files that match the glob patterns and then returns the results.
```

docs/developer-guide/nodejs-api.md
```diff
@@ -17,22 +31,363 @@ While ESLint is designed to be run on the command line, it's possible to use ESL
...
+* (`Promise<LintResult[]>`)<br>
+  The promise that will be fulfilled with an array of [LintResult] objects.
+
+### ◆ eslint.lintText(code, options)
+
+```js
+const results = await eslint.lintText(code, options);
+```
+
+This method checks the given source code text then returns the results.
```
nzakas, line 184:
```suggestion
This method lints the given source code text and then returns the results.
```

docs/developer-guide/nodejs-api.md
```diff
@@ -17,22 +31,363 @@ While ESLint is designed to be run on the command line, it's possible to use ESL
...
+
+If you want to use suitable configuration files to lint the text, passing the `options.filePath` option is important. This method loads configuration files with the same manner as the [`eslint.lintFiles()`][eslint-lintfiles] method checks the file the `options.filePath` option pointers. If the `options.filePath` option is not present, this method uses the configuration file at the current working directory (the `cwd` constructor option).
```
nzakas, line 186:
```suggestion
If you want to use configuration files while linting the text, pass the `options.filePath` option. This method loads configuration files in the same manner as the [`eslint.lintFiles()`][eslint-lintfiles] method. If the `options.filePath` option is not present, this method uses the configuration file at the current working directory (the `cwd` constructor option).
```

docs/developer-guide/nodejs-api.md
```diff
@@ -17,22 +31,363 @@ While ESLint is designed to be run on the command line, it's possible to use ESL
...
+
+If the given `options.filePath` option value is an ignored path by your configurations, this method returns an empty array. If the `options.warnIgnored` option is set along with the `options.filePath` option, this method returns a [LintResult] object at least. In that case, the result may contain a warning that indicates the file was ignored.
```
nzakas, line 188:
```suggestion
If the `options.filePath` value is configured to be ignored, this method returns an empty array. If the `options.warnIgnored` option is set along with the `options.filePath` option, this method returns a [LintResult] object. In that case, the result may contain a warning that indicates the file was ignored.
```

docs/developer-guide/nodejs-api.md
```diff
@@ -17,22 +31,363 @@ While ESLint is designed to be run on the command line, it's possible to use ESL
...
+
+### ◆ ESLint.getErrorResults(results)
+
+```js
+const filteredResults = ESLint.getErrorResults(results);
+```
+
+This method copies the given results and removes warnings. The returned value contains only errors.
+
+This is not instance method.
```
kaicataldo, line 313:
```suggestion
This is not an instance method.
```
or 

```suggestion
This is a static method.
```

docs/developer-guide/nodejs-api.md
```diff
@@ -17,22 +31,363 @@ While ESLint is designed to be run on the command line, it's possible to use ESL
...
+  The [LintResult] objects to filter.
+
+#### Return Value
+
+* (`LintResult[]`)<br>
+  The filtered [LintResult] objects.
+
+### ◆ LintResult type
+
+The `LintResult` value is the information of the linting result of each file. The [`eslint.lintFiles()`][eslint-lintfiles] and [`eslint.lintText()`][eslint-linttext] methods return it. It has following properties:
```
kaicataldo, line 327:
```suggestion
The `LintResult` value is the information of the linting result of each file. The [`eslint.lintFiles()`][eslint-lintfiles] and [`eslint.lintText()`][eslint-linttext] methods return it. It has the following properties:
```

docs/developer-guide/nodejs-api.md
```diff
@@ -17,22 +31,363 @@ While ESLint is designed to be run on the command line, it's possible to use ESL
...
+
+* `filePath` (`string`)<br>
+  The absolute path to the file of this result. This may be the string `"<text>"` rather than a file path if the [`eslint.lintText()`][eslint-linttext] method returned it.
```
kaicataldo, line 330: Can we add an explanation of what `"<text>"` means (i.e. that it's a placeholder)? Or maybe link to some existing documentation if we already have some?

docs/developer-guide/nodejs-api.md
```diff
@@ -17,22 +31,363 @@ While ESLint is designed to be run on the command line, it's possible to use ESL
...
+* `output` (`string | undefined`)<br>
+  The modified source code text. This property is undefined if any fixable messages didn't exist.
+* `source` (`string | undefined`)<br>
+  The original source code text. This property is undefined if any messages didn't exist or the `output` property exists.
+* `usedDeprecatedRules` (`{ ruleId: string; replacedBy: string[] }[]`)<br>
+  The information about the deprecated rules that were used to check this file.
+
+### ◆ LintMessage type
+
+The `LintMessage` value is the information of each linting error. The `messages` property of the [LintResult] type contains it. It has following properties:
```
kaicataldo, line 350:
```suggestion
The `LintMessage` value is the information of each linting error. The `messages` property of the [LintResult] type contains it. It has the following properties:
```

lib/eslint/eslint.js
```diff
@@ -0,0 +1,656 @@
...
+    extensions = null, // ← should be null by default because if it's an array then it suppresses RFC20 feature.
+    fix = false,
+    fixTypes = null, // ← should be null by default because if it's an array then it suppresses rules that don't have the `meta.type` property.
+    globInputPaths = true,
+    ignore = true,
+    ignorePath = null, // ← should be null by default because if it's a string then it may throw ENOENT.
+    overrideConfig = null,
+    overrideConfigFile = null,
+    plugins = {},
+    reportUnusedDisableDirectives = null, // ← should be null by default because if it's a string then it overrides the 'reportUnusedDisableDirectives' setting in config files. And we cannot use `overrideConfig.reportUnusedDisableDirectives` instaed because we cannot configure the `error` severity with that.
```
kaicataldo, line 165:
```suggestion
    reportUnusedDisableDirectives = null, // ← should be null by default because if it's a string then it overrides the 'reportUnusedDisableDirectives' setting in config files. And we cannot use `overrideConfig.reportUnusedDisableDirectives` instead because we cannot configure the `error` severity with that.
```

lib/eslint/eslint.js
```diff
@@ -0,0 +1,656 @@
...
+        ignorePath,
+        reportUnusedDisableDirectives,
+        resolvePluginsRelativeTo,
+        rulePaths,
+        useEslintrc
+    };
+}
+
+/**
+ * Check if a value has one or more properties that that value is not undefined.
```
kaicataldo, line 298:
```suggestion
 * Check if a value has one or more properties and that value is not undefined.
```

tests/lib/cli.js
```diff
@@ -6,7 +6,7 @@
 "use strict";
 
 /*
- * NOTE: If you are adding new tests for cli.js, use verifyCLIEngineOpts(). The
+ * NOTE: If you are adding new tests for cli.js, use verifyESLintOpts(). The
  * test only needs to verify that CLIEngine receives the correct opts.
```
kaicataldo, line 10:
```suggestion
 * test only needs to verify that ESLint receives the correct opts.
```

tests/lib/eslint/eslint.js
```diff
@@ -0,0 +1,6089 @@
...
+                    configFile: "",
+                    envs: [],
+                    globals: [],
+                    ignorePattern: [],
+                    parser: "",
+                    parserOptions: {},
+                    rules: {},
+                    plugins: []
+                }),
+                new RegExp(escapeStringRegExp([
```
kaicataldo, line 145: Appreciate these error messages ❤️

lib/eslint/eslint.js
```diff
@@ -0,0 +1,656 @@
...
+    if (typeof overrideConfig !== "object") {
+        errors.push("'overrideConfig' must be an object or null.");
+    }
+    if (!isNonEmptyString(overrideConfigFile) && overrideConfigFile !== null) {
+        errors.push("'overrideConfigFile' must be a non-empty string or null.");
+    }
+    if (typeof plugins !== "object") {
+        errors.push("'plugins' must be an object or null.");
+    } else if (plugins !== null && Object.keys(plugins).includes("")) {
+        errors.push("'plugins' must not include the empty string in keys.");
```
kaicataldo, line 246:
```suggestion
        errors.push("'plugins' must not include an empty string.");
```

docs/developer-guide/nodejs-api.md
mysticatea: > If you want to use configuration files while linting the text, pass the `options.filePath` option.

Sounds like it doesn't use configuration files if the `options.filePath` option is not present. But actually it uses configuration files on the cwd in that case.

docs/developer-guide/nodejs-api.md
```diff
@@ -17,22 +31,363 @@ While ESLint is designed to be run on the command line, it's possible to use ESL
...
+* (`Promise<boolean>`)<br>
+  The promise that will be fulfilled with whether the file is ignored or not. If the file is ignored then `true`.
+
+### ◆ eslint.loadFormatter(nameOrPath)
+
+```js
+const formatter = await eslint.loadFormatter(nameOrPath);
+```
+
+This method loads a formatter.
```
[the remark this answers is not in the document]
mysticatea: Hmm. I'm not sure what is the proper page, so I added more descriptions.

docs/developer-guide/nodejs-api.md
mysticatea: I added a description.

mysticatea:
Thank you for your review! I applied the suggestions.

mysticatea:
Thank you for your review! I applied suggestions to this PR.

kaicataldo:
@mysticatea Thank you again for picking this up while I was sick ❤️

PR #12953, opened 2020-02-22
Fix markdown for the issue and PR templates

papandreou:
So that the checkboxes are rendered correctly.

#### Prerequisites checklist

- [ ] Documentation update
- [ ] Add a CLI option
- [x] Other, please explain:

When opening another PR I noticed that a bunch of lines starting with `[ ]` that came from the github PR template weren't properly rendered as a list of checkboxes. I changed the templates so they do :hugs:

### DIFF
--- a/.github/ISSUE_TEMPLATE/NEW_RULE.md
+++ b/.github/ISSUE_TEMPLATE/NEW_RULE.md
@@ -24,10 +24,10 @@ assignees: ''
 
 **What category of rule is this? (place an "X" next to just one item)**
 
-[ ] Warns about a potential error (problem)
-[ ] Suggests an alternate way of doing something (suggestion)
-[ ] Enforces code style (layout)
-[ ] Other (please specify:)
+- [ ] Warns about a potential error (problem)
+- [ ] Suggests an alternate way of doing something (suggestion)
+- [ ] Enforces code style (layout)
+- [ ] Other (please specify:)
 
 **Provide 2-3 code examples that this rule will warn about:**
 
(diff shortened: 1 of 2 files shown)


kaicataldo:
Thanks for contributing. The current template is that way by design, because if they were actual checkboxes the PR list view would show 1/x checkboxes checked. It would be nice if we had some way of creating radio buttons that only allowed one to be selected, but markdown unfortunately doesn’t support that.

papandreou:
Haha, okay, that's fair enough. It seems like contributors are confused enough that they only follow the instructions half of the time, though. I went through the first 4 pages of PRs and counted:

* 47 left the list as-is, added one `X` in one of the boxes :ok_hand:
* 39 edited the list down to just one item with `[X]` in front of it
* 5 left the list as-is, added `X` in two of the boxes
* 5 edited the markdown so it's a list of checkboxes, then added one `X` (this includes my PRs)
* 1 edited the list down to one item without `[X]`
* 1 edited the list down to two items with `[X]` in front
* 1 left the list as-is without adding any `X` at all :thinking:
* 1 removed the list altogether :see_no_evil:

I guess it doesn't really matter, actually. But I'll maintain that it does look like a mistake in the templates when you know the gh flavored markdown syntax for a checkbox :smirk_cat:

kaicataldo:
Agreed. I think there are two issues at play:

1. The one you mentioned, where it can easily seem like a typo if you’re used to writing lists in markdown.
2. We as a team haven’t been very good at enforcing these rules.

It definitely would be nice to figure out a better system. In my personal experience, templates in general don’t seem to be very effective in issues or PRs (though they’re better than nothing!).

kaicataldo:
If you have a suggestion for an alternative system, I’d love to hear your idea!

nzakas:
Agree with @kaicataldo that the template format is intentional. I’d suggest opening an issue if you have other ideas about how to improve the template.

PR #13002, opened 2020-03-05
Update: add checking static value in no-throw-literal

yeonjuan:
#### Prerequisites checklist

[ ] Documentation update
[ ] Add autofixing to a rule
[ ] Add a CLI option
[ ] Add something to the core
[ ] Other, please explain:

I added static value checking in no-throw-literal using eslint-utils -getStaticValue.

It will generate more warning so I labeled it `Update`.

> When it was first created, it only prevented literals from being thrown (hence the name), but it has now been expanded to only allow expressions which have a possibility of being an Error object.

This rule's name is `no-throw-literal` which pretends to check only literal. But, as mentioned in the docs, I think this change can be accepted.

Also, this PR contains small [jsdoc fix](https://github.com/eslint/eslint/compare/updateNoThrowLiteral?expand=1#diff-60bc8a6f0c57ce4fbfa4b794e7efc7d3R1287) in ast-utils

### DIFF
--- a/lib/rules/no-throw-literal.js
+++ b/lib/rules/no-throw-literal.js
@@ -6,6 +6,7 @@
 "use strict";
 
 const astUtils = require("./utils/ast-utils");
+const { getStaticValue } = require("eslint-utils");
 
 //------------------------------------------------------------------------------
 // Rule Definition
@@ -35,7 +36,9 @@ module.exports = {
         return {
 
             ThrowStatement(node) {
-                if (!astUtils.couldBeError(node.argument)) {
+                const isStaticValue = node.argument && getStaticValue(node.argument, context.getScope());
+
+                if (!astUtils.couldBeError(node.argument) || isStaticValue) {
                     context.report({ node, messageId: "object" });
                 } else if (node.argument.type === "Identifier") {
                     if (node.argument.name === "undefined") {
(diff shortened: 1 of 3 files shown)


lib/rules/no-throw-literal.js
```diff
@@ -35,7 +36,9 @@ module.exports = {
         return {
 
             ThrowStatement(node) {
-                if (!astUtils.couldBeError(node.argument)) {
+                const isStaticValue = node.argument && getStaticValue(node.argument, context.getScope());
```
yeonjuan: Maybe checking whether `node.arguement` is falsy is useless?
```
throw; // Uncaught SyntaxError: Unexpected token ';'
```

nzakas:
I'm not sure it makes sense to add this into the rule. We don't do a lot of type checking because it's difficult to get correct, and I think only flagging this as a problem in some cases will be confusing.

In any case, I don't believe this is a bug because this was never the intended behavior.

yeonjuan:
@nzakas 
Oh, sorry for mislabeling.
>  We don't do a lot of type checking because it's difficult to get correct, and I think only flagging this as a problem in some cases will be confusing.

I understand it. Thanks for the explanation and maybe This PR should be closed

nzakas:
Let’s see if anyone else feels differently before closing.

kaicataldo:
I agree with @nzakas. Really appreciate the PR regardless, though!

aladdin-add:
closing, as it seems we have a consensus not accepted. Thanks for the working regardless!  ❤️

PR #13029, opened 2020-03-11
Update: Add ignorePattern to no-inline-comments

EdieLemoine:
Update: Add filtering to no-inline-comments and allow Webpack magic comments 

#### Prerequisites checklist

- [ ] Documentation update
- [ ] Bug fix
- [ ] New rule 
- [X] Changes an existing rule
- [ ] Add a CLI option
- [ ] Other, please explain:

no-inline-comments 

Fewer

**How will the change be implemented?**
New default behaviour. 

```js
const language = import(/* webpackChunkName: "my-chunk-name" */'./locale/language');
```

Reports error/warning: `Unexpected comment inline with code.(no-inline-comments)`

Not report on this kind of comment anymore.

I added all the [Webpack magic comments](https://webpack.js.org/api/module-methods/#magic-comments] keywords to the `astUtils.COMMENTS_IGNORE_PATTERN` regex. It looks like this now:

```js
// From this:
/^\s*(?:eslint|jshint\s+|jslint\s+|istanbul\s+|globals?\s+|exported\s+|jscs)/u
// To this:
/^\s*(?:eslint|jshint\s+|jslint\s+|istanbul\s+|globals?\s+|exported\s+|jscs|webpackInclude:|webpackExclude:|webpackChunkName:|webpackMode:|webpackPrefetch:|webpackPreload:)/u
```
This affects the following rules:
* no-inline-comments
* capitalized-comments (+ This is a big plus as it would break the keywords)
* line-comment-position (- Likely not necessary)
* lines-around-comment (- Likely not necessary)
* multiline-comment-style (- Likely not necessary)

I added a filter to the `sourceCode.getAllComments()` call in the `Program` method of the rule's definition so these comments (and the ones that were already present in `COMMENTS_IGNORE_PATTERN`) will be ignored.

Is the `astUtils.COMMENTS_IGNORE_PATTERN` is the correct place for the new exceptions? If so, I will update the docs in every place it applies.
Should this be the new default behaviour?

### DIFF
--- a/docs/rules/no-inline-comments.md
+++ b/docs/rules/no-inline-comments.md
@@ -87,3 +87,25 @@ var quux = (
     </div>
 )
 ```
+
+## Options
+
+### ignorePattern
+
+A regular expression can be provided to make this rule ignore specific comments.
+
+Examples of **correct** code for the `ignorePattern` option:
+
+```js
+/*eslint no-inline-comments: ["error", { "ignorePattern": "webpackChunkName:\\s.+" }]*/
+
+import(/* webpackChunkName: "my-chunk-name" */ './locale/en');
+```
+
+Examples of **incorrect** code for the `ignorePattern` option:
+
+```js
+/*eslint no-inline-comments: ["error", { "ignorePattern": "something" }] */
+
+var foo = 4; // other thing
+```
--- a/lib/rules/no-inline-comments.js
+++ b/lib/rules/no-inline-comments.js
@@ -21,7 +21,17 @@ module.exports = {
             url: "https://eslint.org/docs/rules/no-inline-comments"
         },
 
-        schema: [],
+        schema: [
+            {
+                type: "object",
+                properties: {
+                    ignorePattern: {
+                        type: "string"
+                    }
+                },
+                additionalProperties: false
+            }
+        ],
 
         messages: {
             unexpectedInlineComment: "Unexpected comment inline with code."
@@ -30,6 +40,12 @@ module.exports = {
 
     create(context) {
         const sourceCode = context.getSourceCode();
+        const options = context.options[0];
+        let customIgnoreRegExp;
+
+        if (options && options.ignorePattern) {
+            customIgnoreRegExp = new RegExp(options.ignorePattern, "u");
+        }
 
         /**
          * Will check that comments are not on lines starting with or ending with code
@@ -48,6 +64,11 @@ module.exports = {
 
             // Nothing on both sides
             if (isPreambleEmpty && isPostambleEmpty) {
+                return;
+            }
+
+            // Matches the ignore pattern
+            if (customIgnoreRegExp && customIgnoreRegExp.test(node.value)) {
                 return;
             }
 
@@ -80,9 +101,9 @@ module.exports = {
 
         return {
             Program() {
-                const comments = sourceCode.getAllComments();
-
-                comments.filter(token => token.type !== "Shebang").forEach(testCodeAroundComment);
+                sourceCode.getAllComments()
+                    .filter(token => token.type !== "Shebang")
+                    .forEach(testCodeAroundComment);
             }
         };
     }
--- a/tests/lib/rules/no-inline-comments.js
+++ b/tests/lib/rules/no-inline-comments.js
@@ -88,7 +88,24 @@ ruleTester.run("no-inline-comments", rule, {
                comment
             */}
             </div>
-        )`
+        )`,
+        {
+            code: "import(/* webpackChunkName: \"my-chunk-name\" */ './locale/en');",
+            options: [
+                {
+                    ignorePattern: "(?:webpackChunkName):\\s.+"
+                }
+            ],
+            parserOptions: { ecmaVersion: 2020 }
+        },
+        {
+            code: "var foo = 2; // Note: This comment is legal.",
+            options: [
+                {
+                    ignorePattern: "Note: "
+                }
+            ]
+        }
     ],
 
     invalid: [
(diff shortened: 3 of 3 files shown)


kaicataldo:
Thanks for the PR! While I think this is a use-case that would be good to support, I personally don't think this is the correct solution. The current ignored comments are all either ESLint directive comments or comments for other linters (I believe this is to ease the pain of switching to ESLint from `jshint`, `jscs`, etc.).

I would advocate for adding a new option to `no-inline-comments` that allows you to specify an array of exceptions. Maybe `ignorePattern` to match the option in [`capitalized-comments`](https://eslint.org/docs/rules/capitalized-comments#ignorepattern)?

Please note that we'll need to go through the consensus process outlined [here](https://eslint.org/docs/maintainer-guide/issues#consensus) before we can merge anything. Feel free to wait implementing anything else until we come to a decision so that you don't have to spend more time on it!

@eslint/eslint-team Thoughts?

mdjermanovic:
`line-comment-position`, `capitalized-comments` and `lines-around-comment` rules have `ignorePattern` option. 

This option already allows users to define custom patterns like those for webpack magic comments, so I don't think we should change the default pattern (`astUtils.COMMENTS_IGNORE_PATTERN`) which affects all these rules. It looks like an unnecessary breaking change.

As for the `no-inline-comments` rule, I believe it should have the same `ignorePattern` option (maybe `applyDefaultIgnorePatterns` as well). There is already an open issue #12755 for this enhancement.

kaicataldo:
@EdieLemoine Would you be interested in helping us implement the suggestion above once we come to a consensus on its inclusion?

EdieLemoine:
@kaicataldo sure. Just let me know which solution you guys prefer.

kaicataldo:
I will champion this change (as described [here](https://github.com/eslint/eslint/pull/13029#issuecomment-601504598)). Given the two :+1: on my proposal, we only need one more :+1:.

kaicataldo:
And thanks @EdieLemoine! Once we reach consensus we'll make sure it's clear here (hopefully soon since we only need one more :+1:).

kaicataldo:
@eslint/eslint-team We just need one more :+1: to accept this.

nzakas:
I've added my :+1: so marking as accepted.

mdjermanovic:
Linked #12755 as it seems that this option will fix the issue with `istanbul` comments, too.

mdjermanovic:
@kaicataldo

> I would advocate for adding a new option to `no-inline-comments` that allows you to specify an array of exceptions. Maybe `ignorePattern` to match the option in [`capitalized-comments`](https://eslint.org/docs/rules/capitalized-comments#ignorepattern)?

to clarify, should the option accept an _array_ of "regex" strings or just one string?

The same option in similar rules like `line-comment-position`, `capitalized-comments` and `lines-around-comment` accepts 1 string, which provides basically the same functionality (maybe just a bit less convenient in some cases).

kaicataldo:
My take is that we should math what those other rules are currently doing. Supporting RegExp strings would be an additive change, and we can add it to all the rules listed if we decide we want to go that route in the future.

kaicataldo:
@EdieLemoine Thanks for your patience! Is this still something you're willing to help out with?

EdieLemoine:
@kaicataldo I just updated it. Is this the proper solution?

lib/rules/no-inline-comments.js
```diff
@@ -80,9 +105,10 @@ module.exports = {
 
         return {
             Program() {
-                const comments = sourceCode.getAllComments();
-
-                comments.filter(token => token.type !== "Shebang").forEach(testCodeAroundComment);
+                return sourceCode.getAllComments()
+                    .filter(comment => !astUtils.COMMENTS_IGNORE_PATTERN.test(comment.value))
```
mdjermanovic, line 109: This shouldn't be included in this change.

tests/lib/rules/no-inline-comments.js
```diff
@@ -88,7 +88,24 @@ ruleTester.run("no-inline-comments", rule, {
...
-        )`
+        )`,
+
+        // Webpack
+        {
+            code: `
+              import(
+                /* webpackInclude: /\\.json$/ */
+                /* webpackChunkName: "my-chunk-name" */
+                './locale/en'
```
mdjermanovic, line 99: This example would be valid even without the pattern, since the comments are on separate lines?

tests/lib/rules/no-inline-comments.js
```diff
@@ -88,7 +88,24 @@ ruleTester.run("no-inline-comments", rule, {
...
+              );
+            `,
+            options: [
+                {
+                    ignorePattern: "(?:webpackChunkName|webpackInclude):\\s.+"
+                }
+            ],
+            parserOptions: { ecmaVersion: 2020 }
```
mdjermanovic, line 107: Can we also add some `invalid` examples where the comment doesn't match the configured pattern

tests/lib/rules/no-inline-comments.js
EdieLemoine: Now that's silly 🤦‍♀️

EdieLemoine:
@mdjermanovic I updated the branch.

docs/rules/no-inline-comments.md
```diff
@@ -36,6 +36,26 @@ var bar = 5;
...
 ```
 
+### ignorePattern
+
+A regular expression can be provided to make this rule ignore specific comments.
+
+Examples of **correct** code for the `ignorePattern` option:
+
+```js
+/*eslint no-inline-comments: ["error", { "ignorePattern": "webpackChunkName:\s.+" }]*/
```
mdjermanovic, line 46: Due to recent changes in `levn`, which we use to parse config values in comments, this should be `\\s` like in string literals:

```js
/*eslint no-inline-comments: ["error", { "ignorePattern": "webpackChunkName:\\s.+" }]*/
```

EdieLemoine:
@mdjermanovic I updated it again :)

kaicataldo (dismissed):
This looks really great - thanks! One small comment about clarifying the docs, but otherwise LGTM.

docs/rules/no-inline-comments.md
```diff
@@ -87,3 +87,25 @@ var quux = (
     </div>
 )
 ```
+
+## Options
+
+### ignorePattern
+
+A regular expression can be provided to make this rule ignore specific comments.
```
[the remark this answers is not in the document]
EdieLemoine: Hmm, it's implemented the same way as in lines-around-comment. It's referenced to as a "regular expression" there as well.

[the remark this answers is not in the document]
btmills: That's a good point, @EdieLemoine. Since that's the only change @kaicataldo requested, let's merge this to get it into today's release, and I'll put together a quick PR to change this in both places.

btmills (approved):
Thanks for contributing, @EdieLemoine! I'll update the docs after merging, and this will be in today's release.

PR #13054, opened 2020-03-16
Docs: Clarify that node_modules is ignored by default

### LINKED ISSUE #13006: Documentation bug: Re: linting `node_modules`

6.8.0

Per https://eslint.org/docs/user-guide/command-line-interface , there is an implied need to ignore `node_modules` (which seems no longer to be the case per #1163):

1. Here's one:

> Here's a simple example of a .eslintignore file:
>  node_modules/*

2. Also on https://eslint.org/docs/user-guide/configuring , the need to ignore `node_modules` seems implicit here as well:

> "ignorePatterns": ["temp.js", "node_modules/"],

3. While the following implies the opposite, I think the text could be clarified here a bit too:

> In addition to any patterns in a .eslintignore file, ESLint always ignores files in /node_modules/* and /bower_components/*.
> ...
> \# /node_modules/* and /bower_components/* in the project root are ignored by default

The first line is confusing as it seems to suggest one cannot get linting of `node_modules` or `bower_components` yet it is possible to reenable both of them with `!`.

1. Remove `node_modules` from the examples unless to point out that one can undo the default (it may be helpful to mention that `!.eslintrc.js` (or `!.*.js`) can be added back too, at least until eslint 7 may change this behavior for `!.eslintrc.js`).
2. For the third case, I think it should instead say in the first line "ESLint always ignores files by default in ..." since I am (thankfully!) able to reapply linting of `node_modules` (or `bower_components`) by adding `!node_modules/**` to an ignore (within an `.eslintignore` file or by a command such as `$(npm bin)/eslint --no-inline-config --ignore-pattern='!node_modules/**' .`). (I truly hope this functionality will never be dropped, btw!)

Not at this time.

MKuijpers:
Fixes #13006

#### Prerequisites checklist

[X] Documentation update
[ ] Add autofixing to a rule
[ ] Add a CLI option
[ ] Add something to the core
[ ] Other, please explain:

The ```node_modules``` folder is ignored by default. However, in some parts of the documentation it is still included in ignore configs. I removed the ```node_modules``` entries from the examples and clarified in an example ignore config how it can be added back in.

Whether the added/edited documentation is still correct and if ```node_modules/*``` is the correct pattern.

### DIFF
--- a/docs/user-guide/command-line-interface.md
+++ b/docs/user-guide/command-line-interface.md
@@ -480,7 +480,7 @@ Example:
 
 ESLint supports `.eslintignore` files to exclude files from the linting process when ESLint operates on a directory. Files given as individual CLI arguments will be exempt from exclusion. The `.eslintignore` file is a plain text file containing one pattern per line. It can be located in any of the target directory's ancestors; it will affect files in its containing directory as well as all sub-directories. Here's a simple example of a `.eslintignore` file:
 
-    node_modules/*
+    temp.js
     **/vendor/*.js
 
 A more detailed breakdown of supported patterns and directories ESLint ignores by default can be found in [Configuring ESLint](configuring.md#ignoring-files-and-directories).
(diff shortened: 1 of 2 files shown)


kaicataldo:
Thanks for doing this! Agreed that this is an improvement.

PR #13068, opened 2020-03-20
New: add ls-lint

loeffel-io:
#### Prerequisites checklist

[x] Other, please explain:

Already in use at [Vue.js 3](https://github.com/vuejs/vue-next) - Adding [ls-lint](https://github.com/loeffel-io/ls-lint) provides a clean project structure/filesystem for the eslint project for all updates and contributions.

```yaml
ls:
    '{bin,conf,lib,messages,packages,templates,tools}':
        .js: kebab-case
        .json: kebab-case
        .htm: kebab-case
        .html: kebab-case
        .txt: kebab-case
```

With this config ls-lint checks all your filenames and directory names inside or behind these directories `bin,conf,lib,messages,packages,templates,tools`

Would love to see this in eslint ❤️

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,6 +14,8 @@ jobs:
     - uses: actions/setup-node@v1
     - name: Install Packages
       run: npm install
+    - name: Lint Filesystem
+      run: node Makefile lsLint
     - name: Lint Files
       run: node Makefile lint
     - name: Check Rule Files
--- a/.ls-lint.yml
+++ b/.ls-lint.yml
@@ -1,0 +1,7 @@
+ls:
+    '{bin,conf,lib,messages,packages,templates,tools}':
+        .js: kebab-case
+        .json: kebab-case
+        .htm: kebab-case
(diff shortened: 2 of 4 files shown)


nzakas:
I’m sorry, your project is only a week old, so I don’t think it’s something we should adopt at this point in time. 

In the future, I’d suggest you mention that you are the author of the tool you’re recommending. It’s good practice to be upfront if you are promoting your own project (there’s nothing wrong with doing so, but letting people know that’s what you’re doing is a good idea).

kaicataldo:
Agreed with @nzakas. While I see the value in this, I think we would want to wait until the tool is more mature.

loeffel-io:
Thank you for your answer, i really appreciate it!

I can understand that my project is, for now, too fresh to merge - I'll keep you updated 👍 

`just fyi: the npm package is only a week old, the project itself started a month ago, is stable and now also a part of` [vant](https://github.com/youzan/vant)

kaicataldo:
Sounds good.

PR #13073, opened 2020-03-23
Update: new option singleline-if in padding-line-between-statements

npdev453:
#### Prerequisites checklist

[x] Documentation update

New option to allow specify preferences for padding lines between single-line if's

impr

padding-line-between-statements

No, only with special configuration with 'singleline-if' 

            { blankLine: "never", prev: ["singleline-if"], next: ["if"]},

```js
/*eslint padding-line-between-statements: [
    "error",
    { blankLine: "never", prev: ["singleline-if"], next: ["singleline-if"]},
    { blankLine: "always", prev: ["singleline-if"], next: ["if"]},
]*/

/* correct */
function foo(error, a, b)
   if (error) return bar(error);
   if (!a) return bar(a);
   if (!b) return bar(b);

   if (a > b) {
      boo(a);
   }
}

/* incorrect */
function foo(error, a, b)
   if (error) return bar(error);

   if (!a) return bar(a);

   if (!b) return bar(b);
   if (a > b) {
      boo(a);
   }
}

```

There is no way now to specify rule config for that case, only always/never spaces between all if's

Same things but only will have additional option.

### DIFF
--- a/docs/rules/padding-line-between-statements.md
+++ b/docs/rules/padding-line-between-statements.md
@@ -76,6 +76,7 @@ You can supply any number of configurations. If a statement pair matches multipl
     - `"singleline-const"` is single-line `const` variable declarations.
     - `"singleline-let"` is single-line `let` variable declarations.
     - `"singleline-var"` is single-line `var` variable declarations.
+    - `"singleline-if"` is single-line `if` statements.
     - `"switch"` is `switch` statements.
     - `"throw"` is `throw` statements.
     - `"try"` is `try` statements.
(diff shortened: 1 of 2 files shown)


npdev453:
At clean master brach test crash too with same reasons

kaicataldo:
Please fill out [this template](https://eslint.org/docs/developer-guide/contributing/rule-changes) so we can better evaluate your proposal.

If this is accepted, we'll need to add tests and some examples for the documentation.

npdev453:
@kaicataldo, done.

kaicataldo:
Thank you for filling out the template. This sounds like a good addition to me :+1:. Let's see what the rest of the team thinks.

npdev453:
@kaicataldo can you remove tags?

kaicataldo:
Thanks for the ping - sorry I forgot to do this when I commented!

kaicataldo:

Issue #13077, opened 2020-03-24
[Question] Turn off all rules, specifically from plugins


himynameisdave:
Probably a bit of a weird use case here, but I am looking to create a separate ESLint configuration for doing pre-commit checks, since in our pre-commit hook we only want to lint a (much) smaller sub-set of core ESLint rules.

I figured out how to get a separate config running for the pre-commit command (which involved moving the ESLint configs into their own directory so as [to avoid any weird cascading/hierarchy](https://eslint.org/docs/user-guide/configuring#configuration-cascading-and-hierarchy) issues). Then I just targeted them with the `-c` flag:

**`package.json`**
```json
  "scripts": {
    "lint": "eslint --ext .ts,.js,.jsx,.tsx -c ./lint/.eslintrc.frontend.js .",
    "lint:pre-commit": "eslint --ext .ts,.js,.jsx,.tsx -c ./lint/.eslintrc.precommit.js ."
  }
```

For the paired-down pre-commit config, I borrowed a bit [from the `eslint:all` config](https://github.com/eslint/eslint/blob/master/conf/eslint-all.js#L20-L24), which grabs all the rules and creates an object like `{ [ruleId]: 'off' }` (with a key/value pair for each rule).

It looks a little something like this:

**`.eslintrc.pre-commit.js`**

```js
const allRules = require('../node_modules/eslint/lib/rules');

// These are the only rules we want to flag on pre-commit:
const ERROR_RULES = ['no-undef'];

const rules = Object.entries(allRules).reduce((accRules, [ruleId, rule]) => {
  return {
    ...accRules,
    // Only set specified rules to 'error', all others are 'off'
    [ruleId]: ERROR_RULES.includes(ruleId)
      ? 'error'
      : 'off',
  };
}, {});

module.exports = {
  parser: '@typescript-eslint/parser',
  parserOptions: {
    tsconfigRootDir: __dirname + '../',
    project: './tsconfig.json',
    ecmaFeatures: {
      'jsx': true,
    },
    sourceType: 'module',
  },
  rules,
};
```

No plugins, it just turns off all the core ESLint rules. Uses the TS parser but not the TS plugin. 👍 

☝️ This got me 90% of the way to where I needed to be, but I've hit a roadblock because throughout our codebase we use disable directive comments to disable rules from specific plugins. Here's an example:

```tsx
// eslint-disable-next-line react/no-array-index-key
<option key={i} value={option}>
```

[...]

PR #13079, opened 2020-03-24
New: Add a config:off for disabling all rules

himynameisdave:
#### Prerequisites checklist

[ ] Documentation update
[ ] Add autofixing to a rule
[ ] Add a CLI option
[ ] Add something to the core
[x] Other, please explain:

Adds a config which is similar to `conf/all`, but instead of turning all the rules on, it turns them off. Although there are not a lot of use cases for this, my motivation behind it stems from a question I had about turning off all rules from plugins, where I needed to do this exact thing (see #13077).

Not sure if the ESLint team is interested in moving forward with this, as I understand that this might not be useful for many people, and could lead to some confusion or hesitation. I know that I personally have had a few cases where I thought this would be nice. Just my opinion, but I think that all ESLint plugins and configs _should_ offer an `off` config as an escape hatch (which is why [my personal config does this](https://github.com/himynameisdave/eslint-config-himynameisdave#off)).

- 📄 Added a new config, similar to `eslint:all`, which turns everything off instead.
- 📖 Added some brief documentation for it.
- 🔬 Ensured I had testing in place for it.

My main concern is _is this even worth it?_ To me, I think it's useful, but I am totally open to hearing feedback about if this is something that would benefit the project. Totally okay if this PR just has to die, I just thought I should try and see what people think! 🙂


nzakas:
Thanks for the suggestion. While I can see a use case for plugins doing this, I’m afraid I don’t see a use case for turning all core rules off given that they all start by being off.

himynameisdave:
@nzakas I totally understand that. I guess the one use case that stands above the rest for having a config which turns everything off would be if you wanted to only lint for one specific rule. So you might imagine an `.eslintrc` like this...

```json
{
  "extends": [
    "eslint:off"
  ],
  "rules": {
    "no-undef": "error"
  }
}
```

Again, I understand that this is a bit of a narrow use case, so I'm fine closing this if you/the team don't agree that it's useful.

ljharb:
@nzakas the very critical and imo common use case in large codebases is, "i have a shared config that sets up all my parser and env options, and plugin settings, and extends/plugins, and i want to run eslint with only a single rule enabled". Having any way to disable all rules, and then add `--rule`, solves that.

nzakas:
@ljharb you can do that with the `—no-eslintrc` flag coupled with the `—rule`.

Also note that this proposal would only turn off core rules, not plugin rules, which you’d still need to disable separately.

ljharb:
@nzakas no, i can't, because i do need the eslintrc - every plugin, parserOption, `settings`, etc - the only thing i don't want from eslintrc is `rules`.

Fair enough that this proposal may not solve it, but the use case for what I'm describing is critical.

himynameisdave:
@ljharb && @nzakas yeah I think that, while this would be a nice addition, it's not strictly necessary due to what @nzakas mentioned about disabling the config (`--no-eslintrc`) and then specifying `--rule` for the rule you want to add. As for the settings and parser options etc., that _could_ also be specified via the command line (although it might be a bit verbose to do so).

ljharb:
It’s not practical to figure out every needed parser setting nor every active rule so as to construct a long and brittle command line invocation. A “set every rule to off” command line flag, however, would suffice.

kaicataldo:
Given that we have `eslint:all`, it does feel like this is in line with the current feature set. I'm not against this as long as we make it clear that adding a configuration that configures all rules to `"off"` is the extent of the feature (for the purposes of keeping maintenance overhead low).

himynameisdave:
@kaicataldo `&&` @ljharb `&&` @nzakas I've resolved the merge conflicts and interested in moving this forward (if we agree that it is a useful addition). If we want to move forward, I'd be interested in getting more feedback about my implementation.

nzakas:
We would still need the TSC to approve this proposal before moving forward.

kaicataldo:
**TSC Summary**

This proposal introduces the idea of an internal `eslint:off` configuration to extend off of. Practically speaking, this is the opposite of `eslint:all`. 

This would allow for end users to run one rule (which is useful for cases like introducing a new rule and autofixing it) at a time while keeping the rest of their configuration (parsers, plugins. etc.) intact. There are workarounds to this, but they all involve figuring out the rest of the config and either passing it to ESLint via CLI flags or through a separate configuration file.

**TSC Question**

Do we want to accept this proposal?

btmills:
We discussed this in today's TSC meeting. Based on the comments in this thread and thanks to some real-time clarification by @ljharb, two desires for this feature are that it be usable from the CLI and that it turn off plugin rules. While `eslint:none` nicely complements `eslint:all`, it only turns off core rules, and as a shared config, it couldn't be extended via a CLI flag. To make sure we get this feature right, we'd like to see an [RFC](https://github.com/eslint/rfcs/) that describes the use case, the proposed solution, and outlines a possible implementation.

For more context, the meeting notes are at https://github.com/eslint/tsc-meetings/pull/176, and the full transcript of today's discussion is at https://gitter.im/eslint/tsc-meetings/archives/2020-05-07. This was the first item on the agenda, so you won't have to scroll to find it!

---

Speaking for myself now rather than the team. @himynameisdave if you want to pursue this with a full proposal in an RFC, a static shared config won't be able to turn off plugin rules, so I'd take a look at doing it as a CLI option as a next step. It might also be helpful to more fully describe the desired workflow (and it sounds like @ljharb has ideas here) to make sure the RFC addresses all of them. For example, should this setting be able to affect editor integrations too?

nzakas:
Closing to clear this from our plate. If anyone would like to pursue something similar to this proposal, please open an issue to discuss further.

PR #13140, opened 2020-04-04
Upgrade: levn@0.4.1 (fixes #9366)

### LINKED ISSUE #9366: Escape characters are handled incorrectly in inline config comments

* **ESLint Version:** v4.7.2
* **Node Version:** v8.5.0
* **npm Version:** Yarn v1.0.2

Default

<details>

```yml
rules:
    id-match: [2, "^(([^$\\W]|\\$[a-f\\d]{2})+|[$_]\\w*|[^\\W\\d]\\w*|[A-Z]([A-Z_]*[A-Z])?)$", {
        properties: true
    }]
```

</details>

```js
/* eslint id-match: [2, "^(([^$\\W]|\\$[a-f\\d]{2})+|[$_]\\w*|[^\\W\\d]\\w*|[A-Z]([A-Z_]*[A-Z])?)$", {properties: true}] */

function is$2dvoid(value) {
    return value == null
}
```

**What did you expect to happen?** It to pass (as it does without the comment)

```
/path/to/file.js
  3:10  error  Identifier 'is$2dvoid' does not match the pattern '^(([^$\\W]|\\$[a-f\\d]{2})+|[$_]\\w*|[^\\W\\d]\\w*|[A-Z]([A-Z_]*[A-Z])?)$'  id-match

```

kaicataldo:
#### Prerequisites checklist

[ ] Documentation update
[ ] Add autofixing to a rule
[ ] Add a CLI option
[ ] Add something to the core
[ ] Other, please explain:

fixes #9366

This PR upgrades `levn` to `0.4.1` to fix the issue of backslashes in strings not being escaped correctly in inline directive comments.

Am I missing anything? Apologies - I know the tests are difficult to read.

### DIFF
--- a/tests/lib/linter/linter.js
+++ b/tests/lib/linter/linter.js
@@ -2734,7 +2734,7 @@ describe("Linter", () => {
 
     describe("when evaluating code with comments which have colon in its value", () => {
         const code = String.raw`
-/* eslint max-len: [2, 100, 2, {ignoreUrls: true, ignorePattern: "data:image\/|\\s*require\\s*\\(|^\\s*loader\\.lazy|-\\*-"}] */
+/* eslint max-len: [2, 100, 2, {ignoreUrls: true, ignorePattern: "data:image\\/|\\s*require\\s*\\(|^\\s*loader\\.lazy|-\\*-"}] */
 alert('test');
 `;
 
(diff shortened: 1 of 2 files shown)


tests/lib/linter/linter.js
```diff
@@ -2733,14 +2733,31 @@ describe("Linter", () => {
     });
 
     describe("when evaluating code with comments which have colon in its value", () => {
-        const code = "/* eslint max-len: [2, 100, 2, {ignoreUrls: true, ignorePattern: \"data:image\\/|\\s*require\\s*\\(|^\\s*loader\\.lazy|-\\*-\"}] */\nalert('test');";
+        const code = "/* eslint max-len: [2, 100, 2, {ignoreUrls: true, ignorePattern: \"data:image\\/|\\\\s*require\\\\s*\\\\(|^\\\\s*loader\\\\.lazy|-\\\\*-\"}] */\nalert('test');";
```
aladdin-add, line 2736: nits: can we use `String.raw` to avoid using something like `\\\\`, I think it can improve a little readability.

kaicataldo: Good idea!

gkz:
You should also update `optionator` to 0.9.1 - it uses `levn` internally, and this will keep the versions of `levn` used by your option parsing and comments in sync

tests/lib/linter/linter.js
```diff
@@ -2733,14 +2733,31 @@ describe("Linter", () => {
...
             assert.strictEqual(messages.length, 1);
             assert.strictEqual(messages[0].ruleId, "max-len");
-            assert.strictEqual(messages[0].message, "This line has a length of 122. Maximum allowed is 100.");
+            assert.strictEqual(messages[0].message, "This line has a length of 128. Maximum allowed is 100.");
+            assert.include(messages[0].nodeType, "Program");
+        });
+    });
+
+    describe("when evaluating code with comments that contain escape sequences", () => {
+        const code = '/* eslint max-len: ["error", 1, { ignoreComments: true, ignorePattern: "console\\.log\\\\(" }] */\nconsole.log("test");\nvar a = "test2";';
```
gkz, line 2749: I don't think the `"...console\\.log..."` does anything, since this is `console\.log`, when when passed through `levn` is just `console.log`

gkz: When you switch to using `String.raw` this will become more clear

kaicataldo: This is the issue this PR fixes. Since this string is evaluated at runtime and then used to generate a RegExp, we need to ensure that the `.` is escaped (so that it's evaluated as a literal `.` character in the RegExp).

gkz: You should add a test case like `consoleXlog("test");` and make sure it does not fire if it's anything but a dot.

I'm saying is that what you actually want is an escaped backslash `\`, so you want `\\` (under `String.raw` conditions). In JS console type `"console\.log"`, you will see that the output is `console.log`. You get this right later on when you ensure there is an escaped backlash before the `(` by doing `\\(`, so that it is escaped in the regex

kaicataldo: OH, I understand now - thanks! Updating now.

PR #13176, opened 2020-04-12
Fix: Remove default .js from --ext CLI option

btmills:
#### Prerequisites checklist

* **ESLint Version:** `7.0.0-alpha.3`
* **Node Version:** `13.11.0`
* **npm Version:** `6.13.7`

**What parser (default, Babel-ESLint, etc.) are you using?** default

<details>

```js
"use strict";

module.exports = {
    root: true,
    plugins: [
        "markdown",
        "react"
    ],
    extends: ["eslint:recommended"],
    overrides: [
        {
            files: [".eslintrc.js"],
            env: {
                node: true
            }
        },
        {
            files: ["**/*.md"],
        },
        {
            files: ["**/*.md/*.jsx"],
            env: {
                browser: true,
                es6: true
            },
            globals: {
                React: false
            },
            extends: ["plugin:react/recommended"],
            parserOptions: {
                ecmaFeatures: {
                    jsx: true
                },
                ecmaVersion: 2015,
                sourceType: "module"
            },
        }
    ]
};
```

</details>

````md
# README.md

```jsx
!@#$%^
```
````

```sh
$ eslint README.md
```

Because I included `md` and `jsx` `overrides` in `.eslintrc.js`, ESLint should have seen the syntax error in the Markdown file's JSX code block.

No errors were reported.

PR #12677, `Breaking: lint overrides files`, removed the default `.js` value from the  default CLIEngine options. However, it did not remove the same default value from the `--ext` option parsed by Optionator. Because of this, when running `eslint .` without the `--ext` option, Optionator would insert a default `--ext .js` value that would be passed on to the CLIEngine. ESLint only lints `overrides` files if the `--ext` CLI option is not passed, so the default `--ext .js` option prevented ESLint from ever reaching the `overrides` file path checking flow.

Is this a `Fix:` or a `Breaking:`? It's a bug fix because it enables behavior that was supposed to be working already, but it's also a breaking change because it removes a CLI option's default value. Since we're still in prereleases, I went with `Breaking:` since it won't prevent us from doing anything.

### DIFF
--- a/lib/options.js
+++ b/lib/options.js
@@ -46,7 +46,6 @@ module.exports = optionator({
         {
             option: "ext",
             type: "[String]",
-            default: ".js",
             description: "Specify JavaScript file extensions"
         },
         {
--- a/tests/lib/options.js
+++ b/tests/lib/options.js
@@ -79,11 +79,10 @@ describe("options", () => {
             assert.strictEqual(currentOptions.ext[1], ".js");
         });
 
-        it("should return an array one item when not passed", () => {
+        it("should not exist when not passed", () => {
             const currentOptions = options.parse("");
 
-            assert.isArray(currentOptions.ext);
-            assert.strictEqual(currentOptions.ext[0], ".js");
+            assert.notProperty(currentOptions, "ext");
         });
     });
 


ljharb:
Does this mean that without an overrides including .js files, and without an ext option, eslint won’t lint js by default?

btmills:
> Does this mean that without an overrides including .js files, and without an ext option, eslint won’t lint js by default?

In the absence of an `--ext` option, ESLint will still lint `.js` files, and it will now _also_ lint `overrides` patterns. Check out the [docs changes](docs/user-guide/configuring.md) in #12677 for the authoritative version. Instead of applying the `.js` default at the very beginning of CLI arg parsing, the new [`FileEnumerator#isTargetPath` method](https://github.com/eslint/eslint/pull/12677/files#diff-59f2daabf2bcfb88218153a9f20338bfR234-R258) lints `.js` files if the `--ext` option isn't passed.

btmills:
For more context, take a look at RFC20. This PR is just a fix for one change that was missed in #12677, which is the PR that implemented the RFC.

ljharb:
Thanks for clarifying :-) sounds great to me.

kaicataldo:
Thanks for contributing!

PR #11696, opened 2019-05-10
New: add --resolve-plugins-relative-to flag

not-an-aardvark:
**What is the purpose of this pull request? (put an "X" next to item)**

[x] Add something to the core

**What changes did you make? (Give an overview)**

With the [2018-simplified-package-loading RFC](https://github.com/eslint/rfcs/blob/8bc0b80e0b3e54d10991a4774c41f7375dfcbbfe/designs/2018-simplified-package-loading/README.md) implemented, ESLint always resolves plugins relative to the current working directory. The CWD works well for the most common use case, but is inconvenient for certain integrations. This commit proposes adding a CLI flag to specify an alternative place where plugins should be resolved from.

(Implements https://github.com/eslint/rfcs/pull/18)

**Is there anything you'd like reviewers to focus on?**

Nothing in particular

PR #13234, opened 2020-04-28
Fix: Respect --resolve-plugins-relative-to flag for parser.

Raynos:
The configuration to say that plugins should be loaded
from a relative path should also apply to the parser.

If i dont want to install my plugins locally then i dont
want to install my parser locally either.

This allows me to install both the eslint/typescript plugin as well as the eslint/typescript parser somewhere else and not in `./node_modules`.

#### Prerequisites checklist

[ ] Documentation update
[ ] Add autofixing to a rule
[ ] Add a CLI option
[ ] Add something to the core
[ ] Other, please explain:

Respect the cli `--resolve-plugins-relative-to` option when attempting the find the parser.

I made the method static because the linter complained about no usage of `this` in the method body.

### DIFF
--- a/lib/cli-engine/config-array-factory.js
+++ b/lib/cli-engine/config-array-factory.js
@@ -897,8 +897,7 @@ class ConfigArrayFactory {
     _loadParser(nameOrPath, ctx) {
         debug("Loading parser %j from %s", nameOrPath, ctx.filePath);
 
-        const { cwd } = internalSlotsMap.get(this);
-        const relativeTo = ctx.filePath || path.join(cwd, "__placeholder__.js");
+        const relativeTo = path.join(ctx.pluginBasePath, "__placeholder__.js");
 
         try {
             const filePath = ModuleResolver.resolve(nameOrPath, relativeTo);
--- a/lib/cli-engine/config-array-factory.js
+++ b/lib/cli-engine/config-array-factory.js
@@ -703,7 +703,7 @@ class ConfigArrayFactory {
         }
 
         // Load parser & plugins.
-        const parser = parserName && this._loadParser(parserName, ctx);
+        const parser = parserName && ConfigArrayFactory.loadParser(parserName, ctx);
         const plugins = pluginList && this._loadPlugins(pluginList, ctx);
 
         // Yield pseudo config data for file extension processors.
@@ -894,7 +894,7 @@ class ConfigArrayFactory {
      * @param {ConfigArrayFactoryLoadingContext} ctx The loading context.
      * @returns {DependentParser} The loaded parser.
      */
-    _loadParser(nameOrPath, ctx) {
+    static loadParser(nameOrPath, ctx) {
         debug("Loading parser %j from %s", nameOrPath, ctx.filePath);
 
         const relativeTo = path.join(ctx.pluginBasePath, "__placeholder__.js");


Raynos:
I have signed the CLA

Raynos:
This PR and fix is necessary to allow bundling `eslint` + `typescript` into a `standard` like experience.

For example https://github.com/Raynos/tsdocstandard

kaicataldo:
Looking at the [original RFC](https://github.com/eslint/rfcs/pull/18) and [PR in which it was implemented](https://github.com/eslint/eslint/pull/11696), it’s not clear to me why parsers don’t currently behave this way. It does seem like an omission to me - especially since the case outlined in [this comment](https://github.com/eslint/eslint/pull/13234#issuecomment-622143190) is specifically mentioned in the RFC. Curious if anyone else might have more insight as to why these were not included.

lib/cli-engine/config-array-factory.js
```diff
@@ -894,11 +894,10 @@ class ConfigArrayFactory {
      * @param {ConfigArrayFactoryLoadingContext} ctx The loading context.
      * @returns {DependentParser} The loaded parser.
      */
-    _loadParser(nameOrPath, ctx) {
+    static loadParser(nameOrPath, ctx) {
```
Raynos, line 897: I made the method static because the linter complained about no usage of this in the method body.

Raynos:
With the release of eslint 7.0 I am excited to use it in https://github.com/Raynos/tsdocstandard and also simplify `standard` & `standard-engine` if it supports relative behavior for plugins.

Is there someone who could look at reviewing this small change ? cc @nzakas @yeonjuan

nzakas:
@raynos your change is being reviewed, hence the comments you’re seeing. Please be patient as we all are dealing with a lot right now.

kaicataldo:
@eslint/eslint-team Can we get a few more pairs of eyes on this?

mysticatea:
- The name `--resolve-plugins-relative-to` is clear to be applied to plugins, not parsers. It's intentional and changing this behavior is a breaking change that we need major version up.
- We can use absolute/relative paths in the `parser` and `extends` properties, so we can configure ESLint to load parsers and shareable configs from arbitrary places.
  On the other hand, we cannot use absolute/relative paths in the `plugins` field. Therefore, the problem was, we couldn't reproduce the previous loading behavior of plugins. The `--resolve-plugins-relative-to` option was added to solve the problem.

Raynos:
> We can use absolute/relative paths in the parser 

Here is an example of what I am trying to do ( https://github.com/Raynos/eslint-config-tsdocstandard/blob/master/eslintrc.json#L7 ).

If I change the path to `./node_modules/@typescript-eslint/parser` will that path be relative to `process.cwd()` or will it be relative to the eslintrc.json file itself ?

mysticatea:
> If I change the path to ./node_modules/@typescript-eslint/parser will that path be relative to process.cwd() or will it be relative to the eslintrc.json file itself ?

The eslintrc.json file itself.
It's samely as `require("./node_modules/@typescript-eslint/parser")`.

Raynos:
I will try that and report back.

Raynos:
I have succesfully changed the parser to be a relative reference

https://github.com/Raynos/eslint-config-tsdocstandard/compare/v16.1.2...master

I ran into the edgecases where `./node_modules` did not exist because it was deduped into `../` instead, aka the dependency I wanted to reference existed in a node_modules up the directory tree ( a peer dependency ).

I was able to "combat" this by making the parser a devDependency and adding it to bundledDependencies.

This forces `npm` at least to install it correct in `./node_modules`.

I assume this will break `yarn` & `pnpm`.

You cannot publish an npm package that is a config JSON file with `./node_modules/@typescript-eslint/parser` as that makes broken assumptions about the layout on disk.

You might be able to publish an `eslintrc.js` and use `require.resolve` to find the absolute path to `@typescript-eslint/parser` at runtime

In the parent project ( https://github.com/Raynos/tsdocstandard/compare/v15.4.2...master ) I was able to switch away from my fork of eslint back to `6.8.0`.

Raynos:
https://github.com/Raynos/eslint-config-tsdocstandard/commit/78d647799d36cc349905e2849a43082449b9d5fc

I've made it a JS file and choose to use `require.resolve` instead. This should hopefully support `yarn` & `pnpm`

Raynos:
Can I make a pull request for the documentation instead of modifying the behavior of this CLI flag ? 

Can you recommend which page of the documentation would be useful to update.

mysticatea:
`parser: "@typescript-eslint/parser"` is samely as `require("@typescript-eslint/parser")` in the config file, therefore it looks at `./node_modules/@typescript-eslint/parser` at first, then looking for the ancestor directories if not found. Isn't that what you want?

Anyway, I agree `require.resolve()` is better than the others.
It's nice if we recommend using absolute paths in the `extends`/`parser` settings, in the https://eslint.org/docs/developer-guide/shareable-configs page.

Raynos:
`parser: "@typescript-eslint/parser"` does not do what I expect as it tries to find the module in

`my-app/node_modules` instead of `global-eslint/node_modules`. Using `require.resolve` resolves it relative to the location of the `eslintrc.js` file instead.

Raynos:
Since this turned out not to be necessary I will close this PR. 

Thanks for the alternative suggestion, super useful :+1:

chdh:
Another solution is to use the `NODE_PATH` environment variable instead of the `--resolve-plugins-relative-to` parameter.

Example for a global ESLint installation under Windows:

`set "NODE_PATH=%APPDATA%\npm\node_modules"`

ljharb:
NODE_PATH is both deprecated/discouraged and won't be available in node's ESM implementation, so it's best to avoid using or suggesting that as a workaround.

chdh:
@ljharb Do you know a better solution? The `require.resolve()` method does not work when the ESLint plugins are installed globally.

ljharb:
@chdh The best solution remains to never install eslint or any of its plugins globally, and to always and only install them locally, per-project. Short of that, I think the next best solution is putting a `node_modules` and `.eslintrc` in your $HOME, so that any project without an eslint config and `"root": true` will inherit that config, just for you and not for any other developers on those projects.

Raynos:
I've used `eslint` with `tsdocstandard` and `npm-bin-deps`

`require.resolve` in your config ( https://github.com/Raynos/eslint-config-tsdocstandard/blob/master/eslintrc.js#L3 ) works nicely. That means the plugins need to be a dependency of the config.

[npm-bin-deps](https://www.npmjs.com/package/npm-bin-deps) allows you to not install standard-engine or the config locally but still use them together.

chdh:
I have many small TypeScript projects and I need a separate ESLint configuration (.eslintrc.js file) per project. The ESLint configuration has to be included in the Git repository of each project. The project directories are on different disk drives and not under my $HOME directory.

The installation of ESLint with the plugins and parser (TypeScript, Unicorn) uses 9492 files in 876 folders.

I suggest to add a `--resolve-parser-relative-to` command line parameter. It would be the logical complement to the `--resolve-plugins-relative-to` parameter.

Raynos:
Sounds like you want a monorepo structure on disk with a `node_modules` and then `pkg/A` , `pkg/B`, `pkg/C`, etc.

chdh:
@Raynos No, the projects are too heterogeneous (different customers, different teams, etc.). And I don't want to change the established directory structures just for the ESLint parser.

PR #13245, opened 2020-05-01
Update: treat all literals like boolean literal in no-constant-condition

Zzzen:
#### Prerequisites checklist

[ ] Documentation update
[x] Bug fix ([template](https://raw.githubusercontent.com/eslint/eslint/master/templates/bug-report.md)) #13238
[ ] Add autofixing to a rule
[ ] Add a CLI option
[ ] Add something to the core
[ ] Other, please explain:

Previously, only boolean literals are considered in `isLogicalIdentity`. All literals will be coerced into boolean now.


lib/rules/no-constant-condition.js
```diff
@@ -129,7 +129,7 @@ module.exports = {
                     const isLeftConstant = isConstant(node.left, inBooleanPosition);
                     const isRightConstant = isConstant(node.right, inBooleanPosition);
                     const isLeftShortCircuit = (isLeftConstant && isLogicalIdentity(node.left, node.operator));
-                    const isRightShortCircuit = (isRightConstant && isLogicalIdentity(node.right, node.operator));
+                    const isRightShortCircuit = (inBooleanPosition && isRightConstant && isLogicalIdentity(node.right, node.operator));
```
Zzzen, line 132: #11308 did not handle following code:
```js
if ((foo || true) === 'baz') {}
```

[demo](https://eslint.org/demo#eyJ0ZXh0IjoiLyplc2xpbnQgbm8tY29uc3RhbnQtY29uZGl0aW9uOiBcImVycm9yXCIgKi9cblxuaWYgKChmb28gfHwgdHJ1ZSkgPT09ICdiYXonKSB7fSIsIm9wdGlvbnMiOnsicGFyc2VyT3B0aW9ucyI6eyJlY21hVmVyc2lvbiI6MTEsInNvdXJjZVR5cGUiOiJtb2R1bGUiLCJlY21hRmVhdHVyZXMiOnt9fSwicnVsZXMiOnsic29ydC1pbXBvcnRzIjoyfSwiZW52Ijp7ImVzMjAyMCI6dHJ1ZX19fQ==)

mdjermanovic: This is indeed a bug that should be fixed whether or not #13238 gets accepted.

Could we maybe also this:

```js
(
    !node.parent ||
    node.parent.type !== "BinaryExpression" ||
    !(EQUALITY_OPERATORS.includes(node.parent.operator) || RELATIONAL_OPERATORS.includes(node.parent.operator))
)
```

replace with just `inBooleanPosition`?

I think it might fix similar bugs, such as:

```js
/*eslint no-constant-condition: "error"*/

if ((foo || 'bar' || 'bar') === 'bar'); // error, false positive
```

Though, it seems that the whole condition starting with  `node.operator === "||" &&` might be removed in this PR (if #13238 gets accepted), but we should also doublecheck what's happening with the recursion in `isLogicalIdentity`. It already looks buggy in the actual version:

```js
/*eslint no-constant-condition: "error"*/

if (a && false || b); // error, false positive
```

Zzzen: The cause is an inappropriate  propagation of short circuit.
fixed in https://github.com/eslint/eslint/pull/13245/commits/96d083497afb3bd8fc68b5d87b64579070565228

mdjermanovic:
@Zzzen let's wait to see if the original issue will be accepted, before reviewing.

If it doesn't get accepted, we could just fix the other bug you found in this PR.

lib/rules/no-constant-condition.js
```diff
@@ -66,15 +63,23 @@ module.exports = {
         function isLogicalIdentity(node, operator) {
             switch (node.type) {
                 case "Literal":
-                    return (operator === "||" && node.value === true) ||
-                           (operator === "&&" && node.value === false);
+                    return (operator === "||" && !!node.value) ||
+                           (operator === "&&" && !node.value);
```
mysticatea, line 67: ESTree doesn't guarantee `node.value` is a valid value. For example, if you run ESLint on Node.js that doesn't support BigInt, `node.value` is `null` at bigint literals. It will cause false positive/negative. Similarly, RegExp literals that contain new syntax, bigdecimal literals (stage 1) are, too.

Zzzen: How about reporting an unsupported error like the following code?

```js


class UnsupportedNodeError extends Error {}

function isNodeTruthy(node) {
    if (node.value === null && node.raw !== 'null') {
        throw new UnsupportedNodeError(`unsupported node: ${node.type}`)
    }

    return !!node.value;
}

function isLogicalIdentity(node, operator) {
    switch (node.type) {
        case "Literal":
            return (operator === "||" && isNodeTruthy(node)) ||
                (operator === "&&" && !isNodeTruthy(node));
        // ...
    }
}

function reportIfConstant(node) {
    try {
        if (node.test && isConstant(node.test, true)) {
            context.report({ node: node.test, messageId: "unexpected" });
        }
    } catch (error) {
        if (error instanceof UnsupportedNodeError) {
            context.report({ node: node.test, message: error.message })
        } else {
            throw error
        }
    }
}
```

mysticatea: Even if the runtime doesn't support a syntax, ESLint should check the syntax successfully if ESLint supports it. In other words, the result of linting should be the same on all supported runtimes.

Zzzen: >In other words, the result of linting should be the same on all supported runtimes.

Can we achieve runtime consistency by not supporting bigint on advanced runtime? Though bigint will not behave the same as numbers, just like numbers not behaving the same as booleans currently.

```js
function isNodeTruthy(node) {
    // TODO: bigint is not supported. wait till node v10.4.0+ is popular.
    if (typeof node.value === 'bigint' || (node.value === null && node.raw !== 'null')) {
        return false;
    }

    return !!node.value;
}
```

mysticatea: As per ESTree spec, the `Literal` nodes of BigInt have `bigint` property with the text representation of the bigint value. And the `Literal` nodes of RegExp have `regex` property with `{ pattern: string; flags: string }`. The `value` property of `Literal` nodes of those may be `null`, so we have to check the `bigint`/`regex` properties to distinguish `null` literal and the others.

Zzzen: Currently, eslint is using espree 6.2.1, which does not support bigint or regex with `u` flag. 🤦

mdjermanovic: The actual version is using espree 7.1.0

https://github.com/eslint/eslint/blob/b7d79b12eb55f220f86367d52f9f5a2f503fd4c3/package.json#L59

Zzzen: I was using an old version. Anyway, fallback handling is added. Your suggestions would be helpful.

https://github.com/eslint/eslint/pull/13245/files#diff-aee426385c16cd9eadb5fabd23dfa0bfR64-R84

Zzzen: The code below is working on node 10.0.0.

```js
if(a);
if(/foo/ui);
if(0n);
if(1n && a);
if(1n || a);
```

```
  2:4  error  Unexpected constant condition  no-constant-condition
  3:4  error  Unexpected constant condition  no-constant-condition
  5:4  error  Unexpected constant condition  no-constant-condition
```

tests/lib/rules/no-constant-condition.js
```diff
@@ -90,6 +89,19 @@ ruleTester.run("no-constant-condition", rule, {
...
         "if ((foo || {}) instanceof obj) {}",
+        "if ((foo || 'bar' || 'bar') === 'bar');",
+        {
+            code: "if ((foo || 1n) === 'baz') {}",
+            parserOptions: { ecmaVersion: 11 }
+        },
+        {
+            code: "if (a && 0n || b);",
+            parserOptions: { ecmaVersion: 11 }
+        },
```
aladdin-add, line 100: it seems weird to not report this case. it can be written to `if(b);` ?

mdjermanovic: It's okay now because the whole `a && 0n || b` isn't constant.

This would be reported by one of the enhancements proposed in https://github.com/eslint/eslint/issues/13752. If the rule checks the left operand of `||` as a condition for itself, then it would find out that it's always falsy.

mdjermanovic:
@Zzzen sorry for the delay!

I'm marking this as an accepted bug fix, based on comments and +1s from team members in #13238.

For example, it doesn't make sense to treat `false` and `void foo` as falsy values, but not do the same for `null` and `0`.

This is behavior in the actual version of this rule, and it looks inconsistent:

```js
/* eslint no-constant-condition: error */

if (false && a); // error

if (void 0 && a); // error

if (0 && a); // ok

if (null && a); // ok
```

This PR fixes that, and several [other bugs](https://github.com/eslint/eslint/pull/13245#discussion_r418571282).

mdjermanovic (changes requested):

lib/rules/no-constant-condition.js
```diff
@@ -56,6 +53,39 @@ module.exports = {
...
+                if (typeof node.regex === "object") {
+                    return true;
+                }
+
+                // https://tc39.es/ecma262/#sec-literals-numeric-literals
+                if (typeof node.bigint === "string") {
+                    return !(/^0+$/u.test(node.bigint) || /^0[box]0+$/ui.test(node.bigint));
+                }
+
+                throw new Error(`unsupported literal: ${node.raw}`);
```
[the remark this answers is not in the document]
Zzzen: Done

lib/rules/no-constant-condition.js
```diff
@@ -56,6 +53,39 @@ module.exports = {
...
+
+                // regex is always truthy
+                if (typeof node.regex === "object") {
+                    return true;
+                }
+
+                // https://tc39.es/ecma262/#sec-literals-numeric-literals
+                if (typeof node.bigint === "string") {
+                    return !(/^0+$/u.test(node.bigint) || /^0[box]0+$/ui.test(node.bigint));
+                }
```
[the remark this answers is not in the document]
Zzzen: 👌 That would simplify implementation a lot.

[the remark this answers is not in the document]
mdjermanovic: Thanks for the very quick response!

Sorry if my post was confusing, I meant to remove only the bigint part.

This was ok:

```js
// regex is always truthy
if (typeof node.regex === "object") {
    return true;
}
```

Regex syntax may have more changes in the future, so we will always use `node.regex` in rules.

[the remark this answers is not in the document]
Zzzen: 🤦‍♂️ fixed.

mdjermanovic (commented):
The code looks great, I have just a small JSDoc suggestion.

lib/rules/no-constant-condition.js
```diff
@@ -56,6 +53,34 @@ module.exports = {
         // Helpers
         //--------------------------------------------------------------------------
 
+        /**
+         * checks if a node is truthy
+         * @param {ASTNode} node any literal node
+         * @returns {boolean | null} true when node is truthy, null when it cannot be determined.
```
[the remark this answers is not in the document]
Zzzen: Sweet!

mdjermanovic (approved):
Looks great, thanks!

Also thanks for fixing several different bugs and cleaning up the code.

aladdin-add (approved):

mdjermanovic:
Great work, thanks for contributing!

PR #13252, opened 2020-05-03
Docs: update linter.verifyAndFix docs

revelt:
Currently, if code given to `linter.verifyAndFix` is of a higher generation ECMAScript than is set (and the default setting is ES5), the parser will silently fail and will not emit any nodes. This can be very hard to identify for newcomers — the vast majority of code these days _is_ beyond ES5 and it will cause parser to silently fail (and without emitting anything) with existing example:

```js
const messages = linter.verifyAndFix("var foo", {
    rules: {
        semi: 2
    }
});
```

We should instruct people to set the `parserOptions` and the default example should contain a code sample with `parserOptions` customised.

#### Prerequisites checklist

[x] Documentation update
[ ] Add autofixing to a rule
[ ] Add a CLI option
[ ] Add something to the core
[ ] Other, please explain:

Added one new paragraph and one line to the code example.

No.

### DIFF
--- a/docs/developer-guide/nodejs-api.md
+++ b/docs/developer-guide/nodejs-api.md
@@ -550,13 +550,16 @@ In this way, you can retrieve the text and AST used for the last run of `linter.
 
 ### Linter#verifyAndFix()
 
-This method is similar to verify except that it also runs autofixing logic, similar to the `--fix` flag on the command line. The result object will contain the autofixed code, along with any remaining linting messages for the code that were not autofixed.
+This method is similar to `verify` except that it also runs autofixing logic, similar to the `--fix` flag on the command line. The result object will contain the autofixed code, along with any remaining linting messages for the code that were not autofixed.
+
+The default parser `espree` is set to a default [setting](../user-guide/configuring.md#specifying-parser-options) `ecmaVersion: 5` so if your code is more modern than ES5, remember to set `parserOptions`. When the parser does not recognise the code syntax, it silently fails and does not traverse any AST or emit any nodes; the `linter.verifyAndFix()` result is still returned but it will be a non-fixed, original input.
 
 ```js
 const Linter = require("eslint").Linter;
 const linter = new Linter();
 
 const messages = linter.verifyAndFix("var foo", {
+    parserOptions: { ecmaVersion: 11 },
     rules: {
         semi: 2
     }


docs/developer-guide/nodejs-api.md
```diff
@@ -550,13 +550,16 @@ In this way, you can retrieve the text and AST used for the last run of `linter.
 
 ### Linter#verifyAndFix()
 
-This method is similar to verify except that it also runs autofixing logic, similar to the `--fix` flag on the command line. The result object will contain the autofixed code, along with any remaining linting messages for the code that were not autofixed.
+This method is similar to `verify` except that it also runs autofixing logic, similar to the `--fix` flag on the command line. The result object will contain the autofixed code, along with any remaining linting messages for the code that were not autofixed.
+
+The default parser `espree` is set to a default [setting](../user-guide/configuring.md#specifying-parser-options) `ecmaVersion: 5` so if your code is more modern than ES5, remember to set `parserOptions`. When the parser does not recognise the code syntax, it silently fails and does not traverse any AST or emit any nodes; the `linter.verifyAndFix()` result is still returned but it will be a non-fixed, original input.
```
anikethsaha, line 555: I dont think it silently fails ! can you confirm !?

also `setting` -> `options`

docs/developer-guide/nodejs-api.md
```diff
@@ -550,13 +550,16 @@ In this way, you can retrieve the text and AST used for the last run of `linter.
...
 
 ```js
 const Linter = require("eslint").Linter;
 const linter = new Linter();
 
 const messages = linter.verifyAndFix("var foo", {
+    parserOptions: { ecmaVersion: 11 },
```
anikethsaha, line 562: for `var` , parserOptions.ecmaVersion` is not required. you can add `let` instead ;

docs/developer-guide/nodejs-api.md
revelt: @anikethsaha thank you for swift review. Speaking strictly in terms of `linter.verifyAndFix`, if parser ECMA version is set wrongly and code is not recognised, `verifyAndFix` will not `throw` ("loud" error) — instead it will return unfixed result — that's what I mean by "failing silently". I spotted this troubleshooting an eslint plugin — ECMA version was not set right (but I didn't know at that time) — ES6 code caused `verifyAndFix` to return unfixed result — it was not throwing (or at least propagating inner parser throwing) to the terminal!

revelt: I'm happy to remove the paragraph, just I thought, in proper API documentation, this part of `verifyAndFix` behaviour is not properly documented.

docs/developer-guide/nodejs-api.md
revelt: good point, let's make the code snippet to warrant the use of `ecmaVersion `

docs/developer-guide/nodejs-api.md
anikethsaha: > I'm happy to remove the paragraph, just I thought, in proper API documentation, this part of `verifyAndFix` behaviour is not properly documented.

I don't think you have to remove the whole para cause it does includes some good points. About the silent thing, I will test and I will get back here.

kaicataldo:
I'm not sure I understand the problem. Both `Linter#verify()` and `Linter#verifyAndFix()` report linting errors in the results, not by throwing an error. A parsing error should be treated the same as a rule linting error, and given that this is intended to be used by a consumer, I think this API makes sense. We don't want to be crashing the consumer's application.

kaicataldo:
To further clarify, I'm not understanding why we should specifically be calling out the `ecmaVersion` configuration option or why `Linter#verifyAndFix()` is being singled out in this change. It seems to be working as intended to me. Do you mind clarifying what the issue is?

PR #13254, opened 2020-05-03
Chore: Replace the inquirer dependency with enquirer

### LINKED ISSUE #13208: Package size seems big
6.8.0

The package install size of `eslint` weighs 16.3MB:
https://packagephobia.now.sh/result?p=eslint

This seems quite a lot and decreasing this would benefit initial install times among other things.
Might be some (better) other opportunities to decrease this size.

Seems mainly because of `rxjs` adding 17MB (locally for me `eslint` weighs 39MB) which is used by `inquirer`. Replacing `inquirer` with `enquirer` would shave off about half the the package size.

`inquirer`: [install size](https://packagephobia.now.sh/result?p=inquirer)
`enquirer`: [install size](https://packagephobia.now.sh/result?p=enquirer)

Added benefit would be improving [the start up time](https://github.com/enquirer/enquirer#-performance) (module load time):
```
enquirer: 4.013ms
inquirer: 286.717ms
```

Yes.

Siilwyn:
Closes #13208

#### Prerequisites checklist

[ ] Documentation update
[ ] Add autofixing to a rule
[ ] Add a CLI option
[ ] Add something to the core
[x] Other, please explain:
See https://github.com/eslint/eslint/issues/13208: replace dependency to decrease package size.

Replace the dependency and changed the equivalent methods to work.

Mostly if edge cases work, I've tested some paths but not certain if everything went over like how it was supposed to work in the first place. (never used eslint --init myself before)


lib/init/config-initializer.js
```diff
@@ -428,14 +428,17 @@ function askInstallModules(modules, packageJsonExists) {
...
             message: "Would you like to install them now with npm?",
-            default: true,
-            when() {
-                return modules.length && packageJsonExists;
+            initial: () => true,
+            skip() {
+                return !(modules.length && packageJsonExists);
+            },
+            result(input) {
+                return this.skip ? null : input;
```
Siilwyn, line 441: This result method is needed since enquirer will even return the initial value when the question is skipped, unlike inquirer.

lib/init/config-initializer.js
```diff
@@ -445,110 +448,115 @@ function askInstallModules(modules, packageJsonExists) {
...
 
-    return inquirer.prompt([
+    return enquirer.prompt([
         {
-            type: "list",
+            type: "select",
             name: "purpose",
             message: "How would you like to use ESLint?",
-            default: "problems",
+            initial: 1,
```
Siilwyn, line 463: The returned number matches the `name` value of _nth_ in the `choices` array.

lib/init/config-initializer.js
```diff
@@ -445,110 +448,115 @@ function askInstallModules(modules, packageJsonExists) {
...
-                { name: "Inspect your JavaScript file(s)", value: "auto" }
+                { message: "Use a popular style guide", name: "guide" },
+                { message: "Answer questions about your style", name: "prompt" },
+                { message: "Inspect your JavaScript file(s)", name: "auto" }
             ],
-            when(answers) {
-                return answers.purpose === "style";
+            skip() {
+                return this.state.answers.purpose !== "style";
+            },
```
Siilwyn, line 519: The `skip` method is the reverse of `when` so the conditional statement is reversed.

yeonjuan:
Hi. @Siilwyn Thanks for working this. 😀
I tried it and found two issues.
스크린샷 2020-05-05 오전 2 32 41

1. default value representation: question - `(y/n)` / answer - `(true/false)`

2. **Does your project use TypeScript? (Y/n)**  - this question shows us the default value as `Y` but the default value of the answer is `false`

Siilwyn:
@yeonjuan alright, I've replaced all confirm prompts to have more humane answers.

yeonjuan:
@Siilwyn 
> (I see the commit-message CI fail, but I think this is alright since when this PR is ready to merge it can be squashed to use my first commit which is compliant right?)

Yes but, the PR message should follow the guideline.
Because we use the PR title as a commit message when it squashed(If there are multiple commits)

So, Would you mind change the PR title?

lib/init/config-initializer.js
```diff
@@ -445,114 +450,121 @@ function askInstallModules(modules, packageJsonExists) {
...
+            result(input) {
+                return this.skip ? null : input;
```
yeonjuan, line 525: @Siilwyn 
It seems not working as we expected.
When I answered with the "guide," in this step.  but the [selecting styleguide step](https://github.com/eslint/eslint/pull/13254/files#diff-bc3fbe576813fc2a16ef2036becd32f8R533) was skipped.

It seems the `this.skip` is a function (`truthy` value), so the result always be a `null`


스크린샷 2020-05-05 오후 10 19 31

lib/init/config-initializer.js
```diff
@@ -445,114 +450,121 @@ function askInstallModules(modules, packageJsonExists) {
...
             name: "typescript",
             message: "Does your project use TypeScript?",
-            default: false
+            enabled: "Yes",
+            disabled: "No",
+            initial: 0
         },
         {
-            type: "checkbox",
+            type: "multiselect",
```
yeonjuan, line 503: Does the `enquirer` support a help message for `multiselect`?
IMO, it would be nice if we give help message for it

* inquirer (master branch version)
스크린샷 2020-05-05 오후 10 48 50

* enquirer (pr version)
스크린샷 2020-05-05 오후 10 47 56

Siilwyn: Not really, but I've added it to the message to clear the usage up.

lib/init/config-initializer.js
```diff
@@ -445,114 +450,121 @@ function askInstallModules(modules, packageJsonExists) {
...
             name: "env",
-            message: "Where does your code run?",
-            default: ["browser"],
+            message: "Where does your code run? (Press <space> to select, <a> to toggle all, <i> to invert selection)",
```
yeonjuan, line 505: @Siilwyn 
Oh, I found that `enquirer` support `hint` option. It's non-blocking suggestion :) 

```suggestion
            message: "Where does your code run?",
            hint: "(Press <space> to select, <a> to toggle all, <i> to invert selection)",
```
* message
스크린샷 2020-05-06 오전 1 40 04

* hint
스크린샷 2020-05-06 오전 1 40 51

Siilwyn: Nice! I missed that bit.

Siilwyn:
What happens now?

lib/init/config-initializer.js
nzakas: Can you add that as a comment?

nzakas:
@Siilwyn we still need some other reviews before this can be merged. Stay tuned.

lib/init/config-initializer.js
nzakas: @siilwyn can you add this as a comment in the file?

Siilwyn: Yes :+1: 
(Didn't get a notification for your first comment.)

nzakas:
Oops, looks like that last commit introduced a conflict. Can you take a look?

Siilwyn:
@nzakas should be fixed now!

kaicataldo:
Thanks for contributing!

PR #13268, opened 2020-05-07
Update: no-void add an option to allow at start of concise arrow functions

### LINKED ISSUE #13020: [no-void]: allow void arrow functions
`no-void`

Fewer

Option - could also be implemented as standard behaviour depending on interpretation of the rules description.

```js
const log = x => void console.log(x);
```

Errors

Not error

Yes

-----

There are some [legitimate uses](https://gist.github.com/slikts/dee3702357765dda3d484d8888d3029e) for the void operator, one of which is even mentioned in the opening sentence of the `no-void` docs:

>  It can be used to ignore any side effects expression may produce

The description of the `no-void` rule to me seems to indicate that the rule is actually ideally about preventing the use of `void` to get a "pure" undefined back in the pre-ES5 days when undefined was mutable: 

> The common case of using void operator is to get a "pure" undefined value as prior to ES5 the undefined variable was mutable:

So it might be arguable that this is technically a bug? But I don't have a major preference on if it's an option or not :)

Based off #12688, which was closed for not following the template.

### LINKED ISSUE #13299: [no-void]: allow void arrow functions
`no-void`

Fewer

Option - could also be implemented as standard behaviour depending on interpretation of the rules description.

```js
const log = x => void console.log(x);
```

Errors

Not error

Yes, I have a PR open: #13268

-----

There are some [legitimate uses](https://gist.github.com/slikts/dee3702357765dda3d484d8888d3029e) for the void operator, one of which is even mentioned in the opening sentence of the `no-void` docs:

>  It can be used to ignore any side effects expression may produce

The description of the `no-void` rule to me seems to indicate that the rule is actually ideally about preventing the use of `void` to get a "pure" undefined back in the pre-ES5 days when undefined was mutable in global scope: 

> The common case of using void operator is to get a "pure" undefined value as prior to ES5 the undefined variable was mutable:

So it might be arguable that this is technically a bug? But I don't have a major preference on if it's an option or not :)

Based off #12688, which was closed for not following the template.
Based off #13020, which didn't get much support, but I still think this is a useful minor addition.

G-Rath:
fixes #13020
fixes #13299

#### Prerequisites checklist

[ ] Documentation update
[ ] Add autofixing to a rule
[ ] Add a CLI option
[ ] Add something to the core
[ ] Other, please explain:

I changed `no-void` to support an option that allows using `void `at the start of single line arrow functions.

This allows for the use of `void` to ignore any side effects an expression may produce, as described in the docs for the `no-void` rule.

Nothing in particular.

### DIFF
--- a/docs/rules/no-void.md
+++ b/docs/rules/no-void.md
@@ -92,6 +92,36 @@ void foo;
 void someFunction();
 ```
 
+### allowAtStartOfConciseArrowFunctions
+
+When `allowAtStartOfConciseArrowFunctions` is set to true, the rule will not error if the void operator is used at the start of concise arrow functions.
+This can be used as a way of explicitly marking concise arrow functions as not returning a value without having to add braces, which formatters often expand
+with newlines.
+
+Examples of **incorrect** code for `{ "allowAtStartOfConciseArrowFunctions": true }`:
+
+```js
+const log = function(message) { return void console.log(message); };
+
+const warn = message => {
+  return void console.warn(message);
+};
+
+const error = message => void (message ? console.error(message) : void 0);
+```
+
+Examples of **correct** code for `{ "allowAtStartOfConciseArrowFunctions": true }`:
+
+```js
+const log = message => void console.log(message);
+
+const warn = message => {
+  console.warn(message);
+}
+
+const error = message => void (message ? console.error(message) : 0);
+```
+
 ## When Not To Use It
 
 If you intentionally use the `void` operator then you can disable this rule.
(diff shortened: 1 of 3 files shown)


btmills:
@G-Rath thank you for contributing! Since we're trying not to add any more options to the built-in stylistic rules, I'll close this, but I'd encourage you to release your implementation with this option as a plugin and share it here and on the original issue. If you're interested, I left more details on the discussion in today's TSC meeting in [my comment on the issue](https://github.com/eslint/eslint/issues/13299#issuecomment-659694552).

PR #13282, opened 2020-05-11
Update: changed curly reporting location (refs #12334)

anikethsaha:
#### Prerequisites checklist

[ ] Documentation update
[ ] Add autofixing to a rule
[ ] Add a CLI option
[ ] Add something to the core
[ ] Other, please explain:

Updated the location of reporting for curly braces.

Currently, the `column` while reporting was always `1` except for `else`. 

I made the following changes. 

- for missing parens, it will show the location where is should be placed. 
for example 

```
1 for(;;)
2 foo();
  ^ <- this location will be reported.
```

or 
```js
if(foo) bar;
    // ^ <- this location will be reported
```

- for unexpected braces, 
the first unexpected `{` will be reported.

```js
if(foo) {foo()};
     // ^ 
```

The `else` logic for both is not changed as I think it was working fine earlier

None


anikethsaha:
I am not sure what am I missing in CI  cause the failing test is un-related to my changes and tests are passing in my machine. 

image

any suggestions would be appreciated 👍

mdjermanovic:
Those two failing tests should be fixed by #13277.

kaicataldo:
@anikethsaha Do you mind rebasing now that #13277 has been merged?

lib/rules/curly.js
```diff
@@ -309,6 +309,98 @@ module.exports = {
                 hasUnsafeIf(statement) && isFollowedByElseKeyword(node);
         }
 
+        /**
+         * Examples:
+         *
+         * The following location should be reported
```
kaicataldo, line 315: This documentation is helpful :+1:

lib/rules/curly.js
```diff
@@ -309,6 +309,98 @@ module.exports = {
...
+         * while (bar) { foo(); }
+         *             ^
+         *
+         * do{foo();} while(bar);
+         *   ^
+         *
+         * for (var foo of bar) {console.log(foo)}
+         *                      ^
+         *
+         * Get the location of unexpected parens. mostly the starting parens `{`
```
kaicataldo, line 378: Can we change `parens` to `curly braces` in both the JSDoc blocks and the function name?

lib/rules/curly.js
```diff
@@ -309,6 +309,98 @@ module.exports = {
...
+         *
+         * for (var foo in bar) console.log(foo)
+         *                      ^
+         *
+         * Get the location of missing parent location where it should have been placed.
+         * @param {ASTNode} node the node to check.
+         * @param {string} name name to check.
+         * @returns {Object} Position object.
+         */
+        function getMissingLoc(node, name) {
```
kaicataldo, line 337: There's a lot of duplicated logic between these two methods. Could we combine them into one and pass a parameter indicating whether the curly braces are missing or not?

```js
function getReportLoc(node: ASTNode, name: string, isMissing: boolean) {}

// or, to avoid the boolean trap, we could also use an options object:

function getReportLoc(
    { node, name, isMissing }: { node: ASTNode, name: string, isMissing: boolean }
) {}
```

lib/rules/curly.js
```diff
@@ -309,6 +309,98 @@ module.exports = {
...
+         *
+         * Get the location of unexpected parens. mostly the starting parens `{`
+         * @param {ASTNode} node the node to check.
+         * @param {string} name name to check.
+         * @returns {Object} Position object.
+         */
+        function getUnexpectedParenLoc(node, name) {
+            switch (name) {
+                case "else":
+                    return getElseKeyword(node).loc.start;
```
kaicataldo, line 386: @mdjermanovic Do we want to include `loc.end` as well for these cases?

lib/rules/curly.js
anikethsaha: > function getReportLoc(
    { node, name, isMissing }: { node: ASTNode, name: string, isMissing: boolean }
) {}

I didnt get this, so I went with the first one if its ok !.

let me know if this needs to be changed.

lib/rules/curly.js
mdjermanovic: I'd say - yes. If we are targeting the start location of a particular token, then it probably makes sense to pass the end location of that token as well.

anikethsaha: for unexpected curly braces, I think the start location can be fine as its a single token so start and end location will be the same. 

for missing curly brace,  i will make the changes. 

let me know if this is fine ?

anikethsaha:
Should all of them have the starting location of the body ?

lib/rules/curly.js
mdjermanovic: > for unexpected curly braces, I think the start location can be fine as its a single token so start and end location will be the same.

For a single-character token, the end location isn't the same as its start location. The end location would be that with `column + 1`

anikethsaha: ok then I think it will show like this 

```js
if(foo) {bar}
 //     ^^
```

right ?

mdjermanovic: > ok then I think it will show like this
> 
> ```js
> if(foo) {bar}
>  //     ^^
> ```

It would underline just `{`, because the character at the end loc isn't considered as being included.

It's like a selection between cursor that is positioned at `start` and cursor that is positioned at `end`.

mdjermanovic:
> Should all of them have the starting location of the body ?

That would be reasonable given that it marks the place where the auto-fix insert `{`.

On the other hand, editors usually don't show zero-width ranges well. Also, user might want to insert `{` at a different place, and the message says `"Expected { after..."`.

Maybe we could instead consider reporting the location of the first token (which would be `if`, `else`, `while`, `do`, or `for`):

```js
if (foo) bar();
^^
```

or the last token before the body (which would be `)`, `else` or `do`):

```js
if (foo) bar();
       ^      
```

or maybe even the whole range before the body (though, this could underline a lot of code):

```js
if (foo) bar();
^^^^^^^^      
```

anikethsaha:
> Maybe we could instead consider reporting the location of the first token (which would be `if`, `else`, `while`, `do`, or `for`):
> 
> ```js
> if (foo) bar();
> ^^
> ```

I think this is the current behavior, it shows the column as `1` 


> or the last token before the body (which would be `)`, `else` or `do`):
> 
> ```js
> if (foo) bar();
>        ^      
> ```

let me try this,

> or maybe even the whole range before the body (though, this could underline a lot of code):
> 
> ```js
> if (foo) bar();
> ^^^^^^^^      
> ```

true, it can be longer if the condition inside the is long.

mdjermanovic:
> > Maybe we could instead consider reporting the location of the first token (which would be `if`, `else`, `while`, `do`, or `for`):
> > ```js
> > if (foo) bar();
> > ^^
> > ```
> 
> I think this is the current behavior, it shows the column as `1`

It is the current behavior for `start` location, but it doesn't report `end` location.

If we choose this option, it might make sense to report the whole range of the first token (i.e., both its `start` and `end` locations).

anikethsaha:
**Update :**

Now, the reporting location is as follows.

for `missing` 

```js
if (foo) bar();
    // ^^

if (foo) 
    // ^^   
 bar();
    

// `else`

if (foo) bar(); else bar(1);
    //          ^^^^


// `for`

for (;;) bar();
//     ^^

``` 

for `unexpected`

```js

if (foo) {bar();}
   //    ^^

// `else`
// should this be same as rest for missing ? like pointing at the curly brace ?
if (foo) bar(); else {bar(1);}
    //          ^^^^

for (;;) {bar()};
//       ^^


```

@mdjermanovic let me know if this is ok !

lib/rules/curly.js
kaicataldo: No worries! It looks a bit confusing, but this is indicating type definitions for a destructured object.

anikethsaha: opz, I got it, I was thinking as a point of `js` .

should I change this?

mdjermanovic:
> @mdjermanovic let me know if this is ok !

I'm not sure which of the [three solutions](https://github.com/eslint/eslint/pull/13282#issuecomment-628473591) I suggested would be the most appropriate, or if there's another one I didn't think of.

@kaicataldo thoughts?

kaicataldo:
Could it make sense to always report the locations between the last token before the body and the first token of the body, regardless of whether it's missing or not?

Missing:

```js
if (foo) bar();
       ^^^

if (foo) 
       ^^   
 bar();
^^ 

if (foo) bar(); else bar(1);
       ^^^         ^^^

for (;;) bar();
       ^^^
```

Unexpected:

```js
if (foo) {bar();}
       ^^^^

if (foo) bar(); else {bar(1);}
                   ^^^^

for (;;) {bar()};
       ^^^^

```

anikethsaha:
I  think that would make sense for missing case. 
For Unexpected case, I think it should report the  curly brace itself. 

WDYT ?

kaicataldo:
That makes sense to me!

anikethsaha:
I have made the changes as discussed above. 

Let me know if anything else if required

lib/rules/curly.js
```diff
@@ -309,6 +331,115 @@ module.exports = {
...
+            if (isMissing) {
+
+                if (name === "for" || name === "do") {
+                    const token = sourceCode.getTokenBefore(node.body);
+
+                    return {
+                        start: token.loc.end,
+                        end: node.body.loc.start
+                    };
+                }
```
mdjermanovic, line 409: This could return a zero-length range if the previous token and the body are adjacent:

```js
/* eslint curly: ["error"]*/

for(;;)doSomething();
```

I'm not sure was the idea to underline the previous token as well, or just the range after the previous token?

lib/rules/curly.js
```diff
@@ -309,6 +331,115 @@ module.exports = {
...
+
+                if (name === "for-of" || name === "for-in") {
+                    return {
+                        start: node.right.loc.end,
+                        end: node.body.loc.start
+                    };
+                }
```
mdjermanovic, line 416: Why should be `for-of`, `for-in`, `if` and `while` different than `for` and `do`:

image

lib/rules/curly.js
anikethsaha: It will show the next token

image

mdjermanovic: The goal of #12334 is to avoid zero-length ranges, so it would be nice to find a solution that entirely fixes this rule, if possible.

anikethsaha: how about from the last `;` to `)` ?

lib/rules/curly.js
anikethsaha: I will check this.
It might have come along with https://github.com/eslint/eslint/pull/13282/commits/b3330be4930a8c4d2f9d65b599d73d6e94cf9ba2

lib/rules/curly.js
mdjermanovic: If the idea is to include the first token before the body in underlining, we could take its `start` location.

anikethsaha: Updated location

image

lib/rules/curly.js
anikethsaha: done

lib/rules/curly.js
```diff
@@ -309,6 +331,103 @@ module.exports = {
...
+
+                return createNextLocation(node.body.loc.start);
+            }
+
+            if (name === "if") {
+                return isMissing
+                    ? {
+                        start: node.test.loc.end,
```
mdjermanovic, line 415: This could also underline whitespace between the `test` node and the opening paren:

image

(the same applies to `while`)

Could we maybe unify all cases and thus simplify the code? I think the only difference is whether the "body" is `.body`, `.consequent`, or `.alternate` child, so maybe we could find the "body" node first (based on `name`) and then just take the first token before that node.

lib/rules/curly.js
```diff
@@ -309,6 +331,103 @@ module.exports = {
...
+                        end: node.consequent.loc.start
+                    }
+                    : createNextLocation(node.consequent.loc.start);
```
mdjermanovic, line 418: Can we get the first token and return its `loc` instead of calculating 1-width range?

snitin315:
I will finish it.

snitin315:
Stale. Closing in favor of #14766

PR #13285, opened 2020-05-12
Update: reporting location for semi-spacing (refs #12334)

anikethsaha:
#### Prerequisites checklist

[ ] Documentation update
[ ] Add autofixing to a rule
[ ] Add a CLI option
[ ] Add something to the core
[ ] Other, please explain:

Updated the location when is there unexpected space (both for before and after)

Currently 
```js
var a = 'b'   ;
          //  ^ it will report here instead of the spaces
```

with this change
```js
var a = 'b'   ;
        // ^^^  it will report here
```
For missing space it currently reports 

```js
var a = 'a';var b = 'b';
       //  ^ it reports here i.e over the `;`
```

I think it should report the next token after `;` right ? i.e `v` in this case. 

I didn't make this change for the missing space, let me know if this is fine to change !

[repl](https://eslint.org/demo#eyJ0ZXh0IjoiLyplc2xpbnQgc2VtaS1zcGFjaW5nOiBcImVycm9yXCIqL1xuXG52YXIgZm9vICA7XG52YXIgZm9vO3ZhciBiYXI7Iiwib3B0aW9ucyI6eyJwYXJzZXJPcHRpb25zIjp7ImVjbWFWZXJzaW9uIjo1LCJzb3VyY2VUeXBlIjoic2NyaXB0IiwiZWNtYUZlYXR1cmVzIjp7fX0sInJ1bGVzIjp7fSwiZW52Ijp7fX19)

### DIFF
--- a/lib/rules/semi-spacing.js
+++ b/lib/rules/semi-spacing.js
@@ -136,11 +136,9 @@ module.exports = {
          */
         function checkSemicolonSpacing(token, node) {
             if (astUtils.isSemicolonToken(token)) {
-
                 if (hasLeadingSpace(token)) {
-                    const tokenBefore = sourceCode.getTokenBefore(token);
-
                     if (!requireSpaceBefore) {
+                        const tokenBefore = sourceCode.getTokenBefore(token);
                         const loc = {
                             start: tokenBefore.loc.end,
                             end: token.loc.start
--- a/lib/rules/semi-spacing.js
+++ b/lib/rules/semi-spacing.js
@@ -117,6 +117,18 @@ module.exports = {
         }
 
         /**
+         * Report location example :
+         *
+         * for unexpected space `before`
+         *
+         * var a = 'b'   ;
+         *            ^^^
+         *
+         * for unexpected space `after`
+         *
+         * var a = 'b';  c = 10;
+         *             ^^
+         *
          * Reports if the given token has invalid spacing.
          * @param {Token} token The semicolon token to check.
          * @param {ASTNode} node The corresponding node of the token.
@@ -124,16 +136,24 @@ module.exports = {
          */
         function checkSemicolonSpacing(token, node) {
             if (astUtils.isSemicolonToken(token)) {
-                const location = token.loc.start;
 
+                // console.log(token.loc)
                 if (hasLeadingSpace(token)) {
+                    const tokenBefore = sourceCode.getTokenBefore(token);
+                    const loc = {
+                        start: tokenBefore.loc.end,
+                        end: {
+                            line: token.loc.start.line,
+                            column: token.loc.start.column - 1
+                        }
+                    };
+
                     if (!requireSpaceBefore) {
                         context.report({
                             node,
-                            loc: location,
+                            loc,
                             messageId: "unexpectedWhitespaceBefore",
                             fix(fixer) {
-                                const tokenBefore = sourceCode.getTokenBefore(token);
 
                                 return fixer.removeRange([tokenBefore.range[1], token.range[0]]);
                             }
@@ -141,9 +161,11 @@ module.exports = {
                     }
                 } else {
                     if (requireSpaceBefore) {
+                        const loc = token.loc.start;
+
                         context.report({
                             node,
-                            loc: location,
+                            loc,
                             messageId: "missingWhitespaceBefore",
                             fix(fixer) {
                                 return fixer.insertTextBefore(token, " ");
@@ -155,12 +177,20 @@ module.exports = {
                 if (!isFirstTokenInCurrentLine(token) && !isLastTokenInCurrentLine(token) && !isBeforeClosingParen(token)) {
                     if (hasTrailingSpace(token)) {
                         if (!requireSpaceAfter) {
+                            const tokenAfter = sourceCode.getTokenAfter(token);
+                            const loc = {
+                                start: token.loc.end,
+                                end: {
+                                    line: tokenAfter.loc.start.line,
+                                    column: tokenAfter.loc.start.column - 1
+                                }
+                            };
+
                             context.report({
                                 node,
-                                loc: location,
+                                loc,
                                 messageId: "unexpectedWhitespaceAfter",
                                 fix(fixer) {
-                                    const tokenAfter = sourceCode.getTokenAfter(token);
 
                                     return fixer.removeRange([token.range[1], tokenAfter.range[0]]);
                                 }
@@ -168,9 +198,11 @@ module.exports = {
                         }
                     } else {
                         if (requireSpaceAfter) {
+                            const loc = token.loc.start;
+
                             context.report({
                                 node,
-                                loc: location,
+                                loc,
                                 messageId: "missingWhitespaceAfter",
                                 fix(fixer) {
                                     return fixer.insertTextAfter(token, " ");
--- a/lib/rules/semi-spacing.js
+++ b/lib/rules/semi-spacing.js
@@ -139,12 +139,13 @@ module.exports = {
 
                 if (hasLeadingSpace(token)) {
                     const tokenBefore = sourceCode.getTokenBefore(token);
-                    const loc = {
-                        start: tokenBefore.loc.end,
-                        end: token.loc.start
-                    };
 
                     if (!requireSpaceBefore) {
+                        const loc = {
+                            start: tokenBefore.loc.end,
+                            end: token.loc.start
+                        };
+
                         context.report({
                             node,
                             loc,
--- a/lib/rules/semi-spacing.js
+++ b/lib/rules/semi-spacing.js
@@ -141,10 +141,7 @@ module.exports = {
                     const tokenBefore = sourceCode.getTokenBefore(token);
                     const loc = {
                         start: tokenBefore.loc.end,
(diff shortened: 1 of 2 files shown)


lib/rules/semi-spacing.js
```diff
@@ -117,33 +117,55 @@ module.exports = {
...
 
+                // console.log(token.loc)
                 if (hasLeadingSpace(token)) {
+                    const tokenBefore = sourceCode.getTokenBefore(token);
+                    const loc = {
+                        start: tokenBefore.loc.end,
+                        end: {
+                            line: token.loc.start.line,
+                            column: token.loc.start.column - 1
+                        }
```
mdjermanovic, line 148: The end loc is exclusive (it isn't supposed to be underlined), so this would produce:

image


It can be just: `end: token.loc.start`.

(similar applies to the `after: false` branch)

lib/rules/semi-spacing.js
```diff
@@ -117,33 +117,55 @@ module.exports = {
...
          * Reports if the given token has invalid spacing.
          * @param {Token} token The semicolon token to check.
          * @param {ASTNode} node The corresponding node of the token.
          * @returns {void}
          */
         function checkSemicolonSpacing(token, node) {
             if (astUtils.isSemicolonToken(token)) {
-                const location = token.loc.start;
 
+                // console.log(token.loc)
```
mdjermanovic, line 140: This is probably left from debugging :)

lib/rules/semi-spacing.js
```diff
@@ -117,33 +117,55 @@ module.exports = {
...
                             fix(fixer) {
-                                const tokenBefore = sourceCode.getTokenBefore(token);
 
                                 return fixer.removeRange([tokenBefore.range[1], token.range[0]]);
                             }
                         });
                     }
                 } else {
                     if (requireSpaceBefore) {
+                        const loc = token.loc.start;
```
mdjermanovic, line 164: Should this be just `token.loc`?

(same for the `after: true` branch)

lib/rules/semi-spacing.js
anikethsaha: ahh, I see. it does make things better when being tested in an editor. 

_Just a small question on a different context._

I tried using the `master` branch with `vscode-eslint` using the `file` type dep for `eslint` i.e

`"eslint": file:../<path-to-my-local-eslint>` in this line 

https://github.com/microsoft/vscode-eslint/blob/master/package.json#L451

and opening the vscode's extension debugger doesn't seems to work out for me. did you have any different approach on this?

anikethsaha: > The end loc is exclusive (it isn't supposed to be underlined),

so when we give both start and end value it does puts a line over that right ?
and if the starting and ending column is same i.e

with this

```js
  end: {
                            line: token.loc.start.line,
                            column: token.loc.start.column - 1
                        }
```
and there is only single space

`var a = 'a' ;`
so it will just underline that space only right ? or it won't underline that?

cause if I make it to 

```js
  end:  token.loc.start
`

for `var a = 'a' ;` it will put a line below the space and for `;` as well right . 

_I am not able to work this our with `vscode-eslint` that's why just need to confirm these_

File: lib/rules/semi-spacing.js (hunk 3 above)
Comment (anikethsaha) [reply lib/rules/semi-spacing.js]: done 👍

File: lib/rules/semi-spacing.js (hunk 1 above)
Comment (mdjermanovic) [reply lib/rules/semi-spacing.js:147]: > I tried using the `master` branch with `vscode-eslint` using the `file` type dep for `eslint` i.e
> 
> `"eslint": file:../<path-to-my-local-eslint>` in this line
> 
> https://github.com/microsoft/vscode-eslint/blob/master/package.json#L451
> 
> and opening the vscode's extension debugger doesn't seems to work out for me. did you have any different approach on this?

That line is in `devDependencies`, so I think it's for linting `vscode-eslint` project itself?

Not sure about debugging, but in order to just see how the change looks in the editor you can open a test js project and add `eslint` dependency pointing to local working copy (e.g. `"eslint": "file:../eslint"`, then do `npm install`, also vscode restart is required after each change). Or, install `eslint` normally and overwrite the rule in `node_modules/eslint/lib/rules` (again, vscode restart is required after each change).

Comment (anikethsaha) [reply lib/rules/semi-spacing.js:147]: > That line is in `devDependencies`, so I think it's for linting `vscode-eslint` project itself?

yeah, cause I was testing them on `vscode-eslint` so I thought if I can use the dev dep's eslint to link it with `master` and use it in `vscode-eslint`'s code itself. 



> Not sure about debugging, but in order to just see how the change looks in the editor you can open a test js project and add `eslint` dependency pointing to local working copy (e.g. `"eslint": "file:../eslint"`, then do `npm install`, also vscode restart is required after each change). Or, install `eslint` normally and overwrite the rule in `node_modules/eslint/lib/rules` (again, vscode restart is required after each change).

this seems much better,  thanks I will check this. 👍

Comment (mdjermanovic) [reply lib/rules/semi-spacing.js:147]: > cause if I make it to
> 
> ```js
>   end:  token.loc.start
> ```
> 
> for `var a = 'a' ;` it will put a line below the space and for `;` as well right .

It will underline only the space, `end` doesn't count as being a part of the range.

Comment (anikethsaha) [reply lib/rules/semi-spacing.js:147]: ohh, 
so how can I underline the whole space cause it does need the range right

```js
var a = 'a'      ;
           ^^^^^^
```

mdjermanovic: I think something like this would do the work:

```js
{
    start: tokenBefore.loc.end,
    end: token.loc.start
}
```

(the last character doesn't count)

anikethsaha: done 👍

lib/rules/semi-spacing.js
```diff
@@ -155,22 +173,32 @@ module.exports = {
                 if (!isFirstTokenInCurrentLine(token) && !isLastTokenInCurrentLine(token) && !isBeforeClosingParen(token)) {
                     if (hasTrailingSpace(token)) {
                         if (!requireSpaceAfter) {
+                            const tokenAfter = sourceCode.getTokenAfter(token);
+                            const loc = {
+                                start: token.loc.end,
+                                end: {
+                                    line: tokenAfter.loc.start.line,
+                                    column: tokenAfter.loc.start.column - 1
+                                }
```
mdjermanovic, line 182: This should be just `end: tokenAfter.loc.start`.

tests/lib/rules/semi-spacing.js
```diff
@@ -136,7 +159,31 @@ ruleTester.run("semi-spacing", rule, {
...
+            errors: [
+                {
+                    messageId: "unexpectedWhitespaceAfter",
+                    type: "VariableDeclaration",
+                    line: 1,
+                    column: 13,
+                    endColumn: 15,
+                    endLine: 1
+                }
+            ]
```
mdjermanovic, line 186: Can we flip `endColumn` and `endLine` lines, to have the same order as with `line`, `column`

tests/lib/rules/semi-spacing.js
```diff
@@ -206,7 +253,74 @@ ruleTester.run("semi-spacing", rule, {
...
+            options: [{ before: false, after: false }],
+            parserOptions: { ecmaVersion: 6 },
+            errors: [{
+                line: 1,
+                column: 6,
+                type: "ForStatement",
+                messageId: "unexpectedWhitespaceBefore",
+                endLine: 1,
+                endColumn: 7
+            },
```
mdjermanovic, line 301: Can we group `line`, `column`, `endLine`, `endColumn` lines, for consistency.

(the same applies to some other tests)

lib/rules/semi-spacing.js
```diff
@@ -117,33 +117,51 @@ module.exports = {
...
                             fix(fixer) {
-                                const tokenBefore = sourceCode.getTokenBefore(token);
 
                                 return fixer.removeRange([tokenBefore.range[1], token.range[0]]);
                             }
                         });
                     }
                 } else {
                     if (requireSpaceBefore) {
+                        const loc = token.loc;
```
mdjermanovic, line 160: It would be nice to add `endLine` and `endColumn` to one test to confirm this change as well

(the same applies to the other `const loc = token.loc;` line).

lib/rules/semi-spacing.js
```diff
@@ -117,33 +117,51 @@ module.exports = {
...
             if (astUtils.isSemicolonToken(token)) {
-                const location = token.loc.start;
 
                 if (hasLeadingSpace(token)) {
+                    const tokenBefore = sourceCode.getTokenBefore(token);
+                    const loc = {
+                        start: tokenBefore.loc.end,
+                        end: token.loc.start
+                    };
+
```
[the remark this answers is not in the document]
mdjermanovic: Sorry, it maybe wasn't clear, `const tokenBefore = sourceCode.getTokenBefore(token);` should also go to the `if` below (to mirror the `after` branch).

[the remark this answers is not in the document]
anikethsaha: ohh ok 😄

kaicataldo:
LGTM, thank you!

PR #13290, opened 2020-05-13
Fix mergeWithoutOverwrite with hasOwnProperty

### LINKED ISSUE #13289: Settings object can't have keys that conflict with Object.prototype

* **ESLint Version:** 7.0.0, 6.8.0, probably all others
* **Node Version:** any
* **npm Version:** any

<details>

```json
{
  "settings": {
    "myPlugin": {
      "constructor": "val",
      "hasOwnProperty": "val",
      "isPrototypeOf": "val",
      "propertyIsEnumerable": "val",
      "toLocaleString": "val",
      "toString": "val",
      "valueOf": "val"
    }
  }
}

```

</details>

These settings are available to the plugin.

All the settings are dropped, due to a crude key check in `mergeWithoutOverwrite`, which checks if the setting already exists using `target[key]!==undefined` instead of `Object.prototype.hasOwnProperty`.

Yes

This issue arises is https://github.com/gajus/eslint-plugin-jsdoc, where the `tagNamePreference` object expects a key of `constructor` to set the preference for `@constructor` tag in jsdoc. Downstream task is https://github.com/gajus/eslint-plugin-jsdoc/issues/537.

edg2s:
#### Prerequisites checklist

[ ] Documentation update
[ ] Add autofixing to a rule
[ ] Add a CLI option
[ ] Add something to the core
[ ] Other, please explain:

Only merge to target if the key fails the hasOwnProperty check, rather than just check for undefined.

Replace the `key === "__proto__"` check with a hasOwnProperty check for source.

We can still use an undefined check when deciding if to copy `source[key]` to `target[key]` as by this stage both have been checked with hasOwnProperty.

Fixes #13289


lib/cli-engine/config-array/config-array.js
```diff
@@ -139,13 +139,15 @@ function mergeWithoutOverwrite(target, source) {
     }
 
     for (const key of Object.keys(source)) {
-        if (key === "__proto__") {
+        if (!Object.prototype.hasOwnProperty.call(source, key)) {
```
mysticatea, line 142: The key cames from `Object.keys(source)`, so I guess this is always true?

edg2s: I think it all practical cases this is probably true, however if may be possible for this to fail if source has some complex inheritance.

Given we've been bitten once before, it seems reasonable to be overly cautious here.

edg2s: Looking at the implementation of Object. 
keys this is probably safe to remove. I assume the proto check is left over from when the loop was some differently...

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/keys

mysticatea: Hmm. I'm curious about your case. `Object.keys` is defined to enumerate only own properties: https://tc39.es/ecma262/#sec-object.keys

edg2s: Agreed.

kaicataldo:
Do you mind fixing the linting errors?

kaicataldo:
@mysticatea Do you mind taking a look again?

nzakas:
Closing due to age.

PR #13303, opened 2020-05-15
eqeqeq options for undefined

tristanisme:
#### Prerequisites checklist

- [X] ~The team has reached consensus on the changes proposed in this pull request.~ If not, I understand that the evaluation process will begin with this pull request and won't be merged until the team has reached consensus.

[ ] Documentation update
[ ] Add autofixing to a rule
[ ] Add a CLI option
[ ] Add something to the core
[ ] Other, please explain:

`eqeqeq`

More, but only when a new option is enabled.

New option.

`{ undefined: 'always' | 'never' | 'ignore' }`

```js
// Eg1: Blocked by `undefined: 'always'`
foo == undefined
foo != undefined
// Eg1: Blocked by `undefined: 'never'`
foo === undefined
foo !== undefined
```

Triple or double equal is already enforceable, but it is not possible to configure specifically for `undefined`.

It will be possible to configure `eqeqeq`'s specific handling of `undefined`.

**Why is this useful?**

Currently it's not possible to enforce this reasonably sensible policy.
1) Always use `===` and `!==`.
2) Except when dealing with `null` or `undefined`. Always use `== null`, so `null` and `undefined` mean the same thing.

Suppose our config is...
``` 
eqeqeq: ['error', 'always', { null: 'never' }],
```
Here's how the existing rules behaves...
```
// Blocks all of these (good)...
foo == bar
foo != bar
foo === null
foo == undefined

// Allows all of these...
foo === bar
foo !== bar
foo == null
foo === undefined // <--- Noooo!   ( ⚈̥̥̥̥̥́⌢⚈̥̥̥̥̥̀)
```
With `{ undefined: "never" }`, even if the programmer can do `== undefined`, at least `null` and `undefined` will be treated the same throughout the code. It causes bugs when some parts of the code discriminate but not others.


kaicataldo:
Thanks for the PR. To help us better evaluate this proposal, can you fill out the [template requested above](https://raw.githubusercontent.com/eslint/eslint/master/templates/rule-change-proposal.md)?

tristanisme:
Good questions @kaicataldo .

 > 1. How does the option handle `void 0`?

It does not prevent it. That should be okay though. Nobody in my org has added `== void 0` to the repo, but many have added `=== undefined` without realising they weren't meant to.

> 2. `== null` is a pretty standard convention for this use case. Is there a reason you don't want to use that convention instead?

Ah sorry, that is the convention I want to enforce. Currently `eqeqeq: ['error', 'always', { null: 'never' }],` still allows `=== undefined` however. I improved the PR's explanation.

tristanisme:
Not entirely happy with my code. I'll re-visit.

PR #13305, opened 2020-05-15
Fix: Stop path analyzer on unknown nodes

ilyavolodin:
#### Prerequisites checklist

[ ] Documentation update
[ ] Add autofixing to a rule
[ ] Add a CLI option
[ ] Add something to the core
[ ] Other, please explain:

Updated code path analyzer to bail if no known ESTree nodes are found in AST. I'm trying to get ESLint to work with syntaxes other then ESTree and this is one place where it throws an error, since all the AST nodes are hardcoded in code-path-analyzer and can't be supplied by the parser. Unfortunately, it would be very tricky to create a unit test for this, and since this is a one-line change that shouldn't affect happy path, I figured it would be OK without unit test. 

Nope

### DIFF
--- a/lib/linter/linter.js
+++ b/lib/linter/linter.js
@@ -938,7 +938,8 @@ function runRules(sourceCode, configuredRules, ruleMapper, parserOptions, parser
         });
     });
 
-    const eventGenerator = new CodePathAnalyzer(new NodeEventGenerator(emitter));
+    // only run code path analyzer if the top level node is "Program", skip otherwise
+    const eventGenerator = nodeQueue[0].node.type === "Program" ? new CodePathAnalyzer(new NodeEventGenerator(emitter)) : new NodeEventGenerator(emitter);
 
     nodeQueue.forEach(traversalInfo => {
         currentNode = traversalInfo.node;
--- a/tests/fixtures/parsers/linter-test-parsers.js
+++ b/tests/fixtures/parsers/linter-test-parsers.js
@@ -10,5 +10,6 @@ module.exports = {
     noLineError: require("./no-line-error"),
     enhancedParser2: require("./enhanced-parser2"),
     enhancedParser3: require("./enhanced-parser3"),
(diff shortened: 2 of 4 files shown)


mysticatea:
Thank you.

Hmm. I guess, the analyzer still may crash on a node that has the same type name as one of the ESTree nodes. I think better if we expose a flag for custom parsers to disable code path analysis and modify [lib/linter/linter.js#L941](https://github.com/eslint/eslint/blob/1f175338cba29960aab67a540f516051f9d428c8/lib/linter/linter.js#L941).

ilyavolodin:
Sure, that would probably be ideal way to fix this. This is just a very quick fix to unblock linting unknown nodes. Since a new flag from parser would require RFC and a bunch of code changes, could we merge this one in first, and I will create an RFC for supporting generic ASTs (was planning on doing that anyways), which will include this flag as well as few more things.

mysticatea:
My concern is that this patch is incomplete; if the root node is not Program and the tree contains the node that is the same type as any of ESTree nodes, then the analyzer still can crash. I'd like to stop instantiating the analyzer rather than modifying the inside of the analyzer.

How about, stop instantiating the analyzer if the root is not a Program?

ilyavolodin:
@mysticatea I updated the code to stop code path analysis all together if top level node is not `Program`. I also added unit test to cover this scenario. I also left the old code, just in case, top level node is Program, but followed by unknown nodes. Shouldn't hurt anything anyways.

mysticatea:
Thank you.

Looks the linter does nothing at all if the root node is not a Program. I guess it's not intentional.

ilyavolodin:
Ah, sorry about that. Not sure what I was thinking. Updated with the right code. Should be working fine now. Updated unit test also to verify that it's actually running rules.

kaicataldo:
Marking as accepted since we have the approval of 3/4 of the TSC.

PR #13318, opened 2020-05-18
Update: max-lines reporting loc improvement (refs #12334)

anikethsaha:
#### Prerequisites checklist

[ ] Documentation update
[ ] Add autofixing to a rule
[ ] Add a CLI option
[ ] Add something to the core
[ ] Other, please explain:

Currently `max-lines` reports `1:1` for any numbers of line. 
In this change, it will report the extra lines which are causing the error.

Example :

image

It will report from (loc.start) `2` and till `(loc.end`) `3` (total lines)
refs #12334

***

This can be a bit more red lines as it will goes till eof.

One other option is that to report only the line from where the rest lines are extra. like in this case it would be the Line 3.

Let me know which one to keep.

### DIFF
--- a/lib/rules/max-lines.js
+++ b/lib/rules/max-lines.js
@@ -131,8 +131,6 @@ module.exports = {
                     text
                 }));
 
-                let startLine = 0;
-
                 if (skipBlankLines) {
                     lines = lines.filter(l => l.text.trim() !== "");
                 }
@@ -149,23 +147,18 @@ module.exports = {
                     );
                 }
 
-
                 if (lines.length > max) {
-                    startLine = lines[max].lineNumber;
-                }
+                    const loc = {
+                        start: {
+                            line: lines[max].lineNumber,
+                            column: 0
+                        },
+                        end: {
+                            line: sourceCode.lines.length,
+                            column: lodash.last(sourceCode.lines).length
+                        }
+                    };
 
-                const loc = {
-                    start: {
-                        line: startLine,
-                        column: 0
-                    },
-                    end: {
-                        line: sourceCode.lines.length,
-                        column: lodash.last(sourceCode.lines).length
-                    }
-                };
-
-                if (lines.length > max) {
                     context.report({
                         loc,
                         messageId: "exceed",
--- a/lib/rules/max-lines.js
+++ b/lib/rules/max-lines.js
@@ -131,6 +131,8 @@ module.exports = {
                     text
                 }));
 
+                let startLine = 0;
+
                 if (skipBlankLines) {
                     lines = lines.filter(l => l.text.trim() !== "");
                 }
@@ -146,13 +148,20 @@ module.exports = {
                         l => !lodash.includes(commentLines, l.lineNumber)
                     );
                 }
+
+                lines.forEach((line, i) => {
+                    if (max + 1 === i + 1) {
+                        startLine = line.lineNumber;
+                    }
+                });
+
                 const loc = {
                     start: {
-                        line: max + 1,
+                        line: startLine,
                         column: 0
                     },
                     end: {
-                        line: lines.length,
+                        line: sourceCode.lines.length + 1,
                         column: 0
                     }
                 };
--- a/lib/rules/max-lines.js
+++ b/lib/rules/max-lines.js
@@ -149,11 +149,10 @@ module.exports = {
                     );
                 }
 
-                lines.forEach((line, i) => {
-                    if (max + 1 === i + 1) {
-                        startLine = line.lineNumber;
-                    }
-                });
+
+                if (lines.length > max) {
+                    startLine = lines[max].lineNumber;
+                }
 
                 const loc = {
                     start: {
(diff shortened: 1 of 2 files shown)


lib/rules/max-lines.js
```diff
@@ -127,14 +138,28 @@ module.exports = {
...
+                const loc = {
+                    start: {
+                        line: max + 1,
+                        column: 0
+                    },
+                    end: {
+                        line: lines.length,
+                        column: 0
+                    }
+                };
```
mdjermanovic, line 158: If the settings are `{ "skipComments": true, "skipBlankLines": true }`, the line at index `max` doesn't have to be the first line above the maximum allowed lines.

Also, `lines` were already filtered at this point, so `lines.length` doesn't have to be the last line in the source code.

For example, the current iteration on an example with comments and empty lines would look like this :

image

anikethsaha: in latest commit, the following is the result 


image

lib/rules/max-lines.js
```diff
@@ -127,14 +140,35 @@ module.exports = {
...
+                    lines = lines.filter(
+                        l => !lodash.includes(commentLines, l.lineNumber)
+                    );
                 }
 
+                lines.forEach((line, i) => {
+                    if (max + 1 === i + 1) {
+                        startLine = line.lineNumber;
+                    }
+                });
```
mdjermanovic, line 156: Could this be replaced with just `const startLine = lines[max].lineNumber;` inside `if (lines.length > max`

lib/rules/max-lines.js
```diff
@@ -127,14 +140,35 @@ module.exports = {
...
+
+                const loc = {
+                    start: {
+                        line: startLine,
+                        column: 0
+                    },
+                    end: {
+                        line: sourceCode.lines.length + 1,
+                        column: 0
+                    }
```
mdjermanovic, line 166: This is an invalid location. If we want to report the last location in the source code as the `end` loc, it could be something like:

```js
end: {
    line: sourceCode.lines.length,
    column: lodash.last(sourceCode.lines).length
}
```

anikethsaha:
update :

image

lib/rules/max-lines.js
```diff
@@ -127,14 +140,34 @@ module.exports = {
...
+                    lines = lines.filter(
+                        l => !lodash.includes(commentLines, l.lineNumber)
+                    );
+                }
+
 
-                    lines = lines.filter(l => !lodash.includes(commentLines, l.lineNumber));
+                if (lines.length > max) {
+                    startLine = lines[max].lineNumber;
                 }
```
mdjermanovic, line 155: Can we move this to the other `if (lines.length > max)` below:

```js
if (lines.length > max) {
  const startLine = lines[max].lineNumber;
  const loc = ...
  context.report(...)
}
```

kaicataldo:
@mdjermanovic Do you mind taking one more look?

tests/lib/rules/max-lines.js
```diff
@@ -147,17 +194,243 @@ ruleTester.run("max-lines", rule, {
...
+            ),
+            options: [{ max: 2 }],
+            errors: [
+                {
+                    messageId: "exceed",
+                    data: { max: 2, actual: 4 },
+                    line: 3,
+
+                    endLine: 4
+
```
mdjermanovic, line 242: Are these blank lines necessary (the same applies to other tests below)?

tests/lib/rules/max-lines.js
```diff
@@ -147,17 +194,243 @@ ruleTester.run("max-lines", rule, {
...
+        },
+        {
+            code: ["", "var x = '';", "", "// comment", "", "var b = '',", "c,", "d,", "e", "", "// comment"].join("\n"),
+            options: [{ max: 2, skipComments: true, skipBlankLines: true }],
+            errors: [{
+                data: { max: 2, actual: 5 },
+                messageId: "exceed",
+                line: 7,
+                endLine: 11
+            }]
```
mdjermanovic, line 432: It might be nice to have error properties ordered like in all other tests above.

tests/lib/rules/max-lines.js
```diff
@@ -79,17 +77,38 @@ ruleTester.run("max-lines", rule, {
         {
             code: "var xyz;\nvar xyz;\nvar xyz;",
             options: [2],
-            errors: [{ messageId: "exceed", data: { max: 2, actual: 3 } }]
+            errors: [
+                {
+                    messageId: "exceed",
+                    data: { max: 2, actual: 3 },
+                    line: 3,
+                    endLine: 3
```
mdjermanovic, line 85: Can we also add `column` and `endColumn` to all invalid tests?

anikethsaha: Those would be always `1` so I skipped them as that is a known assertion. 

WDYT ?

mdjermanovic: I think it would be good to verify that at least in some tests. Also, `endColumn` shouldn't be `1` for code that doesn't end with a linebreak?

PR #13354, opened 2020-05-25
Fix: no-unused-vars updated location to last reference (fixes #13181)

### LINKED ISSUE #13181: [no-unused-vars]: reporter should report the last reference instead of declaration
`no-unused-vars`

same

default

Currently, the reporter reports the location in a declaration of a variable which is fine but not for all cases. 

`case 1`
```js
const unusedVar = 1; // reporting here is ok
```

`case2`
```js
let a = 1 // currently its reporting here,  reporting here is not ok as `a` has references though its being self referred 
a = a + 1;
a = a + 2;
a = a + 3; // it should report here
```

In this way, it makes easy to track the error if any.
Another advantage of reporting this way is many editors do provide shortcut keys to jump to the declaration of any variable/function but the opposite is not true for some. 

ref : https://github.com/eslint/eslint/issues/12871, https://github.com/eslint/eslint/issues/13172

reports in the variable declaration

reports in the last reference of the variable if the variable is being self-referred

Yes

anikethsaha:
#### Prerequisites checklist

[ ] Documentation update
[ ] Add autofixing to a rule
[ ] Add a CLI option
[ ] Add something to the core
[ ] Other, please explain:

fixes #13181
Updated the location for `no-unused-vars` to report at the last reference of a variable when not being used.

image
image

Should this be `Update` instead of `Fix` ?


lib/rules/no-unused-vars.js
```diff
@@ -618,13 +618,26 @@ module.exports = {
...
                     // Report the first declaration.
                     if (unusedVar.defs.length > 0) {
-                        context.report({
-                            node: unusedVar.identifiers[0],
-                            messageId: "unusedVar",
-                            data: unusedVar.references.some(ref => ref.isWrite())
-                                ? getAssignedMessageData(unusedVar)
-                                : getDefinedMessageData(unusedVar)
-                        });
+                        if (unusedVar.references.length > 0) {
```
nzakas, line 621: It looks like all of the arguments are the same except for node. Instead of duplicating all the other properties, can you maybe use a ternary expression ( foo ? bar : baz ) when assigning a value to “node”?

PR #13408, opened 2020-06-10
Update: rename id-blacklist to id-denylist (fixes #13407)

### LINKED ISSUE #13407: Remove use of blacklist and whitelist
### Summary
Could you update your API terminology for id-whitelist and id-blacklist to be more inclusive? This could be a breaking change or you could immediately accept additional properties alongside the existing ones and phase them out at the next scheduled release.

I'm not sure whether users need to use these terms in order to use your API, but I see them in the changelog and in the code.

### Motivation
The Code of Conduct says to use welcome and inclusive language and there's certainly more inclusive language that could be used for this part of the api. In addition, you're in the position of setting an example with the terminology that you choose to use. This post does a good job of detailing why the terms derive from prejudice: https://www.ncsc.gov.uk/blog-post/terminology-its-not-black-and-white

### Additional Context
If id-allowlist and id-denylist don't fit exactly, alternative API properties could be excluded-ids, permitted-ids

Please note, this is my first issue request on a public API

Thank you for your understanding of the significance of this issue

Prior conversation on other repositories:

rails/rails#33677
FullHuman/purgecss#428
getsentry/sentry-javascript/issues/2640

kaicataldo:
#### Prerequisites checklist

[ ] Documentation update
[ ] Add autofixing to a rule
[ ] Add a CLI option
[ ] Add something to the core
[ ] Other, please explain:

fixes #13407

This PR renames `id-blacklist` to `id-denylist` in a backwards-compatible way. Additionally, this change removes all instances of the term "blacklist" in our code.

As a follow up, we'll probably want to add a redirect on our website for the documentation.

1. Note that this word shows up in other projects that we have inlined in our repor for performance testing. Should we change those as well?
1. Have I missed anything?


lib/rules/index.js
```diff
@@ -56,7 +56,13 @@ module.exports = new LazyLoadingRuleMap(Object.entries({
     "grouped-accessor-pairs": () => require("./grouped-accessor-pairs"),
     "guard-for-in": () => require("./guard-for-in"),
     "handle-callback-err": () => require("./handle-callback-err"),
-    "id-blacklist": () => require("./id-blacklist"),
+
+    /*
+     * TODO: Remove in next major release.
+     * https://github.com/eslint/eslint/issues/13407
+     */
+    "id-blacklist": () => require("./id-denylist"),
```
mysticatea, line 64: I'm not sure if we can remove this rule because of [our deprecation policy](https://eslint.org/docs/user-guide/rule-deprecation).

Also, `rule.meta.deprecated` should be `true` for this rule.

kaicataldo: Fair point on not being able to remove the rule call-through. Will remove that comment.

In my mind, this isn't a deprecation as much as a rename, so I'm not sure it makes sense to mark it as deprecated. Since the API of the rule itself isn't changing, all existing configurations should still work.

nzakas: I think the question here is what the expected behavior of the user is. If you are expecting people to update their configurations (which seems what you're expecting given the comment about removing the rule in the future), then you need to somehow indicate to the user. The only way we have to indicate that right now is through marking the existing rule as deprecated. We can't really plan on removing a rule without warning people that it's going away.

kaicataldo: @mysticatea @nzakas Does removing this comment and leaving the code as-is resolve your concerns?

nzakas: For me, yes. If we don’t plan on removing the rule in the future then I think it’s fine.

kaicataldo: Updated!

mysticatea: Yes. If you are sure that it doesn't affect our scripts for our website, I don't mind the `deprecated` flag.

nzakas: @mysticatea I think what @kaicataldo and I are saying is that we don't need to set `meta.deprecated` to `true` if we don't intend to remove it.

mysticatea: I think this is the first time that a rule exists but `<rule-id>.md` doesn't exist. Therefore, my concern is the scripts about website, but I'm not familiar with the scripts. I'm OK if the scripts are OK.

kaicataldo: That's a valid concern - I was thinking we could just create a redirect in the website repo. Thoughts?

kaicataldo:
@mysticatea @btmills @nzakas This isn't technically accepted yet - if you support this, would you mind adding a :+1: to my original comment?

Alhadis:
#BlackListsMatter

Issue #7787, opened 2016-12-17
`func-call-spacing` fix creating bugs


isiahmeadows:
**Edit:** fix post to reflect real bug.

In the latest ESLint (v3.12.2), the `func-call-spacing` fix is really unhelpful for multiline call expressions (overlap with `no-unexpected-multiline`), especially when going without semicolons.

```js
// Short snippet
this.cancelled.add(request)
this.decrement(request)
(0, request.reject)(new api.Cancel())
```

These are likely meant to be separate statements, with the below being (99.99% of the time) what was meant:

```js
// Short snippet
this.cancelled.add(request)
this.decrement(request)
;(0, request.reject)(new api.Cancel())
```

Instead, `func-call-spacing` very unhelpfully fixes it to this, leading to runtime errors nearly every time in practice:

```js
// Short snippet
this.cancelled.add(request)
this.decrement(request)(0, request.reject)(new api.Cancel())
```

---

Here's the same issue, but with trailing semicolons:

Input:

```js
// Short snippet
this.cancelled.add(request);
this.decrement(request)
(0, request.reject)(new api.Cancel());
```

Meant:

```js
// Short snippet
this.cancelled.add(request);
this.decrement(request);
(0, request.reject)(new api.Cancel());
```

"Fixed":

```js
// Short snippet
this.cancelled.add(request);
this.decrement(request)(0, request.reject)(new api.Cancel());
```

-----

Any chance this could be fixed to be a little *less* over-zealous?

isiahmeadows:
This has been causing me way too much of a productivity hit due to my editor autofixing on save.

platinumazure:
I agree, I think `no-unexpected-multiline` is one of those rules where the correct fix can be one of two things (either add a necessary semicolon, or put the lines together). We shouldn't be assuming one or the other in the fixer. I would be :+1: for removing the fixer here and letting the user choose.

This is similar (IMO) to why we have no fixer for `no-useless-escape`. In that case, we can't know if the user wanted a backslash (in which case the escape should be doubled up) or if the escape truly was useless (in which case the backslash can be removed).

isiahmeadows:
Actually, I'm working on narrowing down a more precise repro. It's actually a fix for another rule conflicting with `no-unexpected-multilline`.

mysticatea:
I'm confusing.
`no-unexpected-multiline` is not fixable.

isiahmeadows:
[...]

PR #13416, opened 2020-06-13
Update: optional chaining support (fixes #12642)

### LINKED ISSUE #12642: Optional Chaining and Nullish coalescing Operator arrived at Stage 4
6.7

Support two new ES features: Optional Chaining and Nullish coalescing Operator.

- [x] **ESTree:** https://github.com/estree/estree/pull/204 and https://github.com/estree/estree/pull/211
- [x] **Acorn:** https://github.com/acornjs/acorn/pull/890 and https://github.com/acornjs/acorn/pull/891
- [x] **eslint-visitor-keys:** https://github.com/eslint/eslint-visitor-keys/pull/12
- [ ] **Espree:** https://github.com/eslint/espree/pull/441 and https://github.com/eslint/espree/pull/446
- [ ] **ESLint:** https://github.com/eslint/eslint/pull/13196 and https://github.com/eslint/eslint/pull/13416

*(Template inspired by Issue 12629)*

mysticatea:
#### Prerequisites checklist

[X] Other, please explain: Add support for optional chaining syntax. Fixes #12642.

This PR adds support for optional chaining syntax into ESLint. It includes the following changes:

- [x] ESTree update: https://github.com/estree/estree/pull/204
- [x] Acorn update: https://github.com/acornjs/acorn/pull/891
- [ ] `esquery` update: https://github.com/estools/estraverse/pull/113 and for it.
- [x] `eslint-utils` update: https://github.com/mysticatea/eslint-utils/pull/8
- [x] `eslint-visitor-keys` update: https://github.com/eslint/eslint-visitor-keys/pull/12
- [x] `espree` update: https://github.com/eslint/espree/pull/446 (not released yet)
- [x] Code path analysis update: a574e27867246d5bea56d1e8d82a6af24e28fd5d ... It recognizes short-circuiting of optional chaining.
- Rules update:
  - [x] `accessor-pairs` 9125cf19af4681401aa29976b66eb341f6b3e65e ... It handles `ChainExpression` nodes.
  - [x] `array-callback-return` 102c4dba32bd6217f2ce36d23242e4837ab79e08 ... It handles `ChainExpression` nodes.
  - [x] `camelcase` 3de0a1cf5ca1ee0157f418750f6f44b599cb227d ... Some tests are added.
  - [x] `computed-property-spacing` f2dc7f28a8bc65475770b6476586a410f74ffdbb ... Some tests are added.
  - [x] `constructor-super` 80e5e92c22eaef17971308b326e15ad71a7bd38e ... It handles `ChainExpression` nodes.
  - [x] `dot-location` 33e49868164cc5004b09bf2214df763d77cb5d3d, b91d17b9f6bd3e9382537a25e0152a5f906ab154 ... It fixes code with `?.` correctly.
  - [x] `dot-notation` f66ce6ccca11c12ee4d5bac15cd47d76178d9301 ... It fixes code with `?.` correctly.
  - [x] `func-call-spacing` 31412c21998a6590548e4d9ee8646798b6fc35b1 ... It fixes code with ?. correctly.
  - [x] `func-name-matching` 206d25ffe9ab2bd36e027bd1bb36fb9d45656d0b ... It handles `ChainExpression` nodes.
  - [x] `getter-return` 276526ba3c8888c6e499c04a20e04f20e9fc134f ... Some tests are added.
  - [x] `global-require` 2824b224d2c8708f35e41aaf411bcaba175e299f ... It handles `ChainExpression` nodes.
  - [x] `indent` 9a14a09fa25a01915024d3ea587df26026074416 ... It handles `ChainExpression` nodes and `?.` tokens correctly.
  - [x] `new-cap` 7574cfd08076456cbedb93f2014d3a5dc6d8999b ... It handles `ChainExpression` nodes.
  - [x] `newline-per-chained-call` ed71bd4361acde661ef78cb0e12acc26cbbadc40 ... It handles `ChainExpression` nodes and `?.` tokens correctly.
  - [x] `no-alert` 9fd0338300c23a24a3143fefc17fc8e55899dd7a ... It handles `ChainExpression` nodes.
  - [x] `no-eval` abcceae115be04f23c71a49b5f3c67199ffff648 ... It handles `eval?.(code)` as indirect eval as per the spec. It handles `ChainExpression` nodes.
  - [x] `no-extend-native` 45c5aea177771982f8be793199c97fff42ff046c ... It handles `ChainExpression` nodes. And it fixes a bug of edge cases.
  - [x] `no-extra-bind` 38592d26364b9d98456cb140576e46a46ec8b77a ... It handles `ChainExpression` nodes and `?.` tokens correctly. And it fixes a bug of edge cases.
  - [x] `no-extra-boolean-cast` d0c2f829a425fd09db73ef875588aa6435dec3b0 ... It handles `ChainExpression` nodes.
  - [x] `no-extra-parens` a395b1017558081185de0232e68e1ce829a66ed6 ... It recognizes necessary parentheses of avoiding syntax errors (e.g., `new (obj?.p)()`, ``(obj?.p)`template` ``, etc) and necessary parentheses of disconnecting chaining. It handles `ChainExpression` nodes.
  - [x] `no-implicit-coercion` e550194fa2e043197fbedb13e7d692145f9a3596 ... It creates the proper error message for `~foo?.indexOf(bar)`-like code. It handles `ChainExpression` nodes.
  - [x] `no-implicit-eval` 248faf16410d68b3190ff3e2a611c8170dccae63 ... It handles `ChainExpression` nodes.
  - [x] `no-import-assign` b3dee6751be0502a1ca486906b4107a43d08cc55, f65df7503b252bb94896f6e51e4be37495f36c79 ... It handles `ChainExpression` nodes.
  - [x] `no-magic-numbers` c63d3c143d509deadcfe9bf8baede2e026b39789 ... It handles `ChainExpression` nodes.
  - [x] `no-obj-calls` e13059f6179acdd3bf8b0d18a60696920cdab328 ... It handles `ChainExpression` nodes.
  - [x] `no-prototype-builtins` 5286f830cf23c6f884f000384e2b84fd5c3518c7 ... It handles `ChainExpression` nodes.
  - [ ] `no-restricted-syntax` 8a0097a2100c0cdf86fc3a2d717735f8f8dda358 ... Some tests are added. We need to fix `esquery` to support optional chaining with some selectors (https://github.com/estools/estraverse/pull/113).
  - [x] `no-self-assign` d12b2bea7aa09e4605aa802dd93dc1d3b078e363 ... It handles `ChainExpression` nodes.
  - [x] `no-setter-return` a31e04c7511bea56e6fd8b9a1339e9ccc1cd7bdf ... It handles `ChainExpression` nodes.
  - [x] `no-unexpected-multiline` 6c6bd958760c7ad070df9dd94f82ce1154010d2d, 776e96b88c3b96f118477b3f6e8f2f09f05d99e9 ... It handles `?.` tokens correctly. Should this rule ignore the case the previous line ends with `?.` token?
  - [x] `no-unused-expression` 6c00e33b8fa94f529f5a46a60632c9de6966b5f9, 9a91ba196f179e2be27eab57cd90c237a18b9f51 ... It recognizes `ChainExpression` nodes. I updated this rule to enumerate disallowed node types rather than allowed node types in order to ignore unknown nodes.
  - [x] `no-useless-call` f7fd797f4b784a8510646b75cae53cdab02d1a17 ... It handles `ChainExpression` nodes.
  - [x] `no-whitespace-before-property` b5fa99ac0a4f41985bd4723fbab1b39d1d2c92d2 ... It handles `?.` tokens correctly.
  - [x] `operator-assignment` 6468c0c994c97e364234f436f83a504ca8d05346 ... It handles `ChainExpression` nodes.
  - [x] `padding-line-between-statements` 57405388aab12c853b2602562b563d925a4757cc ... It handles `ChainExpression` nodes.
  - [x] `prefer-arrow-callback` d0fc1b1c35ab9dd6c72acff5280847dd7e1f76cb ... It handles `ChainExpression` nodes.
  - [x] `prefer-destructuring` a443a1e4faf3ba67a29b38cf5ab3fc4ada5d20f0 ... Some tests are added.
  - [x] `prefer-exponentiation-operator` ca0b2a8729de5347372132985351f50346f86046 ... It handles `ChainExpression` nodes.
  - [x] `prefer-numeric-literals` d978d368739b5c80da6a5882215d35d223670091 ... It handles `ChainExpression` nodes.
  - [x] `prefer-promise-reject-errors` 5c7368cb9b481c762b15cca8f897b26e96af46c6 ... It handles `ChainExpression` nodes.
  - [x] `prefer-regex-literals` 651bac30a6cdebeeeda50d0696856be35107753e ... It handles `ChainExpression` nodes.
  - [x] `prefer-spread` a3cbe6e7862f1c8002cfe3dc624f89edec1e3ba5 ... It handles `ChainExpression` nodes.
  - [x] `use-isnan` 17ec3d05a1023b07c641b6c0bdbbaaadddd5e959 ... It handles `ChainExpression` nodes.
  - [x] `wrap-iife` 233e73709c41ba57a800c8068a1948bd09b3e2a9 ... It handles `ChainExpression` nodes.
  - [x] `yoda` 89b65bf969778bcde7c9e19a44d4c689a87b1834 ... It handles `ChainExpression` nodes.

- The most rule updates are to handle `ChainExpression` nodes that are at the `callee|object` properties. And some rules needed fixing autofix logic for the `?.` token.
- How about `dot-location` for `obj?.[key]`? The rule ignores computed properties currently, but we can insert line breaks between `?.` and `[`, so we can choose either `obj?.\n[key]` or `obj\n?.[key]`. I guess we have to add a new option for that.

Remaining steps:

1. Merge https://github.com/eslint/espree/pull/446.
1. Release a new version of `espree`.
1. Update this PR with the new `espree`.


ota-meshi:
Hi @mysticatea! Thank you for working on support for Optional chaining!

I have found that using the `:nth-child(1)` selector option with the `no-restricted-syntax` rule causes an error. The `no-restricted-syntax` rule didn't seem to be on the list, so I'll share it just in case.

If you add the following test case to `tests/lib/rules/no-restricted-syntax.js` and run it, you will get an error.

```js
{ code: "a?.b", options: [":nth-child(1)"], parserOptions: { ecmaVersion: 2020 } },
```

```
  1) no-restricted-syntax
       valid
         a?.b:
     TypeError: Cannot read property 'length' of undefined
Occurred while linting <input>:1
      at f (node_modules/esquery/dist/esquery.min.js:1:32431)
      at Function.l [as matches] (node_modules/esquery/dist/esquery.min.js:1:30995)
      at NodeEventGenerator.applySelector (lib/linter/node-event-generator.js:253:21)
      at NodeEventGenerator.applySelectors (lib/linter/node-event-generator.js:281:22)
      at NodeEventGenerator.enterNode (lib/linter/node-event-generator.js:297:14)
      at CodePathAnalyzer.enterNode (lib/linter/code-path-analysis/code-path-analyzer.js:664:23)
      at /Users/otayosuke/git/eslint/lib/linter/linter.js:949:32
      at runRules (lib/linter/linter.js:944:15)
      at Linter._verifyWithoutProcessors (lib/linter/linter.js:1170:31)
      at Linter.verify (lib/linter/linter.js:1233:21)
      at runRuleForItem (lib/rule-tester/rule-tester.js:556:37)
      at testValidTemplate (lib/rule-tester/rule-tester.js:605:28)
      at Context.<anonymous> (lib/rule-tester/rule-tester.js:866:25)
      at processImmediate (internal/timers.js:456:21)
```

It seems that an error occurs when the AST contains a node that `estraverse` does not know.

---

It's not related to the Optional chaining, but I get the same error in JSX.

```js
{ code: "jsx = <div />", options: [":nth-child(1)"], parserOptions: { ecmaFeatures: { jsx: true } } }
```

[Demo](https://eslint.org/demo#eyJ0ZXh0IjoiLyogZXNsaW50IG5vLXJlc3RyaWN0ZWQtc3ludGF4OiBbMiwgOm50aC1jaGlsZCgxKV0gKi9cbmpzeCA9IDxkaXYgLz4iLCJvcHRpb25zIjp7InBhcnNlck9wdGlvbnMiOnsiZWNtYVZlcnNpb24iOjExLCJzb3VyY2VUeXBlIjoibW9kdWxlIiwiZWNtYUZlYXR1cmVzIjp7ImpzeCI6dHJ1ZX19LCJydWxlcyI6e30sImVudiI6e319fQ==)

Should ESLint core rules support JSX? If so I will open a new issue.

mysticatea:
@ota-meshi Thank you for your report!

It's likely a bug in esquery. Hmm, we need to add custom visitor keys support to esquery because our custom parsers can use custom visitor keys but esquery doesn't support it. I'd like to work on that as separate from this PR.

ota-meshi:
@mysticatea Thank you for your response!

If can add custom visitor keys support to esquery as you say, more problems can be solved. I think it's a better solution.
I will open a new issue later.

lib/rules/no-unused-expressions.js
```diff
@@ -101,40 +117,58 @@ module.exports = {
...
+         * Unknown nodes are handled as `false`, then this rule ignores those.
          */
-        function isValidExpression(node) {
-            if (allowTernary) {
-
-                // Recursive check for ternary and logical expressions
-                if (node.type === "ConditionalExpression") {
-                    return isValidExpression(node.consequent) && isValidExpression(node.alternate);
+        const Checker = {
+            __proto__: null,
```
mdjermanovic, line 124: Should we use this, I think it's Annex B?

(though it does seem like a nice feature for object literals)

mdjermanovic:
I believe `no-import-assign` wouldn't catch this:

```js
import * as ns from 'mod'; 
delete ns?.named
```

I can't install to test, though. Probably because `github:eslint/eslint-visitor-keys#add-chain-expression` doesn't exist anymore (tried to change it locally to `1.3.0` but `install` still looks for that branch, probably because it's still used from the `espree` branch).

mdjermanovic:
`radix` rule:

```js
(Number?.parseInt)(foo);
```

(off-topic: we should probably have a rule to disallow such broken chains).

mdjermanovic:
> `no-extra-parens` [a395b10](https://github.com/eslint/eslint/commit/a395b1017558081185de0232e68e1ce829a66ed6) ... It recognizes necessary parentheses of avoiding syntax errors (e.g., `new (obj?.p)()`, `` (obj?.p)`template`  ``, etc)

I don't think we're checking parens around template tags, yet. We should definitely account for optional chains when we fix that (though it seems meaningless to have optional chain as a tag).

> no-unexpected-multiline 6c6bd95 ... It handles ?. tokens correctly. Should this rule ignore the case the previous line ends with ?. token?

I think it should ignore both `?.\n(` and `?.\n[`, it doesn't look like a mistake.

mdjermanovic:
`constructor-super`: it most likely doesn't make sense to have `ChainExpression` in `extends`, but it should be probably recognized as a constructor in `function isPossibleConstructor()`.

`func-call-spacing`: looks like it will check spaces between `?.` and `(`.

mysticatea:
Thank you for your review. I'm working on addressing those.

> (off-topic: we should probably have a rule to disallow such broken chains).

Yes, I thought samely and I have a proposal memo personally: https://gist.github.com/mysticatea/a49a6fc1f6037b3c9f0335195608fba2
I will post it later...

mysticatea:
> I don't think we're checking parens around template tags, yet. We should definitely account for optional chains when we fix that (though it seems meaningless to have optional chain as a tag).

Oh, I didn't know it. The following test case will help to implement the check in the future.

https://github.com/eslint/eslint/blob/9939c4ccc84e6b61e60b024b82f1457c847581ff/tests/lib/rules/no-extra-parens.js#L630

lib/rules/dot-location.js
```diff
@@ -52,31 +52,36 @@ module.exports = {
...
                         node,
-                        loc: dot.loc,
+                        loc: dotToken.loc,
                         messageId: "expectedDotAfterObject",
-                        fix: fixer => fixer.replaceTextRange([tokenBeforeDot.range[1], property.range[0]], `${neededTextAfterToken}.${textBeforeDot}${textAfterDot}`)
+                        *fix(fixer) {
+                            if (astUtils.isDecimalIntegerNumericToken(tokenBeforeDot)) {
+                                yield fixer.insertTextAfter(tokenBeforeDot, " ");
+                            }
+                            yield fixer.insertTextAfter(tokenBeforeDot, dotToken.value);
```
mdjermanovic, line 71: Could this sometimes become `1. ` instead of `1 .`? I think these are going to be sorted [here](https://github.com/eslint/eslint/blob/b7d79b12eb55f220f86367d52f9f5a2f503fd4c3/lib/linter/report-translator.js#L143) and I'm not sure if the min required Node version guarantees stable sort.

mysticatea: Good catch. Looks like V8 7.0 has introduced stable sort, so Node.js is stable sort since 11.0.0.

mdjermanovic:
`no-extra-boolean-cast` false negative `if (Boolean?.(foo));`

lib/rules/no-eval.js
```diff
@@ -230,7 +195,12 @@ module.exports = {
                 "CallExpression:exit"(node) {
                     const callee = node.callee;
 
-                    if (isIdentifier(callee, "eval")) {
+                    /*
+                     * Optional call (`eval?.("code")`) is not direct eval.
+                     * The direct eval is only step 6.a.vi of https://tc39.es/ecma262/#sec-function-calls-runtime-semantics-evaluation
+                     * But the optional call is https://tc39.es/ecma262/#sec-optional-chaining-chain-evaluation
+                     */
```
mdjermanovic, line 202: It works as direct eval in latest Chrome (could be Chrome bug?):

```js
var a = 5;
((a) => { eval?.("console.log(a)") })(10); // 10
```

mysticatea: I reported:

- V8: https://bugs.chromium.org/p/v8/issues/detail?id=10630
- Firefox: https://bugzilla.mozilla.org/show_bug.cgi?id=1647169
- test262: https://github.com/tc39/test262/pull/2667

lib/rules/func-call-spacing.js
```diff
@@ -161,7 +167,20 @@ module.exports = {
                     },
                     messageId: "unexpectedNewline",
                     fix(fixer) {
-                        return fixer.replaceTextRange([leftToken.range[1], rightToken.range[0]], " ");
+                        const range = [leftToken.range[1], rightToken.range[0]];
```
mdjermanovic, line 170: I believe it's a bug that this scenario still has autofix, and that it should have been removed while fixing #7787 issue (it hides bugs in the code reported by `no-unexpected-multiline`), so I'm not sure should we rather just remove this `fix()` now instead of upgrading it for optional calls.

lib/rules/no-extra-parens.js
```diff
@@ -100,10 +100,18 @@ module.exports = {
          * @private
          */
         function isImmediateFunctionPrototypeMethodCall(node) {
-            return node.type === "CallExpression" &&
-                node.callee.type === "MemberExpression" &&
-                node.callee.object.type === "FunctionExpression" &&
-                ["call", "apply"].includes(astUtils.getStaticPropertyName(node.callee));
+            const callNode = astUtils.skipChainExpression(node);
```
mdjermanovic, line 103: I think we should do the same in `function isIIFE`, too.

False positive: `var x = (function () {}?.());`

mdjermanovic:
`astUtils.isDefaultThisBinding`:

```js
"use strict";

obj.foo = (function() {
    return function foo() {
        this; // no-invalid-this false positive
    };
})?.();
```

lib/rules/no-implicit-coercion.js
```diff
@@ -246,7 +248,13 @@ module.exports = {
...
                 operatorAllowed = options.allow.indexOf("~") >= 0;
                 if (!operatorAllowed && options.boolean && isBinaryNegatingOfIndexOf(node)) {
-                    const recommendation = `${sourceCode.getText(node.argument)} !== -1`;
+
+                    /*
+                     * `foo?.indexOf(bar) !== -1` will be true (== found) if the `foo` is nullish. So use `>= 0` in that case.
+                     * Or maybe we should recommend `foo?.includes(bar)` simply.
+                     */
+                    const comparison = node.argument.type === "ChainExpression" ? ">= 0" : "!== -1";
+                    const recommendation = `${sourceCode.getText(node.argument)} ${comparison}`;
```
mdjermanovic, line 257: I think `>= 0` is fine for this rule because recommending `includes` might be beyond its responsibility, and it can't know if the user wants to use ES2016+ array methods.

tests/lib/rules/prefer-spread.js
```diff
@@ -39,7 +39,11 @@ ruleTester.run("prefer-spread", rule, {
         // ignores incomplete things.
         "foo.apply();",
         "obj.foo.apply();",
-        "obj.foo.apply(obj, ...args)"
+        "obj.foo.apply(obj, ...args)",
+
+        // Optional chaining
+        "(a?.b).c.foo.apply(a?.b.c, args);",
+        "a?.b.c.foo.apply((a?.b).c, args);"
```
mdjermanovic, line 46: This might be actually a false negative?

Those parens don't make much sense, but something like the following looks like a valid use of `?.` since it also skips evaluation of arguments:

```js
/* eslint prefer-spread: error */

a?.b.c.apply(a.b, args); // false negative?
```

(not sure is it worth fixing this, as it seems complex)

mdjermanovic:
`array-callback-return` false negative:

```js
arr.map((() => {
    // ...
    return () => {};
})?.());
```

confirmed `radix` false negative:

```js
(Number?.parseInt)(a);
```

lib/rules/func-call-spacing.js
mysticatea: Good catch. I removed autofix that remove line breaks of non-optional calls to avoid hiding `no-unexpected-multiline` errors. I remained about optional calls as-is because `no-unexpected-multiline` doesn't report in that case.

mdjermanovic: :+1: 

There's another alternative for handling comments, but we could discuss that later in #13411

tests/lib/rules/prefer-spread.js
mdjermanovic: If we decide to fix this, the same applies to `no-useless-call`

```js
a?.b.c.call(a.b); // false negative?

a?.b.c.apply(a.b, [1, 2, 3]);  // false negative?
```

lib/rules/no-extra-bind.js
```diff
@@ -61,24 +61,55 @@ module.exports = {
...
+                            // property name or `]` token.
+                            sourceCode.getLastToken(memberNode)
+                        ],
+                        [
+
+                            // `(` token of arguments.
+                            sourceCode.getTokenAfter(
+                                memberNode,
+                                astUtils.isNotClosingParenToken
+                            ),
```
mdjermanovic, line 98: Nice find! Maybe we could just add to the comment that this may also be a `?.` token (e.g., as in `(() => {}).bind?.(foo)`)

lib/rules/array-callback-return.js
```diff
@@ -79,7 +76,14 @@ function getArrayMethodName(node) {
             case "ReturnStatement": {
                 const func = astUtils.getUpperFunction(parent);
 
-                if (func === null || !astUtils.isCallee(func)) {
+                if (
+                    func === null ||
+                    !astUtils.isCallee(
+                        func.parent.type === "ChainExpression"
```
mdjermanovic, line 82: I think this can never happen? 

The existing `func === null || !astUtils.isCallee(func)` condition seems fine, `case "ChainExpression"` will handle `func.parent.parent` if needed.

lib/rules/dot-notation.js
```diff
@@ -141,29 +149,24 @@ module.exports = {
...
-                                // Don't perform any fixes if there are comments between the dot and the property name.
-                                return null;
-                            }
+                        *fix(fixer) {
+                            const dotToken = sourceCode.getTokenBefore(node.property);
 
+                            // A statement that starts with `let[` is parsed as a destructuring variable declaration, not a MemberExpression.
                             if (node.object.type === "Identifier" && node.object.name === "let") {
+                                return; // eslint-disable-line eslint-plugin/fixer-return -- false positive
+                            }
```
mdjermanovic, line 158: Minor edge case: this might be fine if it's optional `let?.class` -> `let?.["class"]`

mdjermanovic:
`new-cap`:

```js
/* eslint new-cap: ["error", { "properties": false }] */

(foo?.Bar)(); // false positive

new (foo?.bar)();  // false positive
```

and a minor note:  `function report()` could also skip chain expression to get property loc.

mdjermanovic:
`wrap-iife` (`isCalleeOfNewExpression`):

```js
/* eslint wrap-iife: ["error", "inside"] */

new (function () {} ?.());
```

would be autofixed to syntax error:

```js
/* eslint wrap-iife: ["error", "inside"] */

new (function () {}) ?.();
```

lib/rules/no-self-assign.js
```diff
@@ -17,56 +17,6 @@ const astUtils = require("./utils/ast-utils");
...
 const SPACES = /\s+/gu;
 
-/**
- * Checks whether the property of 2 given member expression nodes are the same
- * property or not.
- * @param {ASTNode} left A member expression node to check.
- * @param {ASTNode} right Another member expression node to check.
- * @returns {boolean} `true` if the member expressions have the same property.
- */
-function isSameProperty(left, right) {
```
mdjermanovic, line 27: The change can produce new warnings, e.g., `a[b.c] = a[b.c];`. I think its fine.

lib/rules/operator-assignment.js
```diff
@@ -40,45 +40,6 @@ function isNonCommutativeOperatorWithShorthand(operator) {
...
-            return a.value === b.value;
-
-        case "MemberExpression":
-
-            /*
-             * x[0] = x[0]
-             * x[y] = x[y]
-             * x.y = x.y
-             */
-            return same(a.object, b.object) && same(a.property, b.property);
```
mdjermanovic, line 72: This had a bug `a[b] = a.b + 1;`

lib/rules/operator-assignment.js
```diff
@@ -148,12 +109,12 @@ module.exports = {
...
 
             if (isCommutativeOperatorWithShorthand(operator) || isNonCommutativeOperatorWithShorthand(operator)) {
-                if (same(left, expr.left)) {
+                if (astUtils.isSameReference(left, expr.left, true)) {
                     context.report({
                         node,
                         messageId: "replaced",
                         fix(fixer) {
-                            if (canBeFixed(left)) {
+                            if (canBeFixed(left) && canBeFixed(expr.left)) {
```
mdjermanovic, line 117: Is there a situation where `canBeFixed(left)` returns `true`, but `canBeFixed(expr.left)` returns `false`?

lib/linter/code-path-analysis/code-path-analyzer.js
```diff
@@ -244,6 +244,17 @@ function preprocess(analyzer, node) {
...
+        case "CallExpression":
+            if (parent.optional === true && parent.arguments.length >= 1 && parent.arguments[0] === node) {
+                state.makeOptionalRight();
+            }
+            break;
+        case "MemberExpression":
+            if (parent.optional === true && parent.property === node) {
+                state.makeOptionalRight();
+            }
+            break;
```
mdjermanovic, line 256: These might be missing cases where there's an optional call without arguments in the chain?

`a?.().b;`

image

there should be fork from `Identifier(a)`?

Similar: `a?.()(b);`

mdjermanovic:
`astUtils.couldBeError`

Given the function's "could be" name, and the fact that it returns `true` for expressions such as `a || 5`, it should probably return `true` for all `ChainExpression`s.

----

Some maybe off-topic notes:

Aside from chain expressions, this function works in line with its name, but I'm not sure is the logic that checks if there is at least one possible `Error` more suitable than checking if there are no certain non-`Error`s.

In particular, shouldn't `no-trow-literal` report `throw a || 5`?

Also not sure why this function doesn't handle `undefined` identifier itself instead of letting the rules do that, so for example `no-throw-literal` doesn't report `throw undefined || undefined`.

mdjermanovic:
`astUtils.isMethodWhichHasThisArg`:

```js
"use strict";

(arr?.map)(function() {
    this; // no-invalid-this false positive
}, obj);
```

also, for-loop in the function seems to be useless (there can be only 1 iteration)?

tests/lib/rules/prefer-spread.js
mysticatea: Currently, `prefer-spread` and `no-useless-call` use `astUtils.equalTokens()` to check if the callee and the first argument are samely or not. I think that we can use `astUtils.isSameReference()` instead. But I'm hesitating because it looks like a breaking change.

lib/rules/operator-assignment.js
mysticatea: I'm sorry, I cannot remember what I thought at that time... 😢 

Maybe, `a.b = a?.b + c` is reported and not fixed in the first implementation, but then I changed it and forgot to remove this line. I will change the rule to report `a.b = a?.b + c` later...

mdjermanovic: I'm not sure if `operator-assignment` should report `a.b = a?.b + c`, since it looks more like a possible error than a stylistic issue, because this will throw if `a` is nullish.

And, it will be reported by `no-unsafe-optional-chaining` (maybe in this case even with `disallowArithmeticOperators: false`).

tests/lib/rules/prefer-spread.js
mdjermanovic: Agreed, it seems best to not change anything now, and maybe discuss it later (after this PR).

lib/rules/operator-assignment.js
mysticatea: As like `no-useless-escape` rule, it's hard to determine the writer's intention from `a.b = a?.b + c`; The `a` may be nullable (then throw) or the `?.` may be useless (then equal to `a.b`). Therefore, I think that both `operator-assignment` and `no-unsafe-optional-chaining` should report that.

lib/linter/code-path-analysis/code-path-analyzer.js
mysticatea: Good catch!

mdjermanovic:
@mysticatea would it be possible to merge and release this without the `esquery`/`estraverse` update, or is it a blocker?

mysticatea:
> In particular, shouldn't no-trow-literal report throw a || 5?

I feel that the rule should report `a || 5`. We can open an issue for that.

> would it be possible to merge and release this without the esquery/estraverse update, or is it a blocker?

I don't think it's a blocker of this PR.
We might need to fork `esquery` to support `visitorKeys` in query...

----

@mdjermanovic I appreciate your exhaustive review. Thank you very much! I should address your review, but please tell me if not enough.

lib/rules/utils/ast-utils.js
```diff
@@ -464,14 +464,24 @@ function isArrayFromMethod(node) {
...
+            currentNode = currentNode.expression;
+            continue;
         }
+
+        if (currentNode.type === "MemberExpression" && !currentNode.computed) {
+            if (currentNode.property.type === "Identifier") {
+                return arrayMethodPattern.test(currentNode.property.name);
+            }
+
+            currentNode = currentNode.property;
```
mdjermanovic, line 480: I still think this loop is unnecessary, we're checking only one `property`. Also, `currentNode.property` cannot be `MemberExpression` at this point, so there will be no useful iterations afterward.

mysticatea: Really. I'm not sure what I thought.

mdjermanovic:
> > would it be possible to merge and release this without the esquery/estraverse update, or is it a blocker?
> 
> I don't think it's a blocker of this PR.

I'd also vote to merge this without the `esquery` update.

It seems that only some rarely used selectors (those that internally use `estraverse`) won't work well with chain expressions in some cases.

lib/rules/operator-assignment.js
mdjermanovic: Agreed!

tests/lib/rules/operator-assignment.js
```diff
@@ -398,6 +398,20 @@ ruleTester.run("operator-assignment", rule, {
...
+    },
+
+    // Optional chaining
+    {
+        code: "(obj?.a).b = (obj?.a).b + y",
+        output: null,
+        errors: EXPECTED_OPERATOR_ASSIGNMENT
+    },
+    {
+        code: "obj.a = obj?.a + b",
```
mdjermanovic, line 410: Should `no-self-assign` also report similar cases?

It uses `astUtils.isSameReference` so it will report `a.b.c = (a?.b).c;`

But it checks first if they're both `MemberExpression`, so it will not report `a.b = a?.b;`

mysticatea: Thank you! I updated `no-self-assign` rule to catch similar cases.

kaicataldo:
PR to upgrade Espree: https://github.com/eslint/eslint/pull/13501.

PR #13435, opened 2020-06-22
Fix: Replace Infinity with Number.MAX_SAFE_INTEGER (fixes #13427)

### LINKED ISSUE #13427: Config clone before validating kills Object data (7.3.0)
* **ESLint Version: 7.3.0**

The latest update breaks [eslint-plugin-import](https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/no-cycle.md) and I believe some other eslint plugins due to configuration cloning by `JSON.parse(JSON.stringify(Object))`
related PR: https://github.com/eslint/eslint/pull/13034

Any plugin that use any global property like "Infinity" wont work properly.

`JSON.parse(JSON.stringify(Infinity)) => null`

Configuration example https://github.com/airbnb/javascript/blob/master/packages/eslint-config-airbnb-base/rules/imports.js#L236

nzakas:
#### Prerequisites checklist

[ ] Documentation update
[ ] Add autofixing to a rule
[ ] Add a CLI option
[ ] Add something to the core
[ ] Other, please explain:

When cloning a rule's config, if `Infinity` is found, it will be replaced with `Number.MAX_SAFE_INTEGER`. This should result in the same effective config while preserving serialization.

Note: This only fixes the issue introduced in v7.3.0. Configs using `Infinity` will still fail in other areas (like caching and `--print-config`).

We can remove this change once `eslint-config-airbnb` updates their config.

### DIFF
--- a/lib/cli-engine/config-array-factory.js
+++ b/lib/cli-engine/config-array-factory.js
@@ -722,7 +722,12 @@ class ConfigArrayFactory {
          *
          * Refer https://github.com/eslint/eslint/issues/12592
          */
-        const clonedRulesConfig = rules && JSON.parse(JSON.stringify((rules)));
+        const clonedRulesConfig = rules && JSON.parse(
+            JSON.stringify(
+                rules,
+                (key, value) => (value === Infinity ? Number.MAX_SAFE_INTEGER : value)
+            )
+        );
 
         // Flatten `extends`.
         for (const extendName of extendList.filter(Boolean)) {
--- a/tests/fixtures/config-file/cloned-config/configWithInfinity.js
+++ b/tests/fixtures/config-file/cloned-config/configWithInfinity.js
@@ -1,0 +1,6 @@
+module.exports = {
+
+  rules: {
+      "max-len": [ "error", { code: Infinity }]
+  }
+};
--- a/tests/lib/cli.js
+++ b/tests/lib/cli.js
@@ -1172,9 +1172,7 @@ describe("cli", () => {
 
                 assert.strictEqual(exit, 0);
             });
-        });
 
-        describe("config file and input file", () => {
             it("should exit with 1 as camelcase has wrong property type", async () => {
                 const configPath = getFixturePath("config-file", "cloned-config", "eslintConfigFail.js");
                 const filePath = getFixturePath("config-file", "cloned-config", "index.js");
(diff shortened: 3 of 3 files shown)


lib/cli-engine/config-array-factory.js
```diff
@@ -722,7 +722,12 @@ class ConfigArrayFactory {
          *
          * Refer https://github.com/eslint/eslint/issues/12592
          */
-        const clonedRulesConfig = rules && JSON.parse(JSON.stringify((rules)));
+        const clonedRulesConfig = rules && JSON.parse(
+            JSON.stringify(
+                rules,
+                (key, value) => (value === Infinity ? Number.MAX_SAFE_INTEGER : value)
```
ljharb, line 728:
```suggestion
                (key, value) => (value === Infinity ? Number.MAX_SAFE_INTEGER : (value === -Infinity : -Number.MAX_SAFE_INTEGER : value))
```
just in case?

nzakas: I'd prefer not to add any further exceptions here. This fix is to carve out a narrow hole for `eslint-config-airbnb` to fit through without allowing other exceptions through.

ljharb: I suppose we can just wait for bug reports from enterprise users, which upgrade eslint much more slowly than open source users. Either way JSON-cloning is a breaking change, so if you'd prefer to gamble that it won't come up, that's fine ¯\\\_(ツ)_/¯

eaviles: Has anybody considered using something like https://github.com/codeandcats/fast-clone instead of JSON cloning?

eaviles: (actually: https://github.com/davidmarkclements/rfdc)

kaicataldo: @eaviles See https://github.com/eslint/eslint/pull/13034 for the original (long 😆) discussion.

eaviles: Oh wow… thanks for pointing me to this @kaicataldo

nzakas: Just as a reminder (and for posterity): ESLint has never formally supported JS special values in rule configs, which originated as JSON configs and are validated with JSON Schema. A validation bug let people use Infinity and we only learned of that after implementing this change. We don’t want to open the door to more people using special values in the meantime, which is why we are not reverting the original change.

If other bugs are filed, we will handle them on a case-by-case basis. Keep in mind that a large number of users use JSON or YAML config files, and this issue only affects JS config files.

SSANSH:
tags 7.3.1 with following airbnb fix is not available on npm registry is it normal?

PR #13471, opened 2020-07-07
add 'inlineConfigGroup' option

tjenkinson:
#### Prerequisites checklist

- [ ] Documentation update
- [x] Add a CLI option
- [x] Add something to the core
- [ ] Other, please explain:

Add support for a `inlineConfigGroup` option and directive comments that include a group name.

E.g if `inlineConfigGroup` was `build` then `/* eslint[build]-disable */` would be respected and `/* eslint-disable */` would be ignored.

When a group is omitted it defaults to `default`.

**Note this is not finished yet. I haven't checked it works or added tests, but I wanted to get some feedback before continuing**

#### Why?
We have an eslint config for our source code but we have a separate config that we use to run over build output to make sure it doesn't accidentally contain any features that devices don't support.

The problem is that the built version of our app still includes comments, including eslint directives. For the lint that runs over the build we need to ignore those comments. We would also like to include some directive comments in the src specifically for the build linter to ignore false alarms.

E.g. we have a lint rule that prevents es6 methods and one looks for `<something>.find`. This obviously has false positives and for those cases we'd like to be able to do

```js
// eslint[build]-disable-next-line
CoreUtils.find(...)
```

#### TODO
- [ ] check this makes sense
- [ ] tests
- [ ] docs

### DIFF
--- a/conf/config-schema.js
+++ b/conf/config-schema.js
@@ -22,6 +22,7 @@ const baseConfigProperties = {
     rules: { type: "object" },
     settings: { type: "object" },
     noInlineConfig: { type: "boolean" },
+    inlineConfigGroup: { type: "string" },
     reportUnusedDisableDirectives: { type: "boolean" },
 
     ecmaFeatures: { type: "object" } // deprecated; logs a warning when used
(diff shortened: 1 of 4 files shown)


nzakas:
Closing due to age

PR #13543, opened 2020-08-02
Breaking: Reduce carbon footprint by calling ESLint only on first CI job

ai:
#### Prerequisites checklist

- [ ] Documentation update
- [ ] Add a CLI option
- [x] Add something to the core
- [ ] Other, please explain:

Cal ESLint only on first CI job to reduce unnecessary servers work (e.g., reduce carbon footprint by unnecessary work).

I added [`ci-job-number`](https://github.com/ai/ci-job-number/). It returns the job numbers and supports many different CI services. For the second and rest jobs, I print warning (with a note of how to disable this behavior) and exit.

#### The context behind these changes

@pwithnall in [his recent talk](https://events.gnome.org/event/1/contributions/79/attachments/12/30/presentation.pdf) suggested thinking more about how our open source projects affect the environment. I found an interesting problem with ESLint and how people are using ESLint on CI.

```yaml
node_js:
  - node
  - "12"
  - "10"
  - "8"
  - "6"
```

Here is a [Travis CI of my PostCSS](https://travis-ci.org/github/postcss/postcss/builds/691150914). Like many other projects, we many jobs to run tests on all actual Node.js versions. In every job we call `npm test`:

```js
    "test": "eslint . && jest"
```

As a result, we call ESLint for every Node.js version. But since ESLint provides static analysis, ESLint output will be the same for all Node.js versions. As a result, we have 2-4 unnecessary ESLint calls for all commits for all open source projects with multiple Node.js versions in `.travis.yml`.

##### Is it a real problem?

* For `postcss/postcss` every ESLint takes *9 seconds* of CPU time.
* For `postcss/postcss` we have around *250 commits* per year.
* Let’s count that we have *4 Node.js* versions on Travis CI (we had five versions before and now only 3).
* 1 hour of CPU work gives around 1.7 grams of CO2.
* **9000 seconds** of unnecessary CPU time = **42.5 grams of CO2 per year** just for one open-source project.
* ESLint is used in >3 000 000 projects. Let’s assume that 75% of them are dead (of course, it is not fair to use PostCSS’s CPU/jobs/time statistics, so I took a bigger number for dead projects and didn’t count non-GitHub projects).
* We can expect that adding this commit could prevent **31 875 Kg of CO2 per year**. *But please double-check my math or even find a better statistics.*

##### Does somebody already do like this?

* There is [`eslint-ci`](https://github.com/JLHwung/eslint-ci) by @JLHwung, which already adds `ci-job-number` to ESLint. I use `eslint-ci` in all my open-source projects (including PostCSS, Autoprefoxer, and Browserslist). There are also [>200 projects](https://github.com/search?l=JSON&q=%22eslint-ci%22&type=Code), which use `eslint-ci` too.
* I use `ci-job-number` in my [`size-limit`](https://github.com/ai/size-limit/) and [`check-dts`](https://github.com/ai/check-dts).
* [`spech`](https://www.npmjs.com/package/spech), spelling check tool, also uses `ci-job-number`.

##### Why do we need this commit if we have `eslint-ci`?

People think about Travis CI as a free resource (which is not true). It is hard to convince people to avoid spending free resources if they think resources are free for them personally.

##### GitHub Actions can be tricky, and jobs can be used not only for Node.js version.

Yeap, [found](https://github.com/ai/size-limit/issues/165) that GitHub Actions workflows can be very tricky, and jobs can be used not only to test different Node.js versions.

This is why we removed GitHub Actions support from `ci-job-number`. On GitHub Actions, ESLint will run on every job.

We are collecting feedback and evolving `ci-job-number` to have predicted results of every CI service.

### DIFF
--- a/bin/eslint.js
+++ b/bin/eslint.js
@@ -138,6 +138,15 @@ ${message}`);
         return;
     }
 
+    if (require("ci-job-number")() > 1) {
+        console.warn(
+            "To speed up CI ESLint runs only in first job.\n" +
+            "Set CI_JOB_NUMBER=1 environment variable if you need " +
+            "to call ESLint for all job."
+        );
+        return;
+    }
+
     // Otherwise, call the CLI.
     process.exitCode = await require("../lib/cli").execute(
         process.argv,
--- a/bin/eslint.js
+++ b/bin/eslint.js
@@ -142,7 +142,7 @@ ${message}`);
         console.warn(
             "To speed up CI ESLint runs only in first job.\n" +
             "Set CI_JOB_NUMBER=1 environment variable if you need " +
-            "to call ESLint for all job."
+            "to execute ESLint for all jobs."
         );
         return;
     }
--- a/package.json
+++ b/package.json
@@ -49,6 +49,7 @@
     "@babel/code-frame": "^7.0.0",
     "ajv": "^6.10.0",
     "chalk": "^4.0.0",
+    "ci-job-number": "^1.2.2",
     "cross-spawn": "^7.0.2",
     "debug": "^4.0.1",
     "doctrine": "^3.0.0",


package.json
```diff
@@ -49,6 +49,7 @@
     "@babel/code-frame": "^7.0.0",
     "ajv": "^6.10.0",
     "chalk": "^4.0.0",
+    "ci-job-number": "^1.2.2",
```
lifeart, line 52: why we need this package? there is only few lines of code
`process.env.CI_JOB_NUMBER ? parseInt(process.env.CI_JOB_NUMBER, 10) : 0`

lifeart: ah, as I can see there is few different ENV checks https://github.com/ai/ci-job-number/blob/master/index.js#L17

Will it be better to have explicit CI argument, instead of overriding CI_JOB_NAME to get eslint applied to every run?

ai: Sure, I can rename `CI_JOB_NUMBER`.

What name do you suggest?

lifeart: ‘ESLINT_LINT_ONCE’ ?

lifeart: Anyway, linting based on job index looks quite dangerous

ai: What do you think about `DISABLE_CI_JOB_IGNORING`?

We need a environment variable not just for ESLint but for the library, to reuse library in other heavy tasks.

For instance, in PostCSS we are calling spelling check only on the first CI job.

kaicataldo:
While I agree that this a worthy cause, I don’t think this is the responsibility of ESLint and as a result is not the correct solution to this problem. 

The responsibility of how many times a tool like ESLint is called should be the responsibility of the task runner. Some solutions to this:

1. CI can be set up to only run linting once (example [here](https://github.com/babel/babel/blob/a4ebe29b3f983dadabe13fdbd3e32c54a030eae4/.travis.yml#L28)).
1. Write a script that is aware of its context and will run ESLint conditionally.

ai:
> CI can be set up to only run linting once (example here).

We already have an `eslint-ci`, but seems like there is no way how to do it popular because nobody cares about “free resources”.

saibotsivad:
Very interesting proposal, and I definitely like the idea of default settings being thoughtful in ways like this.

With that said, I tried re-doing your math and I think you're off by a factor in one of your calculations:

* Given: 1 hour of CPU work gives around 1.7 grams of CO2
* Given: `postcss/postcss` every ESLint takes 9 seconds of CPU time
* Given: `postcss/postcss` has around 250 commits per year
* Estimated: has 4 Node.js versions on Travis CI

With this ESLint change you'd still have at least one execution, so you'd go from 4 executions to 3.

"Waste" for `postcss/postcss`:

* Executions per year: (3 executions per commit) * (250 commits per year) = (750 executions per year)
* CPU seconds per year: (750 executions per year) * (9 seconds per execution) = (6750 seconds per year)
* CPU hours per year: (6750 seconds per year) * (1/3600 hours per second) = (1.875 hours per year)
* CO2 per year: (1.875 hours per year) * (1.7 grams per hour) = (3.1875 grams per year)

Attempting to extend estimates to all projects using ESLint.

* Estimated: 25% of projects are similar enough to `postcss/postcss`
* Estimated: remaining projects are not active
* Given: ESLint used in 3,000,000 projects

"Waste" for ESLint overall:

* (3,000,000 projects overall) * (0.25 active per overall) = (750,000 comparable projects)
* (750,000 projects) * (3.1875 grams per year per project) = (2,390 kg per year)

Estimated "waste" prevented by this change:

* 2390 kg per year
* Or, 5269 lb per year

Although that's significantly smaller than the number you calculated, **it is still very sizable** and worth consideration as a default option for ESLint, especially if it's made clear in documentation and output results.

kaicataldo:
> We already have an `eslint-ci`, but seems like there is no way how to do it popular because nobody cares about “free resources”.

Can you explain what you mean by this? Not sure I follow.

ai:
> Can you explain what you mean by this? Not sure I follow.

Nothing will change if we will suggest a better CI config in the docs. If we had no significant feedback to `eslint-ci` we will not bring attention to better CI config suggestion in docs.

Until Travis CI is “free” users will use the simplest config with `"test": "eslint . && jest"`.

Think about is as plastic bags in supermarkets. Adding banner with “bring your bag” do not work.

JLHwung:
> > We already have an `eslint-ci`, but seems like there is no way how to do it popular because nobody cares about “free resources”.
> 
> 
> 
> Can you explain what you mean by this? Not sure I follow.

eslint-ci is basically an one-liner
```
If(it is the first CI job) run eslint
```

However we can not replace eslint by eslint-ci unconditionally because it has an implicit assumption that eslint is run in every CI jobs. This may be true if the CI is a matrix of different node versions. However, for projects like Babel, ESLint is run only on the first job so eslint-ci is not applicable, or even worse: it may incorrectly skip eslint if eslint is scheduled as the second CI job.

ai:
Yeap. I am suggesting these changes only as part of major release.

Do you have an example of Travis CI project where these changes will not work?

nzakas:
While I appreciate the thought behind this, I don't see the benefit of adding this change directly into ESLint. We don't want to be in the business of telling people how they should use ESLint in their CI. Everyone's needs are different, and ESLint shouldn't bake-in any assumptions about those needs.

pwithnall:
> We don't want to be in the business of telling people how they should use ESLint in their CI.

(A drive-by comment since I was CCed in:)

I think you probably do. Provide a sensible, efficient default configuration which is trivial for people to use, and they’ll be nudged into using it. If they need something more advanced, provide options to give the flexibility to those who need it. But keep the defaults efficient and easy for people to use, since I bet a large proportion of people setting ESLint up in CI want to do it as fast as possible and move on with their lives.

nzakas:
Thanks for the discussion everyone. I’m glad people are looking out for resource utilization.

ESLint is used in a number of contexts: in editors and IDEs, precommit hooks, manually on the command line, and CI systems. We can’t make any assumptions about the execution context, nor do we want to try to guess the context based on environment variables.

I think the best option for the ESLint project would be to add some documentation on how to set up a Travis job and the options for doing so efficiently.

bin/eslint.js
```diff
@@ -138,6 +138,15 @@ ${message}`);
         return;
     }
 
+    if (require("ci-job-number")() > 1) {
+        console.warn(
+            "To speed up CI ESLint runs only in first job.\n" +
+            "Set CI_JOB_NUMBER=1 environment variable if you need " +
+            "to call ESLint for all job."
```
php-coder, line 145:
```suggestion
            "to call ESLint for all jobs."
```
?

nzakas:
Thanks everyone, we're not moving forward with this pull request. If anyone would like to submit a tutorial on setting up ESLint in a CI environment (as [discussed previously](https://github.com/eslint/eslint/pull/13543#issuecomment-669618628)) please feel free.

PR #13626, opened 2020-08-28
Chore: Test formatted integers in no-dupe-keys (refs #13568)

btmills:
#### Prerequisites checklist

[x] Other, please explain:

Add additional rule tests for numeric separators as part of #13568.

This adds tests for numeric literal object keys that aren't digit-only decimal integer literals, including ES6 binary and octal literals, ES2020 bigint literals, and ES2021 numeric separators. The valid tests have similar-looking literals that are distinct keys, and the invalid tests have differently-formatted literals that result in duplicate keys.

This is in addition to #13574 and #13581. Are there any other rules that might be affected?


mdjermanovic (approved):
LGTM, thanks! Makes sense to verify this rule.

> Are there any other rules that might be affected?

I'd guess all rules where the value of the literal is important would be affected if we were using something else instead of `Literal#value`, like `Number(node.raw)` or `+token.value`. 

It's probably a fairly long list to double-check, and it might include:

* Most of the rules that work on properties of object literals/classes.
* Rules that compare two nodes, like `no-self-assign`.
* Misc rules, like `no-magic-numbers` (with `"ignore"`), `yoda` (with `"exceptRange"`) and maybe some more.

kaicataldo:
LGTM, thanks!

PR #13661, opened 2020-09-06
Docs: Provide install commands for Yarn

nbaksalyar:
npx doesn't work properly with plugins installed with Yarn.

#### Prerequisites checklist

- [x] Documentation update

The installation guide might be slightly confusing for new users - e.g.: I use TypeScript in my project, I install `eslint` with yarn as the guide says, and then run it with `npx eslint --init`. It then fails with an error because the TypeScript plugin has been installed with yarn and `npx` doesn't see it:
```
Error: Failed to load plugin '@typescript-eslint' declared in '.eslintrc.js': Cannot find module 'eslint'
```
This can be fixed by using `yarn run <cmd>` instead of `npx <cmd>`. This PR changes documentation to reflect this.

### DIFF
--- a/docs/user-guide/getting-started.md
+++ b/docs/user-guide/getting-started.md
@@ -24,12 +24,20 @@ You should then set up a configuration file:
 
 ```
 $ npx eslint --init
+
+# or
+
+$ yarn run eslint --init
 ```
 
 After that, you can run ESLint on any file or directory like this:
 
 ```
 $ npx eslint yourfile.js
+
+# or
+
+$ yarn run eslint yourfile.js
 ```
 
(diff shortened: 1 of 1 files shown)


btmills (approved):
Thanks for contributing, @nbaksalyar!

> npx doesn't work properly with plugins installed with Yarn.

I didn't know that, but that's annoying. Is that because of the installation changes in Yarn v2? (I tried it with Yarn v1 and didn't have any issues.) Either way, LGTM because I'm fine with having both npm/npx and yarn commands for consistency.

nbaksalyar:
> Is that because of the installation changes in Yarn v2?

I think so! I use the `berry` (v2.2) version for Yarn.

Thanks for reviewing :)

anikethsaha (approved):

kaicataldo (approved):

kaicataldo:
Thanks for contributing!

PR #13665, opened 2020-09-07
Docs: Update quotes.md

sunny775:
Docs: Changes to the quotes.md section of the documentation
The section descriptions do not match the code samples, hence  I'm proposing these changes

#### Prerequisites checklist

[ ] Documentation update
[ ] Add autofixing to a rule
[ ] Add a CLI option
[ ] Add something to the core
[ ] Other, please explain:

### DIFF
--- a/docs/rules/quotes.md
+++ b/docs/rules/quotes.md
@@ -108,7 +108,7 @@ Examples of additional **correct** code for this rule with the `"double", { "avo
 ```js
 /*eslint quotes: ["error", "double", { "avoidEscape": true }]*/
 
-var single = 'a string containing "double" quotes';
+var double = "a string containing 'single' quotes";
 ```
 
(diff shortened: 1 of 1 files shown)


btmills:
:wave: hi @sunny775, thanks for contributing!

The documentation here is correct as-is. It's demonstrating use of the `avoidEscape` option, which allows using a non-preferred quote style in order to avoid escaping quotes within the string. For example, if I prefer double quotes but have a string that itself needs to contain double quotes as in the first example, normally I'd need to escape them as in `var escaped = "a string containing \"double\" quotes";`. With the `allowEscape` option, even though I prefer double quotes, I'd rather use single quotes than have to escape the string, so I write it as `var single = 'a string containing "double" quotes';`.

sunny775:
Wow. 
That explanation is insightful.
I understand now.
Thanks very much

PR #13787, opened 2020-10-24
Update: disallow if(cond); (fixes #13785)

### LINKED ISSUE #13785: no-extra-semi should prevent `if (condition);`

no-extra-semi

more

I suspect new default behaviour would be more fitting because I can't imagine a good use of `if (cond);` but I'm open to being corrected. A new option would be an alternative.

```js
let a = 0;
if (a === 0); 
```

Ignores the code, produces no warnings. It seems to be by design.

Produce a warning.

Yes but I'm very busy unfortunately so if someone does that first, I won't mind at all. I'd need guidance on whether it should be default behaviour or a new option. Default behaviour change seems easier to implement in this case.

jacekkopecky:
#### Prerequisites checklist

[ ] Documentation update
[ ] Add autofixing to a rule
[ ] Add a CLI option
[ ] Add something to the core
[ ] Other, please explain:

Fixes #13785

- made no-extra-semi report a suggestion to remove a semicolon right after if
- Fixed tests to expect the code to newly report errors.
- added an example of this in the documentation.

- is the documentation enough? It doesn't mention the case of `if (cond) {} else ;` that should fail too, but then it doesn't mention that the rule intentionally allows `while (cond);`.
- do tests normally check the exact content of suggestions? Mine don't because I had no template for that.

### DIFF
--- a/docs/rules/no-extra-semi.md
+++ b/docs/rules/no-extra-semi.md
@@ -17,6 +17,8 @@ function foo() {
     // code
 };
 
+if (x === 5);
+
 ```
 
 Examples of **correct** code for this rule:
@@ -30,11 +32,14 @@ var foo = function() {
     // code
 };
 
+if (x === 5) {}
 ```
 
 ## When Not To Use It
 
 If you intentionally use extra semicolons then you can disable this rule.
+
+If you intentionally use an `if` without a body, please let us know.
 
 ## Related Rules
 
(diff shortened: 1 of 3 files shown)


jacekkopecky:
This is not ready yet, the discussion in #13785 suggests that a new rule, rather than an update to an existing one, could be preferable. Closing for now.

PR #14214, opened 2021-03-15
Chore: Require specific `semver` functions to avoid requiring the whole tree

fregante:
Semi-related to #14098 

This doesn't remove any dependencies, but it helps with startup times (a lil bit) since ESLint now doesn't have to load every single `semver` file.


fregante:
The bot is annoying and noisy. Opening a new PR

PR #14215, opened 2021-03-15
Chore: Require specific `semver` functions

fregante:
Semi-related to #14098 

This doesn't remove any dependencies, but it helps with startup times (a lil bit) since ESLint now doesn't have to load every single `semver` file.

### DIFF
--- a/Makefile.js
+++ b/Makefile.js
@@ -21,7 +21,10 @@ const checker = require("npm-license"),
     markdownlint = require("markdownlint"),
     os = require("os"),
     path = require("path"),
-    semver = require("semver"),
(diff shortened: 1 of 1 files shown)


nzakas:
This file is primarily used for releases so it doesn’t have any influence over ESLint run time. As such, I’d prefer to keep it as-is for clarity.

fregante:
Silly me, I meant to change https://github.com/eslint/eslint/blob/179a910b32e853bc12a9dd71f7c10e762cbeac44/lib/init/config-initializer.js instead of this

I'll send another PR if you don't mind.

PR #14230, opened 2021-03-18
Docs: Add valid example that shows vars in a block scope

edg2s:
#### Prerequisites checklist

[x] Documentation update
[ ] Add autofixing to a rule
[ ] Add a CLI option
[ ] Add something to the core
[ ] Other, please explain:

All the current valid examples show hoisting the var to the function scope. This example makes is clear that this rule allows you to declare vars inside a block scope, just not use them out of that scope.

### DIFF
--- a/docs/rules/block-scoped-var.md
+++ b/docs/rules/block-scoped-var.md
@@ -71,6 +71,13 @@ function doTryCatch() {
         f = build;
     }
 }
+
+function doFor() {
+    for (var x = 1; x < 10; x++) {
+        var y = f(x);
+        console.log(y);
+    }
+}
 ```
 
 ## Further Reading
--- a/docs/rules/block-scoped-var.md
+++ b/docs/rules/block-scoped-var.md
@@ -34,6 +34,13 @@ function doTryCatch() {
         var f = build;
     }
 }
+
+function doFor() {
+    for (var x = 1; x < 10; x++) {
+        var y = f(x);
+    }
+    console.log(y);
+}
 ```
 
 Examples of **correct** code for this rule:


nzakas (changes requested):
Thanks. Can you add a corresponding example in the “invalid” section, too? Note that all of the current examples in the “valid” section are rewritten versions of the code in the “invalid” section, and that is by design.

edg2s:
I did notice that, but I don't think there is a functionally equivalent invalid example for this function. I could move the console.log down but that would be different functionality.

nzakas:
You may want to consider a different example, then, that can show the same thing.

edg2s:
> You may want to consider a different example, then, that can show the same thing.

As I said above, I have considered this. Any invalid case would have a variable being used outside the scope in which it was defined. In order to fix that _without_ changing the functionality you would have to move the variable declaration further up, but the whole point of this PR is to add an example of a var in a block scope.

edg2s:
> Note that all of the current examples in the “valid” section are rewritten versions of the code in the “invalid” section, and that is by design.

And there are many other places where the examples don't (or can't) match up perfectly, e.g. https://eslint.org/docs/rules/no-compare-neg-zero, https://eslint.org/docs/rules/valid-typeof, https://eslint.org/docs/rules/dot-location, ...

nzakas:
The goal isn’t necessarily to show a bad example and then the fixed version, it’s to show an example of what the rule will flag and then a similar example that shows what the rule won’t flag.

Despite other rules not always following this pattern, this rule’s documentation does, so I’d prefer to stick with it.

nzakas (approved):
LGTM. Thanks!

PR #14243, opened 2021-03-23
Chore: "yorkie" -> "simple-git-hooks"

g-plane:
#### Prerequisites checklist

Other, please explain: Infrastructure change.

Migrated the Git hooks manager from [yorkie](https://github.com/yyx990803/yorkie) to [simple-git-hooks](https://github.com/toplenboren/simple-git-hooks).

`simple-git-hooks` is now used in some famous projects like PostCSS, Autoprefixer and so on. ([See this](https://github.com/toplenboren/simple-git-hooks#who-uses-simple-git-hooks))

Andrey Sitnik has stated about `simple-git-hooks` at his pull requests. ([source](https://github.com/sapegin/mrm/pull/149)) Note that `yorkie` is forked from `husky` v4, so we can treat `yorkie` as `husky` v4.

I'm not sure.

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,8 +23,8 @@
     "webpack": "node Makefile.js webpack",
     "perf": "node Makefile.js perf"
   },
-  "gitHooks": {
-    "pre-commit": "lint-staged"
+  "simple-git-hooks": {
+    "pre-commit": "npx lint-staged"
   },
   "lint-staged": {
     "*.js": [
@@ -128,11 +128,11 @@
     "recast": "^0.19.0",
     "regenerator-runtime": "^0.13.2",
     "shelljs": "^0.8.2",
+    "simple-git-hooks": "^2.0.3",
     "sinon": "^9.0.1",
     "temp": "^0.9.0",
     "webpack": "^5.23.0",
-    "webpack-cli": "^4.5.0",
-    "yorkie": "^2.0.0"
+    "webpack-cli": "^4.5.0"
   },
   "keywords": [
     "ast",


nzakas:
Can you explain more about this change? Is there an issue with Yorkie? Does simple-git-hooks provide some benefit?

Note that Yorkie is used by Vue, so it is also fairly well-tested.

g-plane:
- Yorkie is forked but no longer updated.
- simple-git-hooks has no dependencies and smaller install size, faster.

nzakas (approved):
LGTM

nzakas:
Waiting for others to review.

noto-u:
I prefer yorkie for the automatic update and well tested, if ain't broke don't fix philosophy. Simple-git-hooks I noticed that coworkers have to run a command to update configuration when it's changed in master branch.

> If you need multiple verbose commands per git hook, flexible configuration or automatic update of git hooks, please check out the other packages:
> * Lefthook
> * husky
> * pre-commit

nzakas:
@noto-u thanks for the feedback. I didn’t realize that was an issue with simple-git-hooks (mentioned here: https://github.com/toplenboren/simple-git-hooks#when-to-use-it). In that case, I think we should stick with Yorkie.

g-plane:
It shouldn't be a big problem. If we need to update automatically, we just need to add `install` or `postinstall` script to `package.json` file.

nzakas:
I was already skeptical about this change, and now I’m not comfortable with it. We aren’t having any problems with Yorkie, so let’s just keep using it.

PR #14287, opened 2021-04-02
Chore: Remove lodash

stephenwade:
#### Prerequisites checklist

[ ] Documentation update
[ ] Add autofixing to a rule
[ ] Add a CLI option
[ ] Add something to the core
[x] Other, please explain: _Reduce dependencies_

Addresses part of #14098. Removes lodash entirely from the project, replacing all uses with native code, custom code, and a handful of other tiny modules.

<details>
<summary>node_modules size comparison</summary>

```diff
~/git/eslint-test-master$ echo '{ "dependencies": { "eslint": "eslint/eslint" } }' > package.json
~/git/eslint-test-master$ npm install >/dev/null
~/git/eslint-test-master$ du -sh node_modules
22M	node_modules

~/git/eslint-test-stephenwade$ echo '{ "dependencies": { "eslint": "stephenwade/eslint#remove-lodash-3" } }' > package.json
~/git/eslint-test-stephenwade$ npm install >/dev/null
~/git/eslint-test-stephenwade$ du -sh node_modules
17M	node_modules

~/git$ diff -U0 <(cd eslint-test-master; du -hd1 node_modules) <(cd eslint-test-stephenwade; du -hd1 node_modules) | grep "^[-+]\d"
-4.9M	node_modules/lodash
-16K	node_modules/escape-string-regexp
+20K	node_modules/escape-string-regexp
+28K	node_modules/deep-extend
-240K	node_modules/@babel
+256K	node_modules/@babel
-22M	node_modules
+17M	node_modules
```
</details>

There is one fairly major change in here. I replaced lodash.template by creating a JS file in place of each template. Let me know if I should take a different approach with that or if I should split it out into a separate PR.


stephenwade:
I'll take a look at the CI failure tomorrow. Looks like I missed a .flat call

bcomnes (approved):

package.json
```diff
@@ -50,16 +50,21 @@
     "@eslint/eslintrc": "^0.4.0",
     "ajv": "^6.10.0",
     "chalk": "^4.0.0",
+    "clone": "^2.1.2",
```
g-plane, line 53: Recommend: [rfdc](https://github.com/davidmarkclements/rfdc). `clone` hasn't been updated about three years.

tools/eslint-fuzzer.js
```diff
@@ -10,13 +10,26 @@
 //------------------------------------------------------------------------------
 
 const assert = require("assert");
-const lodash = require("lodash");
+const mapValues = require("map-values");
```
fregante, line 13: I don't think this is a useful dependency. Now it can be replicated with:

```js
Object.fromPairs(Object.entries(obj).map(([value, key]) => cb(value, key, obj))))
```

Or likely something easier to read and more specific to the situation

package.json
```diff
@@ -71,17 +75,20 @@
...
+    "omit": "^1.0.1",
     "optionator": "^0.9.1",
     "progress": "^2.0.0",
     "regexpp": "^3.1.0",
+    "rfdc": "^1.3.0",
     "semver": "^7.2.1",
     "strip-ansi": "^6.0.0",
     "strip-json-comments": "^3.1.0",
-    "table": "^6.0.4",
+    "table": "^6.0.9",
```
fregante, line 89: This change should be dropped since `table` will be dropped by https://github.com/eslint/eslint/pull/14316

tests/lib/rules/no-invalid-this.js
```diff
@@ -79,7 +80,7 @@ function extractPatterns(patterns, type) {
             thisPattern.code += " /* should error */";
         }
 
-        return lodash.omit(thisPattern, ["valid", "invalid"]);
+        return omit(["valid", "invalid"], thisPattern);
```
fregante, line 83:
```suggestion
        delete thisPattern.valid;
        delete thisPattern.invalid;
        return thisPattern;
```

lib/rule-tester/rule-tester.js
```diff
@@ -477,13 +481,13 @@ class RuleTester {
                  * Assumes everything on the item is a config except for the
                  * parameters used by this tester
                  */
-                const itemConfig = lodash.omit(item, RuleTesterParameters);
+                const itemConfig = omit(RuleTesterParameters, item);
```
fregante, line 484: Not as straightforward, but lets us skip a barely-used dependency
```suggestion
                const itemConfig = {...item};
                for (const parameter of RuleTesterParameters) {
                	delete[parameter];
                }
```

package.json
```diff
@@ -71,17 +75,20 @@
...
     "levn": "^0.4.1",
-    "lodash": "^4.17.21",
+    "map-values": "^1.0.1",
     "minimatch": "^3.0.4",
     "natural-compare": "^1.4.0",
+    "omit": "^1.0.1",
     "optionator": "^0.9.1",
     "progress": "^2.0.0",
     "regexpp": "^3.1.0",
+    "rfdc": "^1.3.0",
```
fregante, line 85: I think `deepExtend({}, obj)` is equivalent to `clone(obj)`. Can this dependency be dropped too?

package.json
mdjermanovic: It's fine to have this change included since this PR can be merged for a minor or patch release, while https://github.com/eslint/eslint/pull/14316 has to wait for the next major.

tools/eslint-fuzzer.js
stephenwade: I would prefer to use the dependency. I think

```js
mapValues(ruleConfigs, sample)
```

is way more readable than

```js
Object.fromEntries(Object.entries(ruleConfigs).map(([value, key]) => [value, sample(key)]))
```

In my opinion, the addition of a tiny (16K) module is worth the readability here.

fregante: Why use a “tiny 16k dependency” when you can use a 1-liner in some other file? It makes zero sense for a PR that attempts to reduce dependencies. 

Like I said, that’s the exact equivalent but you can find better solutions for the 2 places where it’s used.

stephenwade: I hadn't thought of creating a function. That's a good idea. I'll do that and remove the dependency.

stephenwade: Object.fromEntries is not available until Node 12. I have added a TODO comment to replace this dependency with a helper function when we drop support for Node 10.

Resolved in a28cf2db1.

lib/rules/comma-dangle.js
```diff
@@ -144,23 +143,33 @@ module.exports = {
...
+
+            /**
+             * Returns the last element of an array
+             * @param {any[]} array The input array
+             * @returns {any} The last element
+             */
+            function last(array) {
+                return array[array.length - 1];
+            }
+
```
aladdin-add, line 155: there is a stage-3 proposal: https://github.com/tc39/proposal-relative-indexing-method
so it will be written to `array.at(-1)` someday. 😄

tests/lib/rules/no-invalid-this.js
stephenwade: Resolved in ada00d52f.

lib/rule-tester/rule-tester.js
stephenwade: Resolved in ada00d52f.

lib/linter/apply-disable-directives.js
```diff
@@ -124,7 +122,21 @@ module.exports = ({ directives, problems, reportUnusedDisableDirectives = "off"
         .map(directive => Object.assign({}, directive, { unprocessedDirective: directive }))
         .sort(compareLocations);
 
-    const lineDirectives = lodash.flatMap(directives, directive => {
+    /**
+     * Returns a new array formed by applying a given callback function to each element of the array, and then flattening the result by one level.
+     * TODO(stephenwade): Replace this with array.flatMap when we drop support for Node v10
```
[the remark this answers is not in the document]
stephenwade: I am hoping this PR can be merged before the next major release. If Node 10 support is dropped before this is merged, I will update this PR at that time.

package.json
stephenwade: Resolved in f9ac034a2.

package.json
```diff
@@ -71,17 +75,20 @@
...
     "semver": "^7.2.1",
     "strip-ansi": "^6.0.0",
     "strip-json-comments": "^3.1.0",
-    "table": "^6.0.4",
+    "table": "^6.0.9",
     "text-table": "^0.2.0",
+    "upper-case-first": "^2.0.2",
```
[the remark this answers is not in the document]
stephenwade: Resolved this in 7dff2566a by adding `lib/shared/string-utils`.

aladdin-add (approved):
LGTM, thanks!

mdjermanovic (changes requested):
@stephenwade thanks for all this work! I left a couple of suggestions, didn't manage to review all details yet.

lib/shared/deprecation-warnings.js
```diff
@@ -36,7 +36,7 @@ const deprecationWarningMessages = {
  * @param {string} errorCode The warning message to show.
  * @returns {void}
  */
-const emitDeprecationWarning = lodash.memoize((source, errorCode) => {
+const emitDeprecationWarning = memoize((source, errorCode) => {
```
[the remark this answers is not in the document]
stephenwade: Thanks for catching that! I've removed the fast-memoize dependency in 6232136f3.

tests/lib/shared/string-utils.js
```diff
@@ -0,0 +1,39 @@
...
+
+    it("does not change the case if the first letter is already uppercase", () => {
+        assert(upperCaseFirst("Alphabet") === "Alphabet");
+        assert(upperCaseFirst("One Two Three") === "One Two Three");
+    });
+
+    it("properly handles an empty string", () => {
+        assert(upperCaseFirst("") === "");
+    });
+});
```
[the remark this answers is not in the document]
stephenwade: Done in e46f4716c.

lib/rules/object-curly-newline.js
```diff
@@ -69,6 +68,24 @@ function normalizeOptionValue(value) {
...
 }
 
+/**
+ * Checks if an object is a plain object
+ * @param {any} object The object to check
+ * @returns {boolean} `true` if the option is a plain object, otherwise `false`
+ */
+function isPlainObject(object) {
+    return Object.prototype.toString.call(object) === "[object Object]";
+}
```
mdjermanovic, line 78: In the context of this rule, we just need to check if the given value is an object. Can we do something like this, to not rely on `"[object Object]"`:

```js
/**
 * Check if a value is an object.
 * @param {any} value The value to check
 * @returns {boolean} `true` if the value is an object, otherwise `false`
 */
function isObject(value) {
    return typeof value === "object" && value !== null;
}
```

lib/linter/node-event-generator.js
```diff
@@ -166,23 +194,32 @@ function tryParseSelector(rawSelector) {
     }
 }
 
+const selectorCache = {};
```
mdjermanovic, line 197: Can we use `Map` instead of object here? Selectors can be user-defined (e.g., in `no-restricted-syntax` rule), so it's better to avoid usual issues with object-maps.

mdjermanovic (changes requested):

lib/rules/indent.js
```diff
@@ -1560,7 +1566,7 @@ module.exports = {
          * 2. Don't set any offsets against the first token of the node.
          * 3. Call `ignoreNode` on the node sometime after exiting it and before validating offsets.
          */
-        const offsetListeners = lodash.mapValues(
+        const offsetListeners = mapValues(
```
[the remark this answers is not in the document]
stephenwade: Resolved in c20cce3eb.

lib/rules/object-curly-newline.js
stephenwade: Resolved in a1bd47baa.

lib/linter/node-event-generator.js
stephenwade: Resolved in a1bd47baa.

package.json
```diff
@@ -52,14 +52,17 @@
     "chalk": "^4.0.0",
     "cross-spawn": "^7.0.2",
     "debug": "^4.0.1",
+    "deep-extend": "^0.5.1",
```
mdjermanovic, line 55: There are some differences between `deep-extend` and `lodash.merge`/`lodash.cloneDeep`:

```js
const lodash = require("lodash");
const extend = require("deep-extend");

const test1 = [{ foo: 5 }, { foo: undefined }];

console.log(lodash.merge({}, ...test1)); // { foo: 5 }
console.log(extend({}, ...test1)); // { foo: undefined }

const test2 = [{ foo: [1, 2, 3] }, { foo: [4] }];

console.log(lodash.merge({}, ...test2)); // { foo: [4, 2, 3] }
console.log(extend({}, ...test2)); // { foo: [4] }

const test3 = { foo: {} };
test3.foo.bar = test3; // circular

lodash.cloneDeep(test3); // ok
extend({}, test3); // RangeError: Maximum call stack size exceeded

```

We should estimate the impact of the differences, or use some other dependency.

The latest `0.6.0` works the same as `0.5.1` on these examples. Also, was there a specific reason to require `^0.5.1` instead of `^0.6.0`?

stephenwade: > Also, was there a specific reason to require `^0.5.1` instead of `^0.6.0`?

Weird, that's what happens when I run `npm i -s deep-extend`. I've bumped this up to `^0.6.0` in b3d6d9a7b.

stephenwade: #

stephenwade: I tested 8 different Node modules for deep merging. None of them had quite the same behavior as `lodash`.

<details>
<summary>Differences in behavior between various Node modules for deep merging</summary>

## `lodash.merge`

```jsx
const f = require("lodash").merge;

// test undefined values
f({}, { foo: 5 }, { foo: undefined });
// { foo: 5 }

// test arrays
f({}, { foo: [1, 2, 3] }, { foo: [4] });
// { foo: [4, 2, 3] }

// test circular referencces
const obj = { foo: {} };
obj.foo.bar = obj;
f({}, obj);
// { foo: <ref *1> { bar: { foo: [Circular *1] } } }
```

## `deep-extend`

```jsx
const f = require("deep-extend");

// test undefined values
f({}, { foo: 5 }, { foo: undefined });
// { foo: undefined }

// test arrays
f({}, { foo: [1, 2, 3] }, { foo: [4] });
// { foo: [ 4 ] }

// test circular referencces
const obj = { foo: {} };
obj.foo.bar = obj;
f({}, obj);
// Uncaught RangeError: Maximum call stack size exceeded
```

## `deepmerge`, `merge-deep`

```jsx
const deepmerge = require("deepmerge");
const f = (...params) => deepmerge.all(params);
// or
const f = require("merge-deep");

// test undefined values
f({}, { foo: 5 }, { foo: undefined });
// { foo: undefined }

// test arrays
f({}, { foo: [1, 2, 3] }, { foo: [4] });
// { foo: [ 1, 2, 3, 4 ] }

// test circular referencces
const obj = { foo: {} };
obj.foo.bar = obj;
f({}, obj);
// Uncaught RangeError: Maximum call stack size exceeded
```

## `@common-utilities/merge-objects`

```jsx
const { mergeObjects } = require("@common-utilities/merge-objects");
const f = (...params) => params.reduce(mergeObjects);

// test undefined values
f({}, { foo: 5 }, { foo: undefined });
// { foo: 5 }

// test arrays
f({}, { foo: [1, 2, 3] }, { foo: [4] });
// { foo: [ 1, 2, 3, 4 ] }

// test circular referencces
const obj = { foo: {} };
obj.foo.bar = obj;
f({}, obj);
// { foo: <ref *1> { bar: { foo: [Circular *1] } } }
```

## `assign-deep`, `merge`

```jsx
const f = require("assign-deep");
// or
const f = require("merge").recursive;

// test undefined values
f({}, { foo: 5 }, { foo: undefined });
// { foo: undefined }

// test arrays
f({}, { foo: [1, 2, 3] }, { foo: [4] });
// { foo: [ 4 ] }

// test circular referencces
const obj = { foo: {} };
obj.foo.bar = obj;
f({}, obj);
// { foo: <ref *1> { bar: { foo: [Circular *1] } } }
```

## `deep-assign-lite`

```jsx
const f = require("deep-assign-lite");

// test undefined values
f({}, { foo: 5 }, { foo: undefined });
// { foo: 5 }

// test arrays
f({}, { foo: [1, 2, 3] }, { foo: [4] });
// { foo: [ 4 ] }

// test circular referencces
const obj = { foo: {} };
obj.foo.bar = obj;
f({}, obj);
// Uncaught RangeError: Maximum call stack size exceeded
```

---
</details>

I'll look into where this function is used to see what the differences would mean for actual usage.

stephenwade: Below I've evaluated the behavior changes of merging undefined values, merging arrays, and circular references as they apply to each use of `extend`.

<details>
<summary>lib/linter/linter.js</summary>

https://github.com/eslint/eslint/blob/b3d6d9a7b2108e5de4fccbde248f0b4fae00254b/lib/linter/linter.js#L530-L532

- Undefined values could be an issue if a plugin defines an environment with an `undefined` value somewhere.
- Array merging and circular references are not an issue since the [`ParserOptions` type](https://github.com/eslint/eslint/blob/b3d6d9a7b2108e5de4fccbde248f0b4fae00254b/lib/shared/types.js#L14-L26) does not allow arrays or objects as values.

https://github.com/eslint/eslint/blob/b3d6d9a7b2108e5de4fccbde248f0b4fae00254b/lib/linter/linter.js#L533

- Undefined values could be an issue if the user provides a `parserOptions` object with an `undefined` value somewhere.
- Array merging and circular references are not an issue since the [`ParserOptions` type](https://github.com/eslint/eslint/blob/b3d6d9a7b2108e5de4fccbde248f0b4fae00254b/lib/shared/types.js#L14-L26) does not allow arrays or objects as values.

---
</details>

<details>
<summary>lib/rule-tester/rule-tester.js</summary>

https://github.com/eslint/eslint/blob/b3d6d9a7b2108e5de4fccbde248f0b4fae00254b/lib/rule-tester/rule-tester.js#L328-L333

- Undefined values, array merging, and circular references are not issues. I looked at all the uses of the `RuleTester` constructor and none of them use `undefined`, arrays, or circular references.

https://github.com/eslint/eslint/blob/b3d6d9a7b2108e5de4fccbde248f0b4fae00254b/lib/rule-tester/rule-tester.js#L372

- Undefined values and array merging are not an issue since one of the parameters is an empty object.
- Circular references are not an issue since `testerDefaultConfig` is [defined in this file](https://github.com/eslint/eslint/blob/b3d6d9a7b2108e5de4fccbde248f0b4fae00254b/lib/rule-tester/rule-tester.js#L112) and does not contain circular references.

https://github.com/eslint/eslint/blob/b3d6d9a7b2108e5de4fccbde248f0b4fae00254b/lib/rule-tester/rule-tester.js#L468

- Undefined values, array merging, and circular references are not issues. `testerConfig` is defined in the `RuleTester` constructor, and none of the uses of the `RuleTester` constructor use `undefined`, arrays, or circular references.

https://github.com/eslint/eslint/blob/b3d6d9a7b2108e5de4fccbde248f0b4fae00254b/lib/rule-tester/rule-tester.js#L490-L493

- I can't look through all tests to check the config on each one, but since the tests still pass I don't think any of them use `undefined` or circular references.
- Array merging is not an issue since `config` is a clone of `testerConfig` which is defined in the `RuleTester` constructor, and none of the uses of the `RuleTester` constructor use arrays.

---
</details>

<details>
<summary>tests/lib/rules/no-invalid-this.js</summary>

https://github.com/eslint/eslint/blob/b3d6d9a7b2108e5de4fccbde248f0b4fae00254b/tests/lib/rules/no-invalid-this.js#L73

- Undefined values and array merging are not an issue since one of the parameters is an empty object.
- Circular references are not an issue since each `pattern` is [defined in this file](https://github.com/eslint/eslint/blob/b3d6d9a7b2108e5de4fccbde248f0b4fae00254b/tests/lib/rules/no-invalid-this.js#L103) and no pattern contains circular references.

---
</details>

Summary: I believe that the differences in array merging and circular references will not affect end users. The differences in merging `undefined` values might affect users of `parserOptions` and custom environments.

There are 2 replacements I [tested](https://github.com/eslint/eslint/pull/14287#discussion_r628662466) that ignore `undefined` values, and neither are very popular by download count. Of the 2 options, I think [@common-utilities/merge-objects](https://www.npmjs.com/package/@common-utilities/merge-objects) is the better option, but I would understand not wanting to use either one based on their download counts.

In my view, the options are:
1. Switch to a not-popular package for merging objects so the `undefined` behavior doesn't change.
2. Accept that the behavior of `undefined` will change and wait until ESLint v8 to merge this PR (since it would be a breaking change).

/cc @mdjermanovic What are your thoughts?

mdjermanovic: I'd guess that none of the changes from `lodash.merge/lodash.cloneDeep` to `deep-extend` would cause a significant impact, but to be safe, maybe we can use [lodash.merge package](https://www.npmjs.com/package/lodash.merge) in this PR (which could be merged for ESLint v7), and replace it with `deep-extend` in ESLint v8.0.0.

@stephenwade @aladdin-add @eslint/eslint-tsc thoughts?

nzakas: I think we can probably just use lodash.merge. I’m not sure there’s a value in doing another switch after that because I think all of this will end up handled inside FlatConfigArray, so we will be able to remove any other merge-related dependencies when eslintrc is purged.

stephenwade: I think it would make sense to use `lodash.merge` for now, but use of the per-method packages is [discouraged by Lodash](https://lodash.com/per-method-packages) so I think we should switch away from them in ESLint v8.0.0.

I saw that a [`deepMerge` function](https://github.com/eslint/eslint/pull/14321/files#diff-93aef33006c32545a3aa83cf3dfedd4959e55327e14abf62744aa845c5d789d8R54-R60) is going to be included in the new config system (#14321). Could we extract that into `lib/shared` and use it for v8 instead of `lodash.merge` or `deep-extend`?

cc: @nzakas @mdjermanovic

nzakas: We could. I haven’t verified that it works exactly the same as lodash.merge, though. So we would need to do some further testing. If you’re up for it, I’d say go for it.

lib/source-code/token-store/utils.js
```diff
@@ -36,11 +30,10 @@ function getStartLocation(token) {
...
 exports.search = function search(tokens, location) {
-    return lodash.sortedIndexBy(
-        tokens,
-        { range: [location] },
-        getStartLocation
-    );
+    const val = getStartLocation({ range: [location] });
+    const index = tokens.findIndex(el => val <= getStartLocation(el));
+
+    return index === -1 ? tokens.length : index;
```
mdjermanovic, line 36: Similar to the previous comment, `lodash.sortedIndexBy` most likely uses a binary search (and jsdoc description for this `function search` says "Binary-searches..."), while `findIndex` performs a linear search.

I think `utils.search` is used only on arrays of comments, which typically shouldn't be large arrays, so the difference _might_ be small.

@eslint/eslint-tsc thoughts about this, should we look for a binary search library?

mysticatea: I also think the difference might be small.

By the way, `getStartLocation({ range: [location] })` is `{ range: [location] }.range[0]`, so it does nothing.

mdjermanovic (changes requested):

lib/source-code/source-code.js
```diff
@@ -534,7 +533,8 @@ class SourceCode extends TokenStore {
          * To figure out which line rangeIndex is on, determine the last index at which rangeIndex could
          * be inserted into lineIndices to keep the list sorted.
          */
-        const lineNumber = lodash.sortedLastIndex(this.lineStartIndices, index);
+        const rangeIndex = [...this.lineStartIndices].reverse().findIndex(el => index >= el);
+        const lineNumber = rangeIndex === -1 ? 0 : this.lineStartIndices.length - rangeIndex;
```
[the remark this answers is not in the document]
stephenwade: Suggestion committed in cba69f23c.

lib/source-code/token-store/utils.js
stephenwade: Useless `val` declaration removed in 00ac18335.

lib/source-code/source-code.js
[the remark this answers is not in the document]
stephenwade: findIndex is slower, but not exponentially so, and both operations can run millions of times per second.

```
$ node benchmark-getLocFromIndex.js
sortedLastIndex x 10,168,021 ops/sec ±1.34% (90 runs sampled)
Array#findIndex x 2,795,631 ops/sec ±1.41% (89 runs sampled)
```

Benchmark code: https://github.com/stephenwade/my-eslint-benchmark

lib/source-code/token-store/utils.js
stephenwade: findIndex is slower, but not exponentially so, and both operations can run millions of times per second.

findIndex is slower, but not exponentially so, and both operations can run millions of times per second.

```
$ node benchmark-search.js
sortedIndexBy x 7,188,306 ops/sec ±0.84% (89 runs sampled)
Array#findIndex x 2,764,162 ops/sec ±1.65% (84 runs sampled)
```

Benchmark code: https://github.com/stephenwade/my-eslint-benchmark

stephenwade: I used [hyperfine](https://github.com/sharkdp/hyperfine) to benchmark the test suite on this branch and the master branch. This branch ran slightly faster on my computer, although the results were very close.

Screenshot of benchmark results running the npm test command. The remove-lodash-3 branch runs in about 172 seconds and the master branch runs in about 174 seconds.

I believe that the speed difference in this method will not be noticeable and that we should stick with findIndex.

lib/source-code/source-code.js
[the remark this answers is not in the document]
stephenwade: See https://github.com/eslint/eslint/pull/14287/commits/02510b2c35222d12741ac469753156e86a41f844#r636050582 for project-wide benchmarking results. I believe the speed difference in this function will not be noticeable.

lib/cli-engine/formatters/html-template-message.js
```diff
@@ -0,0 +1,25 @@
...
+        ruleId
+    } = it;
+
+    return `
+<tr style="display:none" class="f-${parentIndex}">
+    <td>${lineNumber}:${columnNumber}</td>
+    <td class="clr-${severityNumber}">${severityName}</td>
+    <td>${encodeHTML(message)}</td>
+    <td>
+        <a href="${ruleUrl}" target="_blank" rel="noopener noreferrer">${ruleId ? ruleId : ""}</a>
```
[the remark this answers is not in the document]
stephenwade: Resolved in e4d94ce7c.

lib/cli-engine/formatters/html.js
```diff
@@ -4,17 +4,31 @@
...
-const messageTemplate = lodash.template(fs.readFileSync(path.join(__dirname, "html-template-message.html"), "utf-8"));
-const resultTemplate = lodash.template(fs.readFileSync(path.join(__dirname, "html-template-result.html"), "utf-8"));
+const encodeHTML = (function() {
+    const encodeHTMLRules = {
+        "&": "&#38;",
+        "<": "&#60;",
+        ">": "&#62;",
+        '"': "&#34;",
+        "'": "&#39;",
+        "/": "&#47;"
```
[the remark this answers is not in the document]
stephenwade: Resolved in e4d94ce7c.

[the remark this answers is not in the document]
stephenwade: I had copied the list of characters from [doT](https://github.com/olado/doT/blob/031d3bb7520eed6b93886df2b650b7fce12a7007/doT.js#L33), but I don't think the `/` is necessary. I've removed it.

mdjermanovic (approved):
LGTM, thanks for the great work!

mdjermanovic:
Thanks for contributing!

PR #14321, opened 2021-04-13
Update: Implement FlatConfigArray (refs #13481)

nzakas:
#### Prerequisites checklist

[ ] Documentation update
[ ] Add autofixing to a rule
[ ] Add a CLI option
[x] Add something to the core
[ ] Other, please explain:

Implements FlatConfigArray (refs #13481)

I added `FlatConfigArray` class and associated tests to make sure everything is working okay. `FlatConfigArray` is not referenced anywhere while running ESLint currently, as implementing `FlatConfigArray` is meant to be a standalone task. In the next task, I'll work on using it when `eslint.config.js` is found.

Merging this does *not* result in any end-user-facing changes. It will result in `FlatConfigArray` tests being run with `npm test`

1. Are there any cases I missed?
2. Is the code easy enough to follow? (The old config system was so complicated that the code was hard to follow, I want this version to be easier for people to look at and figure out what's going on.)


lib/config/flat-config-schema.js
```diff
@@ -0,0 +1,372 @@
...
+
+
+/**
+ * Validates that a given string is the form pluginName/objectName.
+ * @param {string} value The string to check.
+ * @returns {void}
+ * @throws {TypeError} If the string isn't in the correct format.
+ */
+function assertIsPluginMemberName(value) {
+    if (!/[@a-z0-9-_$]+\/[a-z0-9-_$]+/iu.test(value)) {
```
btmills, line 77: Are we intentionally precluding scoped package-style plugin names here? I realize that the plugin's name is an arbitrary object key that doesn't need to match the npm package name, but I wouldn't be surprised if matching package names becomes a common pattern among users for simplicity.

lib/config/flat-config-schema.js
```diff
@@ -0,0 +1,372 @@
...
+            throw new TypeError("Expected \"script\", \"module\", or \"commonjs\".");
+        }
+    }
+};
+
+//-----------------------------------------------------------------------------
+// Full schema
+//-----------------------------------------------------------------------------
+
+exports.flatConfigSchema = {
```
btmills, line 352: This PR doesn't include the `name` key that was included in the RFC. Is that an intentional omission?

lib/config/default-config.js
```diff
@@ -0,0 +1,48 @@
...
+
+    has(target, property) {
+        return Rules.has(property);
+    }
+});
+
+exports.defaultConfig = [
+    {
+        plugins: {
+            "@": {
```
btmills, line 36: Is there any reason we'd want this to be a `Symbol` rather than a special string?

lib/config/flat-config-schema.js
```diff
@@ -0,0 +1,372 @@
...
+const globalsSchema = {
+    merge: "assign",
+    validate(value) {
+
+        assertIsObject(value);
+
+        for (const key of Object.keys(value)) {
+
+            // avoid hairy edge case
+            if (key === "__proto__") {
```
btmills, line 142: This check occurs in a number of places. Is this class of edge cases sufficiently hairy that it's worth looking into an implementation (like using `Map`s internally instead) that avoids them entirely? Alternatively, are we comfortable with a few footguns lying around in this file since we ideally won't need to visit it often once this is merged?

btmills (commented):
I am pleasantly surprised by how much less is going on here than in the existing config system. Outsourcing all of the file path matching to `@humanwhocodes/config-array` nicely separates the "what needs to be merged" logic from the "what are the rules for merging each field"  logic. It's great to see the vision of a simpler config coming together :tada:

lib/config/flat-config-schema.js
nzakas: Yes. The `name` key is one of the default keys that are automatically included with `ConfigArray`. The other two are `files` and `ignores`. These three are used internally so they need a very specific validation/merge strategy for everything to work, hence, they are defined by default.

lib/config/flat-config-schema.js
nzakas: I copied this check over from the eslintrc functionality figuring it was probably added for a reason. I think switching to maps at this point is probably premature and will add complexity that isn't really necessary for the number of times this edge case might occur.

lib/config/flat-config-schema.js
nzakas: I suppose I was assuming that people would stop using the fully scoped package name in the plugin reference once the config was in JavaScript. It probably doesn't hurt to allow it, though.

tests/lib/config/flat-config-array.js
```diff
@@ -0,0 +1,1153 @@
...
+
+                await assertInvalidConfig([
+                    {
+                        languageOptions: {
+                            foo: true
+                        }
+                    }
+                ], "Unexpected key \"foo\" found.");
+
+            });
```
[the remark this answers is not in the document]
nzakas: I'll add it. I was struggling to come up with all possible variations.

nzakas:
I believe everything has been addressed, but let me know if I missed anything.

lib/config/flat-config-schema.js
btmills: Ah, I missed that `name` is a first-class part of `ConfigArray`. Thanks for explaining.

lib/config/flat-config-schema.js
btmills: I couldn’t find any occurrence of these checks prior to https://github.com/eslint/eslint/pull/11546, so I’ll assume they were introduced there to guard against prototype pollution. I have no problem leaving this as-is given it’s been this way for a couple years and is only rarely touched.

lib/config/flat-config-schema.js
```diff
@@ -0,0 +1,362 @@
...
+
+/** @type {ObjectPropertySchema} */
+const processorSchema = {
+    merge: "replace",
+    validate(value) {
+        if (typeof value === "string") {
+            assertIsPluginMemberName(value);
+        } else if (value && typeof value === "object") {
+            if (typeof value.preprocess !== "function" || typeof value.postprocess !== "function") {
+                throw new TypeError("Object must have a preprocess() and a postprocess() method.");
```
btmills, line 226: There's not currently a test exercising this check.

lib/config/flat-config-schema.js
```diff
@@ -0,0 +1,362 @@
...
+        // manually validate that plugins are not redefined
+        for (const key of keys) {
+
+            // avoid hairy edge case
+            if (key === "__proto__") {
+                continue;
+            }
+
+            if (key in first && key in second && first[key] !== second[key]) {
+                throw new TypeError(`Cannot redefine plugin "${key}".`);
```
btmills, line 188: There's not currently a test exercising this check.

lib/config/flat-config-schema.js
```diff
@@ -0,0 +1,362 @@
...
+
+        for (const key of Object.keys(value)) {
+
+            // avoid hairy edge case
+            if (key === "__proto__") {
+                continue;
+            }
+
+            if (key !== key.trim()) {
+                throw new TypeError(`Global "${key}" has leading or trailing whitespace.`);
```
btmills, line 147: There's not currently a test exercising this check.

lib/config/flat-config-array.js
```diff
@@ -0,0 +1,125 @@
...
+
+            languageOptions.parser = plugins[pluginName].parsers[parserName];
+        }
+
+        // Check processor value
+        if (processor && typeof processor === "string") {
+            const { pluginName, objectName: processorName } = splitPluginIdentifier(processor);
+
+            if (!plugins || !plugins[pluginName] || !plugins[pluginName].processors || !plugins[pluginName].processors[processorName]) {
+                throw new TypeError(`Key "processor": Could not find "${processorName}" in plugin "${pluginName}".`);
```
btmills, line 111: There's not currently a test exercising this check.

btmills (approved):
I just did a thorough pass through the schema to make sure everything matches up. I'm excited that we'll be able to alert users to invalid configurations as close to the source as possible with some of the checks you've added. There were a few particularly helpful ones that aren't covered by a test, so I pointed those out in case you'd like to add those. LGTM either way.

nzakas:
Nice! I’ll plan on adding tests for those.

lib/config/flat-config-schema.js
nzakas: Looks like it's tested on line 369. (For some reason, can't get a direct link.)

lib/config/flat-config-schema.js
nzakas: Tested on line 260.

lib/config/flat-config-schema.js
btmills: I was searching for the whole string, but I see the test covers the important part. It's quite possible I made the same mistake for the other checks I commented on.

lib/config/flat-config-schema.js
nzakas: Added

lib/config/flat-config-array.js
nzakas: Added

btmills (approved):
Nice! Sorry about the false positive on a couple of those.

lib/config/flat-config-array.js
```diff
@@ -0,0 +1,125 @@
...
+ */
+class FlatConfigArray extends ConfigArray {
+
+    /**
+     * Creates a new instance.
+     * @param {*[]} configs An array of configuration information.
+     * @param {{basePath: string, baseConfig: FlatConfig}} options The options
+     *      to use for the config array instance.
+     */
+    constructor(configs, { basePath, baseConfig = defaultConfig }) {
```
mdjermanovic, line 55: Is `baseConfig` parameter necessary, shouldn't we always use `defaultConfig` as the first element?

lib/config/default-config.js
```diff
@@ -0,0 +1,48 @@
...
+                parsers: {
+                    espree: require("espree")
+                },
+                rules: defaultRulesConfig
+            }
+        },
+        ignores: ["node_modules/**"],
```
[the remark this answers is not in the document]
nzakas: These are all glob patterns now. I’ll update them to match the current defaults.

lib/config/flat-config-array.js
nzakas: I use this in the tests to avoid needing to load the entire default config when I need to test specific functionality that shouldn’t be impacted by any changes to the default config.

lib/config/default-config.js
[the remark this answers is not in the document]
mdjermanovic: Does it also mean that we'll interpret the content of `.eslintignore` file differently, depending on the format of the found eslint configuration file: 

* as a list of `.gitignore` patterns (actual behavior) if the config is `.eslintrc.*`.
* as a list of glob patterns if the config is `eslint.config.js`

lib/config/flat-config-array.js
```diff
@@ -0,0 +1,125 @@
...
+ * @author Nicholas C. Zakas
+ */
+
+"use strict";
+
+//-----------------------------------------------------------------------------
+// Requirements
+//-----------------------------------------------------------------------------
+
+const { ConfigArray, ConfigArraySymbol } = require("@humanwhocodes/config-array");
```
mdjermanovic, line 12: Will `ConfigArray` support negated patterns (prefixed with `!`) in `files` and/or `ignores`?

lib/config/default-config.js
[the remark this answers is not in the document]
nzakas: No. There is no change to how .eslintignore will be interpreted. It will be read and compiled into a function that is then included in a flat config ignores key. (Can be either a glob pattern or a function.)

lib/config/flat-config-array.js
nzakas: It already does

mdjermanovic: What's the expected behavior in this case:

```js
{
    files: ["*.js", "!foo.js"]
}
```

By the RFC, it might be expected that the config applies to all `*.js` files except `foo.js`:

> These patterns can be negated by prefixing them with `!`, which effectively mimics the behavior of `excludedFiles` in `.eslintrc`.

However, `ConfigArray` checks if the given file path matches at least one of the patterns in the `files` array, so the config will apply to any file path (`a.js`, `foo.js`, `a.ts`, `package.json`...).

lib/config/flat-config-schema.js
```diff
@@ -0,0 +1,362 @@
...
+
+
+/**
+ * Validates that a given string is the form pluginName/objectName.
+ * @param {string} value The string to check.
+ * @returns {void}
+ * @throws {TypeError} If the string isn't in the correct format.
+ */
+function assertIsPluginMemberName(value) {
+    if (!/[@a-z0-9-_$]+\/(?:[a-z0-9-_$]+)+/iu.test(value)) {
```
mdjermanovic, line 77: The second `+` in `(?:[a-z0-9-_$]+)+` looks unnecessary (and might cause catastrophic backtracking if we add `$` anchor or anything else after `+` in the pattern), did you mean something else?

lib/config/flat-config-schema.js
```diff
@@ -0,0 +1,362 @@
...
+    [0, 0], ["off", 0],
+    [1, 1], ["warn", 1],
+    [2, 2], ["error", 2]
+]);
+
+const globalVariablesValues = new Set([
+    true, "writable", "writeable",
+    false, "readonly",
+    "off"
+]);
```
mdjermanovic, line 34: Aside from these, we also allow: `"true"` (string), `null`, `"false"` (string), and `"readable"`.

https://github.com/eslint/eslintrc/blob/62ea4bd74b78fbeff12ffb21f1f978817601d4d1/lib/shared/config-ops.js#L108-L129

lib/config/flat-config-schema.js
```diff
@@ -0,0 +1,362 @@
...
+            if (key === "__proto__") {
+                continue;
+            }
+
+            if (key !== key.trim()) {
+                throw new TypeError(`Global "${key}" has leading or trailing whitespace.`);
+            }
+
+            if (!globalVariablesValues.has(value[key])) {
+                throw new TypeError("Expected \"readonly\", \"writeable\", or \"off\".");
```
mdjermanovic, line 151:
```suggestion
                throw new TypeError("Expected \"readonly\", \"writable\", or \"off\".");
```

lib/config/flat-config-array.js
nzakas: This behavior is pretty well-defined in the RFC: `files` are treated as OR statements. If you want to do an AND, then you need to group them in an array:

```
{
    files: [["*.js", "!foo.js"]]
}
```

lib/config/flat-config-schema.js
nzakas: "writable" is a misspelling, so we don't document that even though we support it. (This option was incorrectly implemented with the misspelling initially.)

lib/config/flat-config-schema.js
nzakas: Ooh nice catch. Yeah, the intent was to allow `foo/bar/baz` to map what we have in eslintrc, but clearly I messed that pattern up.

nzakas:
Comments addressed.

mdjermanovic:
@nzakas regarding https://github.com/eslint/eslint/pull/14321#discussion_r622205891, isn't it the opposite (`writable` - correct, `writeable` - misspelling)?

Our [documentation](https://eslint.org/docs/user-guide/configuring/language-options#specifying-globals) uses `"writable"`, and says that `"writeable"` can be used for historical reasons.

lib/config/default-config.js
```diff
@@ -0,0 +1,51 @@
...
+            "@": {
+                parsers: {
+                    espree: require("espree")
+                },
+                rules: defaultRulesConfig
+            }
+        },
+        ignores: [
+            "**/node_modules/**",
+            ".*"
```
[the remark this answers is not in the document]
nzakas: I copied this from the eslintrc default ignore patterns for consistency.

lib/config/flat-config-schema.js
```diff
@@ -0,0 +1,362 @@
...
+
+const ruleSeverities = new Map([
+    [0, 0], ["off", 0],
+    [1, 1], ["warn", 1],
+    [2, 2], ["error", 2]
+]);
+
+const globalVariablesValues = new Set([
+    true, "true", "writable", "writeable",
+    false, "false", "readonly", "readable",
```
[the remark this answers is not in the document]
nzakas: Geez, I had no idea that was allowed. But yes, we should support that for compat sake

lib/config/rule-validator.js
```diff
@@ -0,0 +1,155 @@
...
+                const schema = getRuleOptionsSchema(rule);
+
+                if (schema) {
+                    this.validators.set(rule, ajv.compile(schema));
+                }
+            }
+
+            const validateRule = this.validators.get(rule);
+
+            if (validateRule && Array.isArray(config.rules[ruleId])) {
```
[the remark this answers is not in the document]
nzakas: Good point. I’ll update that.

lib/config/default-config.js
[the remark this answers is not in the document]
mdjermanovic: Yes, but doesn't the RFC include the change in default patterns?

> With `eslint.config.js`, there are three ways that files and directories can be ignored:

> 1. **Defaults** - by default, ESLint will ignore `node_modules` and `.git` directories only. This is different from the current behavior where ESLint ignores `node_modules` and all files/directories beginning with a dot (`.`).

lib/config/rule-validator.js
```diff
@@ -0,0 +1,155 @@
...
+     * @returns {void}
+     * @throws {Error} If a rule's configuration does not match its schema.
+     */
+    validate(config) {
+
+        if (!config.rules) {
+            return;
+        }
+
+        for (const ruleId of Object.keys(config.rules)) {
```
mdjermanovic, line 121: We are currently not doing any validations for rules that are set to `0`/`"off"` severity.

This configuration is valid with the actual config system:

```js
rules: {
        "dot-location": [0, "foo"], // invalid option
        "foo": 0, // rule doesn't exist
        "foo/foo": 0, // plugin doesn't exist
    }
```

lib/config/default-config.js
```diff
@@ -0,0 +1,51 @@
...
+//-----------------------------------------------------------------------------
+// Helpers
+//-----------------------------------------------------------------------------
+
+/*
+ * Because we try to delay loading rules until absolutely necessary, a proxy
+ * allows us to hook into the lazy-loading aspect of the rules map while still
+ * keeping all of the relevant configuration inside of the config array.
+ */
+const defaultRulesConfig = new Proxy({}, {
```
mdjermanovic, line 23: Why we named this "Config"? It's an object with built-in rules, not a configuration?

lib/config/flat-config-array.js
```diff
@@ -0,0 +1,125 @@
...
+        if (config === "eslint:recommended") {
+            return recommendedConfig;
+        }
+
+        if (config === "eslint:all") {
+            return allConfig;
+        }
+
+        return config;
+    }
```
mdjermanovic, line 82: If user makes a typo, like `"eslint:recomended"`, they'll get message `Error: All arguments must be objects.`

Might be nice to throw an error saying what `string` value was found and which `string` values are allowed?

lib/config/rule-validator.js
```diff
@@ -0,0 +1,155 @@
...
+        pluginName = ruleIdParts.join("/");
+    }
+
+    if (!config.plugins || !config.plugins[pluginName]) {
+        throw new TypeError(`Key "rules": Key "${ruleId}": Could not find plugin "${pluginName}".`);
+    }
+
+    if (!config.plugins[pluginName].rules || !config.plugins[pluginName].rules[ruleName]) {
+        throw new TypeError(`Key "rules": Key "${ruleId}": Could not find "${ruleName}" in plugin "${pluginName}".`);
+    }
```
mdjermanovic, line 43: Should we have a different message for built-in rules, this might be confusing:

`TypeError: Key "rules": Key "no-undeff": Could not find "no-undeff" in plugin "@".`

lib/config/default-config.js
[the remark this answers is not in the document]
nzakas: Huh. It appears so! Thanks for double-checking that. I seem to be caught in this maze of some areas where we wanted things to be consistent with eslintrc and some where they are different. Hard to keep track.

lib/config/flat-config-schema.js
```diff
@@ -0,0 +1,362 @@
...
+            }
+
+            /*
+             * In any other situation, then the second rule config takes
+             * precedence. If it's an array, we return a copy;
+             * otherwise we return the value, which is just the severity.
+             */
+            result[ruleId] = secondIsArray
+                ? second[ruleId].slice(0)
+                : second[ruleId];
```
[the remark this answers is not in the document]
nzakas: Sorry, what is the "actual config system"?

lib/config/flat-config-schema.js
```diff
@@ -0,0 +1,362 @@
...
+};
+
+/** @type {ObjectPropertySchema} */
+const processorSchema = {
+    merge: "replace",
+    validate(value) {
+        if (typeof value === "string") {
+            assertIsPluginMemberName(value);
+        } else if (value && typeof value === "object") {
+            if (typeof value.preprocess !== "function" || typeof value.postprocess !== "function") {
```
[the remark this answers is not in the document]
nzakas: eslintrc doesn't check for these, so you get a hairy error message if they are missing. I'd like to be more explicit in checking for these.

lib/config/flat-config-schema.js
```diff
@@ -0,0 +1,362 @@
...
+            if (key === "__proto__") {
+                continue;
+            }
+
+            if (key !== key.trim()) {
+                throw new TypeError(`Global "${key}" has leading or trailing whitespace.`);
+            }
+
+            if (!globalVariablesValues.has(value[key])) {
+                throw new TypeError("Expected \"readonly\", \"writeable\", or \"off\".");
```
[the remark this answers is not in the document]
nzakas: Good idea :+1:

lib/config/flat-config-schema.js
```diff
@@ -0,0 +1,362 @@
...
+            reportUnusedDisableDirectives: booleanSchema
+        }
+    },
+    languageOptions: {
+        schema: {
+            ecmaVersion: numberSchema,
+            sourceType: sourceTypeSchema,
+            globals: globalsSchema,
+            parser: parserSchema,
+            parserOptions: objectAssignSchema
```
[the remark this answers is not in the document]
nzakas: Fixed.

lib/config/rule-validator.js
nzakas: This is something we can address later. I don't want to tie up this PR one small UX discussions like this.

lib/config/default-config.js
nzakas: Fair point.

lib/config/flat-config-schema.js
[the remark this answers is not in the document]
nzakas: I'm guessing "actual config system" means eslintrc? Can you please use the term eslintrc instead, I've been having a hard time understanding these comments.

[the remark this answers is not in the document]
mdjermanovic: Yes eslintrc, sorry for the confusion.

lib/config/flat-config-array.js
nzakas: I'm going to hold off on addressing error cases until flat config is integrated into ESLint. We have a lot of options for messaging error conditions, and I don't think it's going to be clear what exactly is needed where until we can play with it.

lib/config/rule-validator.js
nzakas: Got it. Thanks.

lib/config/flat-config-schema.js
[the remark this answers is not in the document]
nzakas: Should be fixed now.

nzakas:
I think I've fixed everything. @mdjermanovic thanks for the amazing level of detail in your comments.

tests/lib/config/flat-config-array.js
```diff
@@ -0,0 +1,1413 @@
...
+ * @param {Object} rulesConfig The rules config portion of a config.
+ * @returns {Array} The rules config object.
+ */
+function normalizeRuleConfig(rulesConfig) {
+    for (const ruleId of Object.keys(rulesConfig)) {
+        rulesConfig[ruleId] = [2];
+    }
+
+    return rulesConfig;
+}
```
mdjermanovic, line 124: Some tests unrelated to this PR [are failing](https://github.com/eslint/eslint/pull/14321/checks?check_run_id=2494731957), probably because this directly modifies the given object (`eslint-recommended` and `eslint-all`).

lib/config/flat-config-schema.js
```diff
@@ -0,0 +1,444 @@
...
+
+    for (const key of Object.keys(result)) {
+
+        // avoid hairy edge case
+        if (key === "__proto__") {
+            continue;
+        }
+
+        const firstValue = first[key];
+        const secondValue = second[key];
```
mdjermanovic, line 81: In some cases, this function will be called with `second === undefined`, so `second[key]` would crash.

Example:

```js
[
    {
        languageOptions: {
            parserOptions: {
                ecmaFeatures: {
                    jsx: true
                }
            }
        }
    },
    {
        languageOptions: {
            ecmaVersion: 2021
        }
    }
]
```

Throws `TypeError: Key "languageOptions": Key "parserOptions": Cannot read property 'jsx' of undefined`

lib/config/flat-config-schema.js
```diff
@@ -0,0 +1,444 @@
...
+    return typeof value === "undefined";
+}
+
+/**
+ * Deeply merges two objects.
+ * @param {Object} first The base object.
+ * @param {Object} second The overrides object.
+ * @returns {Object} An object with properties from both first and second.
+ */
+function deepMerge(first, second) {
```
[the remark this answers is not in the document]
nzakas: Good catch. I think the expectation here is that an array overwrites an array. I'll double-check that.

tests/lib/config/flat-config-array.js
nzakas: Doh! Yup, I'll fix that. Thanks.

nzakas:
Should be set now (hopefully).

lib/config/flat-config-schema.js
```diff
@@ -0,0 +1,452 @@
...
+     * overwrite properties from the first. This sets up a baseline to use
+     * later rather than needing to inspect and change every property
+     * individually.
+     */
+    const result = {
+        ...first,
+        ...second
+    };
+
+    for (const key of Object.keys(second)) {
```
stephenwade, line 81:
```suggestion
    for (const key of Object.keys(result)) {
```

If you only check keys in `second`, you fail to deeply clone objects in `first`. Since this returns a new object instead of modifying `first`, I think that all objects should be deeply cloned to prevent surprises.

```js
const o1 = { a: { x: 1, y: 2 } };
const o2 = { b: { z: 3 } };

const result = deepMerge(o1, o2);

result.b.z = 6;
console.log(o2.b.z) // 3 ✅

result.a.x = 5;
console.log(o1.a.x) // 5 🚫
```

nzakas: Good point.

lib/config/flat-config-array.js
```diff
@@ -0,0 +1,125 @@
...
+ *      name.
+ */
+function splitPluginIdentifier(identifier) {
+    const parts = identifier.split("/");
+
+    return {
+        objectName: parts.pop(),
+        pluginName: parts.join("/")
+    };
+}
```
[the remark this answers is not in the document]
mdjermanovic: The same logic is used in rule validator:

https://github.com/eslint/eslint/blob/86b7bddb421031db1a5c517a6ed6d1f23b388ba7/lib/config/rule-validator.js#L24-L35

so it would look for `es-syntax` rule in `node/no-unsupported-features` plugin, instead of `no-unsupported-features/es-syntax` rule in `node` plugin.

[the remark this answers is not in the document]
nzakas: So far this seems to be working okay with my tests. But again, this is just the first draft, so we have plenty of time to find and address errors. It will be easier to find once I start integrating.

mdjermanovic (approved):
LGTM!

I'm not sure if the feature of unignoring files that are ignored in the base config (or a shared config) is supported, but we can discuss that when we start integrating.

PR #14573, opened 2021-05-06
Breaking: Require `meta.hasSuggestions` for rules with suggestions

### LINKED ISSUE #14312: Reevaluate the `meta.docs.suggestion` property

There is a rule property `meta.docs.suggestions` mentioned in the [Rule Basics](https://eslint.org/docs/developer-guide/working-with-rules#rule-basics) documentation that was added along with the [Suggestions API](https://eslint.org/docs/developer-guide/working-with-rules#providing-suggestions) / [Suggestions RFC](https://github.com/eslint/rfcs/pull/30):

There are a few potential problems with this property:

1. It is missing from some of ESLint's examples / test cases depicting usage of the suggestions API (more discussion in: https://github.com/eslint/eslint/pull/14298).
2. Its name and location are inconsistent with a related property `meta.fixable`.
3. Its presence is not enforced, so humans and tooling cannot reliably use it for determining whether a rule provides suggestions (as opposed to `meta.fixable` whose presence is enforced for autofixable rules) (more discussion in: https://github.com/eslint/eslint/pull/14292).

Note that ESLint itself doesn't currently need to know which rules are suggestable (provide suggestions), whereas it does need to know which rules are fixable, so differences between `meta.docs.suggestions` and `meta.fixable` could be justified if we wanted to keep `meta.docs.suggestions` as an optional property (as opposed to `meta.fixable` which is a required property).

**Your take on the correct solution to the problem.**

There are several potential solutions here of varying size.

1. Keep `meta.docs.suggestion` as an optional property, fully document it, and add a lint rule [eslint-plugin/require-meta-docs-suggestion](https://github.com/not-an-aardvark/eslint-plugin-eslint-plugin/pull/105) for those wishing to enforce its presence. 
    - Maintains the status quo, no disruption to anyone
    - The property would still be of limited usefulness as we can't assume plugin rules would provide it
2. Keep `meta.docs.suggestion`, and begin to enforce its presence as a required property for suggestable rules.  Implementation in https://github.com/eslint/eslint/pull/14292.
    - Ensures that humans and tooling have a reliable means of determining which rules are suggestable
    - But now there would be a noticeable naming/location inconsistency between this property and `meta.fixable` which would both be required properties
    - This is a breaking change that requires a one-line change in all rules that forgot to specify this property
3. Remove the `meta.docs.suggestion` property entirely and do not add a replacement for it.
    - Simplifies the API by removing a potentially unnecessary property
    - But now there's no recommended way to determine what rules provide suggestions
4. Rename the property to `meta.suggestable` for consistency with `meta.fixable`, and enforce its presence for suggestable rules.
     - Ensures that humans and tooling have a reliable means of determining which rules are suggestable
     - Cleans up the API by addressing the naming/location inconsistency with `meta.fixable` 
     - This is a breaking change that requires a one-line change in all suggestable rules

My preferred solution is 4 because it yields the most useful and consistent end-result, although this is also the most disruptive change, so I'm interested to hear what others prefer.

Yes

bmish:
#### Prerequisites checklist

[ ] Documentation update
[] Add autofixing to a rule
[ ] Add a CLI option
[X] Add something to the core
[ ] Other, please explain:

Fixes #14312.

* Renames the old `meta.docs.suggestion` property to `meta.hasSuggestions` to indicate whether a rule provides suggestions (improved clarity and alignment with the `meta.fixable` property which is mandatory for fixable rules)
* Enforces the presence of the new property for rules that provide suggestions (the presence of the old property was not enforced)
* Enforces that the old property is no longer provided
* Updates core rules that provide suggestions to use the new property
* Adds the new property to test cases and documentation (some of which were missing it before)

This ensures we have a clear and reliable property to use for both humans and tooling to determine whether a rule provides suggestions.

Before merging:

- [ ] Double-check that all core rules with suggestions have been updated with the new property


lib/linter/linter.js
```diff
@@ -922,6 +922,12 @@ function runRules(sourceCode, configuredRules, ruleMapper, parserOptions, parser
                         if (problem.fix && rule.meta && !rule.meta.fixable) {
                             throw new Error("Fixable rules should export a `meta.fixable` property.");
                         }
+                        if (problem.suggestions && !(rule.meta && rule.meta.hasSuggestions === true)) {
+                            throw new Error("Rules with suggestions should set the `meta.hasSuggestions` property to `true`.");
```
nzakas, line 926:
```suggestion
                            throw new Error("Rules with suggestions must set the `meta.hasSuggestions` property to `true`.");
```

lib/linter/linter.js
```diff
@@ -922,6 +922,12 @@ function runRules(sourceCode, configuredRules, ruleMapper, parserOptions, parser
...
+                        }
+                        if (rule.meta && rule.meta.docs && typeof rule.meta.docs.suggestion !== "undefined") {
```
nzakas, line 928: I don’t think we need this as this property is already ignored. The check on line 925 covers the most important case.

tests/lib/rule-tester/rule-tester.js
```diff
@@ -1813,6 +1813,17 @@ describe("RuleTester", () => {
...
+
+        it("should throw an error if a rule that doesn't have `meta.hasSuggestions` enabled produces suggestions", () => {
+            assert.throws(() => {
+                ruleTester.run("suggestions-missing-hasSuggestions-property", require("../../fixtures/testers/rule-tester/suggestions").withoutHasSuggestionsProperty, {
+                    valid: [],
+                    invalid: [
+                        { code: "var foo = bar;", output: "5", errors: 1 }
+                    ]
+                });
+            }, "Rules with suggestions should set the `meta.hasSuggestions` property to `true`.");
```
nzakas, line 1825:
```suggestion
            }, "Rules with suggestions must set the `meta.hasSuggestions` property to `true`.");
```

nzakas (changes requested):
This looks really good. Just a couple small suggestions. Nice work.

lib/linter/linter.js
bmish: I'm worried that if we don't enforce the absence of the old, obsolete property, then we'll end up with the confusing and messy situation where some rules have both the new property (`meta.hasSuggestions`) and the old property (`meta.docs.suggestion`). This will happen because plugin developers performing the migration to ESLint 8 may only see the exception `Rules with suggestions should set the meta.hasSuggestions property to true` and not realize or forget that they're also supposed to cleanup/remove the old property. Furthermore, if some rules still have the old property, then some people/tooling may still try to rely on it.

Disallowing the old property also fits with the spirit of many of the [ESLint 6](https://eslint.org/docs/user-guide/migrating-to-6.0.0#rule-config-validating) and [ESLint 7](https://eslint.org/docs/user-guide/migrating-to-7.0.0#rule-tester-strict) changes to ensure that configurations/tests are validated more strictly and do not contain unknown properties. There's a separate question of whether we should apply stricter to the `meta` object in general (I'm in favor of stricter validation wherever possible and would like to pursue this if there's interest), but at the least I hope we disallow this obviously-outdated property.

nzakas: My point is that it doesn’t matter if people have meta.Docs.suggestions because that value is never used by us. If they are using it for some reason, they can keep using it. 

People can put whatever they want in meta.Docs as it’s just designed to hold documentation-type metadata. I can’t think of a reason why we would want to throw an error for anything in that section. People are free to use stuff in there as they see fit.

nzakas: Maybe we could change the wording a bit?

> Rules with suggestions must set the meta.hasSuggestions property to true. meta.docs.suggestions is ignored by ESLint.

That way it sounds less like you need to remove meta.docs.suggestions and more like, “hey, we aren’t using this value.”

nzakas (approved):
LGTM. @eslint/eslint-tsc looking for another set of eyes.

docs/developer-guide/working-with-rules.md
```diff
@@ -396,7 +399,8 @@ module.exports = {
             unnecessaryEscape: "Unnecessary escape character: \\{{character}}.",
             removeEscape: "Remove the `\\`. This maintains the current functionality.",
             escapeBackslash: "Replace the `\\` with `\\\\` to include the actual backslash character."
-        }
+        },
+        hasSuggestions: true
```
btmills, line 403: We include a note about the `fixable` property under the "Applying Fixes" headline on line 307. Can you add a similar `hasSuggestions` note under the "Providing Suggestions" header? Ideally someone can get everything they need to know from that section without having to read through to the examples in this sub-section.

btmills (commented):
Implementation and tests look good. Thank you for doing this! Nice touch giving a separate exception message with `meta.docs.suggestion`. Just one request for an extra note in the docs to match how we treat `fixable`, then LGTM.

docs/developer-guide/working-with-rules.md
bmish: Great idea, added.

btmills (approved):
LGTM, thanks!

mdjermanovic (commented):
I have a couple of suggestions about the documentation. Everything else looks good!

mdjermanovic (approved):
LGTM, thanks!

PR #14578, opened 2021-05-08
Docs: add more examples for `no-sequences` rule

snitin315:
#### Prerequisites checklist

[X] Documentation update
[ ] Add autofixing to a rule
[ ] Add a CLI option
[ ] Add something to the core
[ ] Other, please explain:

Add more examples.

Fixes https://github.com/eslint/eslint/issues/14572

Nope

### DIFF
--- a/docs/rules/no-sequences.md
+++ b/docs/rules/no-sequences.md
@@ -39,8 +39,6 @@ switch (val = foo(), val) {}
 while (val = foo(), val < 42);
 
 with (doSomething(), val) {}
-
-const foo = (val) => (console.log('bar'), val);
 ```
 
 Examples of **correct** code for this rule:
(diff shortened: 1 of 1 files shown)


aladdin-add (approved):
LGTM, thanks!

mdjermanovic (commented):
The added examples LGTM, but since we already had a similar example with arrow function (right above the new examples) I think the idea in https://github.com/eslint/eslint/issues/14572#issuecomment-834897231 was to add a note. @nzakas ?

nzakas:
Yes, I think we should add a note that explicitly states this exception case as it’s different from anywhere else in the rule.

nzakas:

snitin315:
Yes, thanks for the ping. I will push an update today.

snitin315:
Done ✅

nzakas (commented):
I was more thinking we should have a specific paragraph in the page saying something like:

### Note about arrow function bodies

If an arrow function body is a statement rather than a block, and that statement contains a sequence, you need to use double parentheses around the statement to indicate that the sequence is intentional.

(include examples here)

nzakas (approved):
Exactly what I was looking for. Thank you!

docs/rules/no-sequences.md
```diff
@@ -63,8 +61,32 @@ switch ((val = foo(), val)) {}
...
+
+If an arrow function body is a statement rather than a block, and that statement contains a sequence, you need to use double parentheses around the statement to indicate that the sequence is intentional.
+
+Examples of **incorrect** code for arrow functions:
+
+```js
+/*eslint no-sequences: "error"*/
+const foo = (val) => (console.log('bar'), val);
+
+const foo = () => ((bar = 123), 10));
```
fisker, line 76: These is an extra `)` in this example.

snitin315: Good catch, thanks. I will send a fix.

PR #14609, opened 2021-05-21
Chore: Remove needless filter call

Connormiha:
#### Prerequisites checklist

[ ] Documentation update
[ ] Add autofixing to a rule
[ ] Add a CLI option
[ ] Add something to the core
[x] Other, please explain:

I removed needless `filter` call and moved checking to next `forEach`.

### DIFF
--- a/lib/rules/line-comment-position.js
+++ b/lib/rules/line-comment-position.js
@@ -88,7 +88,11 @@ module.exports = {
             Program() {
                 const comments = sourceCode.getAllComments();
 
-                comments.filter(token => token.type === "Line").forEach(node => {
+                comments.forEach(node => {
+                    if (node.type !== "Line") {
+                        return;
+                    }
+
                     if (applyDefaultIgnorePatterns && (defaultIgnoreRegExp.test(node.value) || fallThroughRegExp.test(node.value))) {
                         return;
                     }


ljharb:
To me this doesn't seem like an improvement. `comments` isn't going to have millions of items in it, so the perf of this is effectively identical.

Connormiha:
> To me this doesn't seem like an improvement. `comments` isn't going to have millions of items in it, so the perf of this is effectively identical.

Filter looks very strange at this place. Because after filter, was called `forEach` with filtering by many params(via if + return).

ljharb:
I code like that all the time. It makes the forEach - the side-effecting part - smaller, which makes the code more readable.

Connormiha:
But this code has already included another `if -> return` inside forEach:
```js
if (applyDefaultIgnorePatterns && (defaultIgnoreRegExp.test(node.value) || fallThroughRegExp.test(node.value))) {
   return;
}

if (ignorePattern && customIgnoreRegExp.test(node.value)) {
   return;
}
```

`filter` does not reduce side-effect part at this place.

Before: `filter` + `forEach`(with 2 `if -> return`)
After: `forEach`(with 3 `if -> return` )

ljharb:
That’s an argument for moving those conditions to the filter, not for moving another condition into the forEach.

nzakas:
Folks, what we have here is just a difference of opinion on how to structure such code. In these cases, we default to what is already present in the codebase as there isn’t a discernible benefit to making the change and it’s not worth a long discussion when this comes down to personal preference.

PR #14617, opened 2021-05-23
Breaking: Fixable disable directives (fixes #11815)

### LINKED ISSUE #11815: --report-unused-disable-directives should be autofixable
v5.16.0

I want to enable `--report-unused-disable-directives` on a large existing codebase, but it surfaces too many problems to be practical to fix manually. So I'd need to write a script. At that point, why not make the directive autofixable?

Run `eslint --fix --report-unused-disable-directives`

Yes

JoshuaKGoldberg:
#### Prerequisites checklist

[ ] Documentation update
[ ] Add autofixing to a rule
[ ] Add a CLI option
[x] Add something to the core
[ ] Other, please explain:

Implements https://github.com/eslint/rfcs/pull/78 as described.

~I'm not confident I got all the places to update docs about the addition to `fixTypes` and friends...~ Thanks!

~Would you like me to split out the added logic in `apply-disable-directives.js` into a new file and individually unit test it? I was dubious that re-searching through the source file for the previously retrieved disable directives is the right approach for performance, so I held off in the initial commit.~ I ended up storing a `parentComment` in `linter.js` > `createDisableDirectives`, to hold off on re-scanning.


lib/linter/apply-disable-directives.js
```diff
@@ -18,6 +18,76 @@ function compareLocations(itemA, itemB) {
...
 
+/**
+ * Returns whether a comment is alone on its line.
+ * @param {Comment} comment The comment node which will be deleted.
+ * @param {string} line Contents of the comment's line.
+ * @returns {boolean} Whether the comment is alone on its line.
+ */
+function commentIsAloneOnLine(comment, line) {
+    return comment.type === "Block" ? line.length === comment.value.length + "/**/".length : line.startsWith("//");
+}
```
mdjermanovic, line 29: This logic could be wrong in some cases:

```js
/* eslint semi:2 */

const code = `const x = 5
// comment in the code`; // eslint-disable-line semi

```

autofixed to:

```js
/* eslint semi:2 */

const code = `const x = 5

```

The logic for block comments might be also wrong if the comment is multiline and the line where it starts happen to have the same length as the comment.

lib/linter/apply-disable-directives.js
```diff
@@ -87,6 +157,10 @@ function applyDirectives(options) {
     const unusedDisableDirectives = options.directives
         .filter(directive => directive.type === "disable" && !usedDisableDirectives.has(directive))
         .map(directive => ({
+            fix: {
+                range: getDirectiveRemovalRange(directive, options.sourceCode),
+                text: ""
+            },
```
mdjermanovic, line 163: When we are removing individual rules, we could accidentally remove all rules but leave the directive (that will then disallow all rules):

```js
/* eslint no-undef: 2, no-proto: 2, no-iterator: 2 */

// eslint-disable-next-line no-proto,, no-iterator
undef_foo();
``` 

autofixed to:

```js
/* eslint no-undef: 2, no-proto: 2, no-iterator: 2 */

// eslint-disable-next-line 
undef_foo();

```

This doesn't happen with 1 comma (`no-proto, no-iterator`) because the ranges are overlapping in that comma (one fix wants to remove `no-proto,`, the other `, no-iterator`), so only one fix gets applies in the first pass, while the second pass removes the whole comment because at that point there is only one rule in the directive.

The `,,` example is an edge case, but relying on the overlapping of successive ranges seems error-prone. Can we instead replace the whole comment with a new version that doesn't contain the rule?

Fox example, instead of a fix on a smaller range:

`no-proto,` -> ` `

the error for `no-proto` would produce a fix on the range of the whole comment:

`// eslint-disable-next-line no-proto,, no-iterator` -> `// eslint-disable-next-line , no-iterator`

mdjermanovic (changes requested):

aladdin-add:
friendly ping @JoshuaKGoldberg

JoshuaKGoldberg:
Hey thanks for the reviews and the ping! I'm about halfway through resolving the feedback locally and hope to finish within this week :)

lib/linter/apply-disable-directives.js
```diff
@@ -18,6 +18,76 @@ function compareLocations(itemA, itemB) {
...
+    // If the comment has space between it and whatever else is on its line, collapse the space
+    if (commentStart !== 0 && ["\n", " "].includes(sourceCode.text[commentStart - 1])) {
+        if (comment.range[1] === sourceCode.text.length) {
+            return [comment.range[0] - 1, comment.range[1]];
+        }
+
+        if (sourceCode.text[comment.range[1]] === " ") {
+            return [comment.range[0], comment.range[1] + 1];
+        }
+    }
```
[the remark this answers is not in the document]
JoshuaKGoldberg: Great, it's a bit of a relief not to have to do all that logic! In the latest commits I've simplified it to always replace with a `" "`, to keep things simple.

JoshuaKGoldberg:
Status update: I've added a "grouping" flow so that the logic knows whether an entire comment should be removed rather than only some of the rule IDs within. I ran out of time today to update the existing unit tests and hope to get to that soon, after which I'll touch up some of the mentioned edge cases around commas and spaces.

lib/linter/apply-disable-directives.js
```diff
@@ -18,6 +18,121 @@ function compareLocations(itemA, itemB) {
...
+/**
+ * Parses details from directives to create output Problems.
+ * @param {Directive[]} allDirectives Unused directives to be removed.
+ * @param {SourceCode} sourceCode A SourceCode object for the given text.
+ * @returns {{ fix, position, ruleIds }} Details for later creation of output Problems.
+ */
+function processUnusedDisableDirectives(allDirectives, sourceCode) {
+    const directiveGroups = groupByLine(allDirectives);
+
+    return lodash.flatMap(directiveGroups).map(directives => {
```
aladdin-add, line 122: lodash has been removed from deps in https://github.com/eslint/eslint/pull/14287. can you just use the `Array.flatMap()`?

JoshuaKGoldberg: Exciting! `Array.flatMap()` is unavailable until support for Node 10 is dropped, but I can use the `flatMap` added in that PR once I merge in from `eslint:master`. 👍

JoshuaKGoldberg:
@mdjermanovic I _think_ I've addressed all the edge cases you pointed out? It no longer attempts to get spacing perfectly right, and instead always plops a `" "` in as the fix.

Also, apologies in advance for not writing unit tests for the new edge cases: if this new approach looks good to you, is there a particular way you'd like me to do that? I'm thinking of pulling out `processUnusedDisableDirectives` into its own file so I can unit test that directly, but I can't tell if that goes against the spirit of the existing ESLint tests setup.

TSMMark:
As an eslint user I'm really looking forward to this. I ran lint with `--report-unused-disable-directives` at work on our codebase and there were several hundred offenses. We're holding off on making those fixes manually, knowing this new feature is around the corner, and we'll soon be able to easily autofix everything for free! Thanks for your efforts @JoshuaKGoldberg and eslint core team!

Brian-Bartels:
+1 to this fix being a huge benefit to my work. I recently had to manually remove 130 disables and the whole time I was wishing this existed

lib/linter/apply-disable-directives.js
```diff
@@ -16,6 +16,139 @@ function compareLocations(itemA, itemB) {
...
+
+/**
+ * Creates removal details for a set of directives within the same comment.
+ * @param {Directive[]} directives Unused directives to be removed.
+ * @param {Token} commentToken The backing Comment token.
+ * @param {SourceCode} sourceCode A SourceCode object for the given text.
+ * @returns {{ fix, position, ruleIds }[]} Details for later creation of output Problems.
+ */
+function createIndividualDirectivesRemoval(directives, commentToken, sourceCode) {
+    const commentStart = sourceCode.lineStartIndices[commentToken.loc.start.line - 1] + commentToken.loc.start.column;
```
mdjermanovic, line 64:
```suggestion
    const commentStart = commentToken.range[0];
```

Nodes, tokens, and comments have a `range` property. We can use `commentToken.range[0]` instead of calculating it from `loc`.

Also, it looks like we don't have any tests where this function will be used?

lib/linter/apply-disable-directives.js
```diff
@@ -16,6 +16,139 @@ function compareLocations(itemA, itemB) {
...
+ * @param {Directive[]} directives Unused directives to be removed.
+ * @param {Token} commentToken The backing Comment token.
+ * @param {SourceCode} sourceCode A SourceCode object for the given text.
+ * @returns {{ fix, position, ruleIds }} Details for later creation of an output Problem.
+ */
+function createCommentRemoval(directives, commentToken, sourceCode) {
+    const range = [
+        sourceCode.lineStartIndices[commentToken.loc.start.line - 1] + commentToken.loc.start.column,
+        sourceCode.lineStartIndices[commentToken.loc.end.line - 1] + commentToken.loc.end.column
+    ];
```
mdjermanovic, line 97: Similar to previous, we already have this range as `commentToken.range`.

mdjermanovic (changes requested):

mdjermanovic (changes requested):

lib/linter/apply-disable-directives.js
```diff
@@ -16,6 +16,139 @@ function compareLocations(itemA, itemB) {
...
+                range: [
+                    commentStart + 2 + ruleStart,
+                    commentStart + 2 + ruleEnd + (commentContents[ruleEnd] === "," ? 1 : 0)
+                ],
+                text: ""
+            },
+            position: directive.unprocessedDirective
+        };
+    });
+}
```
mdjermanovic, line 84: This could still find a wrong index, for example:

```js
/* eslint no-undef: 2, no-undefined: 2 */

/* eslint-disable-next-line no-undefined, no-undef */
undefined;

```

will be autofixed to:

```js
/* eslint no-undef: 2, no-undefined: 2 */

/* eslint-disable-next-line ined,  */
undefined;

```

(first pass removed `no-undef` from `no-undefined`, the second removed `no-undef`).

Could we maybe use regular expressions to more precisely locate the rule? It could be something like this:

```suggestion
function createIndividualDirectivesRemoval(directives, commentToken, sourceCode) {
    const listOffset = /^\s*\S+\s+/u.exec(commentToken.value)[0].length;
    const listText = commentToken.value
        .slice(listOffset) // remove eslint-*
        .split(/\s-{2,}\s/u)[0] // remove -- directive comment
        .trimRight();
    const listStart = commentToken.range[0] + 2 + listOffset;

    return directives.map(directive => {
        const { ruleId } = directive;
        const match = new RegExp(String.raw`(?:^|,)\s*${ruleId}\s*(?:$|,)`, "u").exec(listText);
        const ruleOffset = match.index;
        const ruleEndOffset = ruleOffset + match[0].length;
        const ruleText = listText.slice(ruleOffset, ruleEndOffset);

        return {
            description: ruleId,
            fix: {
                range: [
                    listStart + ruleOffset + (ruleText.startsWith(",") && ruleText.endsWith(",") ? 1 : 0),
                    listStart + ruleEndOffset
                ],
                text: ""
            },
            position: directive.unprocessedDirective
        };
    });
}
```

mdjermanovic (changes requested):

mdjermanovic:
> I'm not confident I got all the places to update docs about the addition to `fixTypes` and friends

API docs are good (no need to update the CLIEngine section as it will be removed soon).

CLI docs: [here](https://github.com/eslint/eslint/blob/master/docs/user-guide/command-line-interface.md#options) (under "Fixing problems"), [here](https://github.com/eslint/eslint/blob/master/docs/user-guide/command-line-interface.md#--fix-type), and [here](https://github.com/eslint/eslint/blob/e837d811e9275e43ccc266aba10e9c0833e0376c/lib/options.js#L154) (this is printed with `eslint -h`).

lib/shared/types.js
```diff
@@ -125,7 +125,7 @@ module.exports = {};
  * @property {Record<string,string>} [messages] The messages the rule reports.
  * @property {string[]} [replacedBy] The IDs of the alternative rules.
  * @property {Array|Object} schema The option schema of the rule.
- * @property {"problem"|"suggestion"|"layout"} type The rule type.
+ * @property {"directive"|"problem"|"suggestion"|"layout"} type The rule type.
```
mdjermanovic, line 128: I think this is a typedef for `meta` properties in rule objects, so `rule.meta.type` cannot be `"directive"`.

lib/cli-engine/cli-engine.js
```diff
@@ -325,6 +325,23 @@ function getRule(ruleId, configArrays) {
...
+ * @returns {boolean} Whether the message should be fixed.
+ */
+function shouldMessageBeFixed(message, lastConfigArrays, fixTypes) {
+    if (!message.ruleId) {
+        return fixTypes.has("directive");
+    }
+
+    const rule = message.ruleId && getRule(message.ruleId, lastConfigArrays);
+
+    return rule && rule.meta && fixTypes.has(rule.meta.type);
```
mdjermanovic, line 342:
```suggestion
    return Boolean(rule && rule.meta && fixTypes.has(rule.meta.type));
```

mdjermanovic:
> Also, apologies in advance for not writing unit tests for the new edge cases: if this new approach looks good to you, is there a particular way you'd like me to do that? 

Sorry, I missed this.

The approach with grouping errors into one error when all rules from the comment are unused LGTM. I was thinking about something different, but this looks better :+1: 

> I'm thinking of pulling out `processUnusedDisableDirectives` into its own file so I can unit test that directly, but I can't tell if that goes against the spirit of the existing ESLint tests setup.

We'd prefer unit tests for modules (`apply-disable-directives.js`) over tests for internal functions like `processUnusedDisableDirectives`.

It would be also good to add some higher-level tests, in `tests/lib/linter/linter.js`. Also, if it's easier to write tests in `linter.js` than unit tests for `apply-disable-directives.js`, it's perfectly fine that most test cases end up in `linter.js`.

lib/linter/apply-disable-directives.js
JoshuaKGoldberg: Nice, that `range` property -> not needing `sourceCode.lineStartIndices` makes this a lot simpler! Thanks 😄

JoshuaKGoldberg: Awesome, that `range` property completely gets rid of the need to pass around `sourceCode`. Thanks!

tests/lib/linter/linter.js
```diff
@@ -3347,6 +3359,44 @@ var a = "test2";
...
+                            text: " "
+                        },
+                        severity: 1,
+                        nodeType: null
+                    }
+                ]
+            );
+        });
+
+        it("reports problems for partially unused eslint-disable comments (in config)", () => {
```
JoshuaKGoldberg, line 3373: Was there anything in particular that should be tested at this level? I put in some unit-level ones into `apply-disable-directives.js` around commas, spacing, and ordering of rule IDs.

btmills (commented):
I’m still working my way through tests, but here’s a few quick fixes.

docs/user-guide/command-line-interface.md
```diff
@@ -47,7 +47,7 @@ Specifying rules and plugins:
 Fixing problems:
   --fix                          Automatically fix problems
   --fix-dry-run                  Automatically fix problems without saving the changes to the file system
-  --fix-type Array               Specify the types of fixes to apply (problem, suggestion, layout)
+  --fix-type Array               Specify the types of fixes to apply (directive, problem, suggestion, layout)
```
mdjermanovic, line 50: We should also update [`--fix-type` section](https://github.com/eslint/eslint/blob/master/docs/user-guide/command-line-interface.md#--fix-type) in the same document.

mdjermanovic (changes requested):

lib/linter/apply-disable-directives.js
```diff
@@ -16,6 +16,123 @@ function compareLocations(itemA, itemB) {
...
+
+    return directives.map(directive => {
+        const { ruleId } = directive;
+        const match = new RegExp(String.raw`(?:^|,)\s*${ruleId}\s*(?:$|,)`, "u").exec(listText);
+        const ruleOffset = match.index;
+        const ruleEndOffset = ruleOffset + match[0].length;
+        const ruleText = listText.slice(ruleOffset, ruleEndOffset);
+
+        return {
+            description: ruleId,
```
mdjermanovic, line 62:
```suggestion
            description: `'${ruleId}'`,
```

nzakas:
@JoshuaKGoldberg just checking in to see if you're still working on this?

JoshuaKGoldberg:
Sorry, I am - I missed the notification for the most recent round of reviews. Thanks for the ping (again)! I'll have them fixed up by the end of this weekend.

tests/lib/linter/linter.js
JoshuaKGoldberg: Marking as resolved since there's been a review since the comment.

lib/linter/apply-disable-directives.js
```diff
@@ -82,17 +201,22 @@ function applyDirectives(options) {
...
+                ? `Unused eslint-disable directive (no problems were reported from ${description}).`
                 : "Unused eslint-disable directive (no problems were reported).",
-            line: directive.unprocessedDirective.line,
-            column: directive.unprocessedDirective.column,
+            line: position.line,
+            column: position.column,
             severity: options.reportUnusedDisableDirectives === "warn" ? 1 : 2,
-            nodeType: null
+            nodeType: null,
+            fix: options.disableFixes ? null : fix
```
mdjermanovic, line 219:
```suggestion
            ...options.disableFixes ? {} : { fix }
```

`null` isn't an allowed value of the `fix` property, the property should be omitted if no fix is available.

tests/lib/linter/apply-disable-directives.js
```diff
@@ -438,7 +697,8 @@ describe("apply-disable-directives", () => {
                         ruleId: null,
                         message: "Unused eslint-disable directive (no problems were reported).",
                         line: 1,
-                        column: 5,
+                        column: 1,
+                        fix: null,
```
mdjermanovic, line 701:
```suggestion
```

mdjermanovic (changes requested):
Looks great, I have only one suggestion about the `fix` property.

mdjermanovic (approved):
LGTM, thanks!

TSMMark:
Tested this branch on a medium sized codebase, and it seems to work beautifully. So far I've only noticed one minor issue:

I don't know if this is in-scope, or if there's another lint rule that will report/fix this, but I noticed when eslint-disable rules are removed, there's a possibility to end up with multiple adjacent spaces in the disable comment.

Example:

Screen Shot 2021-08-03 at 3 24 05 PM

Thanks again for your work on this

mdjermanovic:
@TSMMark thanks for testing, and glad it works well for you!

> I don't know if this is in-scope, or if there's another lint rule that will report/fix this, but I noticed when eslint-disable rules are removed, there's a possibility to end up with multiple adjacent spaces in the disable comment.
> 
> Example:
> 
> Screen Shot 2021-08-03 at 3 24 05 PM

It's always difficult to produce a fix with the right formatting. We're usually making a minimal change that fixes the reported problem, and then rely on other rules to fix the whitespace afterward. Unfortunately, in this case there is no such rule. Since this is a minor issue that doesn't affect the correctness of this feature, and the solution doesn't seem trivial, I wouldn't address it in this PR (which is already fairly complex). If you open a separate issue, someone might be able to tweak the fix in another PR.

PR #14622, opened 2021-05-25
Update: ecmaVersion defaults to 5, and allows "latest"

aladdin-add:
#### Prerequisites checklist

[ ] Documentation update
[ ] Add autofixing to a rule
[ ] Add a CLI option
[ x] Add something to the core
[ ] Other, please explain:

* set ecmaVersion defaults to 5, make it not depend on espree's defaults
* allow ecmaVersion => "latest"

all these changes should not affect 3rd-party parsers.

### DIFF
--- a/lib/linter/linter.js
+++ b/lib/linter/linter.js
@@ -37,6 +37,7 @@ const
 const debug = require("debug")("eslint:linter");
 const MAX_AUTOFIX_PASSES = 10;
 const DEFAULT_PARSER_NAME = "espree";
+const DEFAULT_ECMA_VERSION = 5;
 const commentParser = new ConfigCommentParser();
 const DEFAULT_ERROR_LOC = { start: { line: 1, column: 0 }, end: { line: 1, column: 1 } };
 
@@ -432,10 +433,20 @@ function getDirectiveComments(filename, ast, ruleMapper, warnInlineConfig) {
 
 /**
  * Normalize ECMAScript version from the initial config
- * @param  {number} ecmaVersion ECMAScript version from the initial config
+ * @param {Parser} parser The parser which uses this options.
+ * @param {number} ecmaVersion ECMAScript version from the initial config
  * @returns {number} normalized ECMAScript version
  */
-function normalizeEcmaVersion(ecmaVersion) {
+function normalizeEcmaVersion(parser, ecmaVersion) {
+    if (parser === espree) {
+        if (ecmaVersion === void 0) {
+            return DEFAULT_ECMA_VERSION;
+        }
+
+        if (ecmaVersion === "latest") {
+            return espree.latestEcmaVersion;
+        }
+    }
 
     /*
      * Calculate ECMAScript edition number from official year version starting with
@@ -521,12 +532,13 @@ function normalizeVerifyOptions(providedOptions, config) {
 
 /**
  * Combines the provided parserOptions with the options from environments
- * @param {string} parserName The parser name which uses this options.
+ * @param {Parser} parser The parser which uses this options.
  * @param {ParserOptions} providedOptions The provided 'parserOptions' key in a config
  * @param {Environment[]} enabledEnvironments The environments enabled in configuration and with inline comments
  * @returns {ParserOptions} Resulting parser options after merge
  */
-function resolveParserOptions(parserName, providedOptions, enabledEnvironments) {
+function resolveParserOptions(parser, providedOptions, enabledEnvironments) {
+
     const parserOptionsFromEnv = enabledEnvironments
         .filter(env => env.parserOptions)
         .reduce((parserOptions, env) => merge(parserOptions, env.parserOptions), {});
@@ -542,12 +554,7 @@ function resolveParserOptions(parserName, providedOptions, enabledEnvironments) 
         mergedParserOptions.ecmaFeatures = Object.assign({}, mergedParserOptions.ecmaFeatures, { globalReturn: false });
     }
 
-    /*
-     * TODO: @aladdin-add
-     * 1. for a 3rd-party parser, do not normalize parserOptions
-     * 2. for espree, no need to do this (espree will do it)
-     */
-    mergedParserOptions.ecmaVersion = normalizeEcmaVersion(mergedParserOptions.ecmaVersion);
+    mergedParserOptions.ecmaVersion = normalizeEcmaVersion(parser, mergedParserOptions.ecmaVersion);
 
     return mergedParserOptions;
 }
@@ -606,7 +613,7 @@ function getRuleOptions(ruleConfig) {
  */
 function analyzeScope(ast, parserOptions, visitorKeys) {
     const ecmaFeatures = parserOptions.ecmaFeatures || {};
-    const ecmaVersion = parserOptions.ecmaVersion || 5;
+    const ecmaVersion = parserOptions.ecmaVersion || DEFAULT_ECMA_VERSION;
 
     return eslintScope.analyze(ast, {
         ignoreEval: true,
@@ -1123,7 +1130,7 @@ class Linter {
             .map(envName => getEnv(slots, envName))
             .filter(env => env);
 
-        const parserOptions = resolveParserOptions(parserName, config.parserOptions || {}, enabledEnvs);
+        const parserOptions = resolveParserOptions(parser, config.parserOptions || {}, enabledEnvs);
         const configuredGlobals = resolveGlobals(config.globals || {}, enabledEnvs);
         const settings = config.settings || {};
 
--- a/tests/lib/linter/linter.js
+++ b/tests/lib/linter/linter.js
@@ -3493,6 +3493,22 @@ var a = "test2";
         });
 
         describe("ecmaVersion", () => {
+            it("the default ECMAScript version is 5", () => {
+                const messages = linter.verify("let x = 0;", {});
+
+                assert.strictEqual(messages.length, 1);
+            });
+
+            it("supports ECMAScript version 'latest'", () => {
+                const messages = linter.verify("let x = 5 ** 7;", {
+                    parserOptions: {
+                        ecmaVersion: "latest"
+                    }
+                });
+
+                assert.strictEqual(messages.length, 0);
+            });
+
             describe("it should properly parse let declaration when", () => {
                 it("the ECMAScript version number is 6", () => {
                     const messages = linter.verify("let x = 5;", {
--- a/tests/lib/linter/linter.js
+++ b/tests/lib/linter/linter.js
@@ -11,6 +11,7 @@
 
 const assert = require("chai").assert,
     sinon = require("sinon"),
+    espree = require("espree"),
     esprima = require("esprima"),
     testParsers = require("../../fixtures/parsers/linter-test-parsers");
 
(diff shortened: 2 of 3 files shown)


lib/linter/linter.js
```diff
@@ -542,12 +547,10 @@ function resolveParserOptions(parserName, providedOptions, enabledEnvironments)
...
     }
 
-    /*
-     * TODO: @aladdin-add
-     * 1. for a 3rd-party parser, do not normalize parserOptions
-     * 2. for espree, no need to do this (espree will do it)
-     */
-    mergedParserOptions.ecmaVersion = normalizeEcmaVersion(mergedParserOptions.ecmaVersion);
+    // for a 3rd-party parser, do not normalize parserOptions
+    if (parserName === DEFAULT_PARSER_NAME || parserName === void 0) {
```
mdjermanovic, line 551: This could be a problem for rules that expect normalized `ecmaVersion`, like `comma-dangle`:

https://github.com/eslint/eslint/blob/85a2725b1fade5538e727102d9701ccb503e54d4/lib/rules/comma-dangle.js#L53

nzakas: Agree. We should always normalize ecmaVersion.

aladdin-add: I have not thought it was used in rules, will update later.

nzakas (commented):
We also need to update the docs to mention that “latest” is allowed.

lib/linter/linter.js
aladdin-add: it could be another problem to normalize a 3rd-party parser `ecmaVersion`. as @mdjermanovic 
 mentioned [here](https://github.com/eslint/eslint/pull/14622#discussion_r640580610) .

the problem is eslint cannot tell a 3rd-party parser's default, and "latest".

aladdin-add: after some thinking, I would suggest not to change it.

If it's an issue, it has existed in the previous releases. The change is not a bugfix, let's make another change to fix it.

mdjermanovic: Can we retain the exact same behavior as it is now for third-party parsers, and make changes only if it's default parser?

aladdin-add: pushed a commit 99604d4e1617bda1622426c8efd30aa397de8c81 to revert "always normalize ecmaVersion".

mdjermanovic: In my opinion we should always do the following, regardless of the parser, because that's the current behavior:

https://github.com/eslint/eslint/blob/ec28b5a2bdc69f34ce29d670f5e84d2446774a00/lib/linter/linter.js#L444

But the new normalizations (`undefined` -> `5`, `"latest"` -> `espree.latestEcmaVersion`) should be done only when `espree` is used.

aladdin-add: 👍 good catch! I missed it somehow. will update

aladdin-add: should be addressed now. thanks!

nzakas (approved):
LGTM

lib/linter/linter.js
```diff
@@ -432,10 +433,19 @@ function getDirectiveComments(filename, ast, ruleMapper, warnInlineConfig) {
...
 /**
  * Normalize ECMAScript version from the initial config
- * @param  {number} ecmaVersion ECMAScript version from the initial config
+ * @param {number} ecmaVersion ECMAScript version from the initial config
+ * @param {string} parserName the parser's name
  * @returns {number} normalized ECMAScript version
  */
-function normalizeEcmaVersion(ecmaVersion) {
+function normalizeEcmaVersion(ecmaVersion, parserName) {
+    if (parserName === DEFAULT_PARSER_NAME || parserName === void 0) {
```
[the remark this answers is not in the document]
nzakas: Setting the parser name should definitely not affect the ecmaVersion even after upgrading Espree. If that’s the case in this PR, we should fix that.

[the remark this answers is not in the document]
nzakas: I just took a deeper look at this. There’s no reason to check the parser name in this function. We just remove the check and I think things will work correctly.

nzakas (changes requested):

lib/linter/linter.js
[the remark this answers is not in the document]
aladdin-add: it will be a breaking change to the normalize a 3rd parser (e.g. @typescript-eslint/parser) ecmaVersion: `undefined` => 5. 

as @mdjermanovic mentioned here: https://github.com/eslint/eslint/pull/14622#discussion_r640580610

tests/lib/linter/linter.js
```diff
@@ -3445,6 +3445,16 @@ var a = "test2";
...
+
+                it("the ECMAScript version is 'latest'", () => {
+                    const messages = linter.verify("let x = 5 ** 7;", {
+                        parserOptions: {
+                            ecmaVersion: "latest"
+                        }
+                    });
+
+                    assert.strictEqual(messages.length, 0);
+                });
```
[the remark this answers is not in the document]
aladdin-add: resolved in 033f08092c109ff0c7ac71d78adaee18de558cf6

lib/linter/linter.js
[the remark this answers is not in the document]
ljharb: it's set in `parserOptions` though.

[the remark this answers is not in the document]
nzakas: @aladdin-add we want this to be non-breaking.

@ljharb thats true and that value is still used in a bunch of places outside of the parser for legacy reasons. The new config system fixes this but for eslintrc we need to live with this weirdness.

nzakas:
There seems to be some confusion about how ecmaVersion currently works and it seems like some folks are assuming everyone understands already. Can someone outline the various scenarios, including when `parser` is and isn’t present, and what the result is currently? I think that will help provide the missing context to this PR.

aladdin-add:
well, I will try. the current ecmaVersion normalization is simple:
* undefined => undefined
* 2015 => 6


That's say, eslint depends on the parser's defaults option. In the PR, it changed the normalization to:
* undefined => 5
* latest => espree.latestEcmaVersion
* 2015 => 6

it will be breaking, when all the following conditions are met:
* not set ecmaVersion
* using a 3rd-party parser, but it's default ecmaVersion is not 5.

ljharb:
@aladdin-add and also, when that parser actually pays attention to the ecmaVersion?

mdjermanovic:
ESLint currently doesn't automatically set `parserOptions.ecmaVersion` if `parserOptions.ecmaVersion` wasn't specified in user's configuration.

Espree's default is `ecmaVersion: 5`. Other parsers have different defaults.

Espree is about to change its default `ecmaVersion`, but we don't want that change to affect ESLint users.

Ideally, ESLint should automatically set `parserOptions.ecmaVersion` to `5` if `parserOptions.ecmaVersion` wasn't specified in user's configuration, whenever ESLint is used with Espree but not with another parser (as it would be a breaking change if that parser had a different default).

The problem is: it seems difficult for `Linter` to know that the parser is Espree when user specifies `parser: "espree"`, because in that case `@eslint/eslintrc` loads parser and sends the parser object and absolute path to `Linter`.

nzakas:
Thanks everyone, so I think what we want is this:

| **ecmaVersion** | **parser** | **result** |
|--------------------|------------|-----------|
| undefined | undefined | `parser: espree`, `ecmaVersion: 5` |
| undefined | `"espree"` | `parser: espree`, `ecmaVersion: 5` |
| undefined | `"other-parser"` | `parser: "other-parser"`, `ecmaVersion: undefined` |
| 2022 | undefined | `parser: "espree"`, `ecmaVersion: 2022` |
| 2022 | `"espree"` | `parser: "espree"`, `ecmaVersion: 2022` |
| 2022 | `"other-parser"` | `parser: "other-parser"`, `ecmaVersion: 2022` |

> The problem is: it seems difficult for Linter to know that the parser is Espree when user specifies parser: "espree", because in that case @eslint/eslintrc loads parser and sends the parser object and absolute path to Linter.

Can we compare that object to `require("espree")` to determine this? Or compare the absolute path to `require.resolve("espree")`?

aladdin-add:
> Can we compare that object to require("espree") to determine this? Or compare the absolute path to require.resolve("espree")?

I just thought a very edge case: someone might want to use a different version of espree.
`parser: "./node_modules/espree"`

```txt
node_modules
├── espree
└── eslint
    └── node_modules
        └── espree
```

But I think it's fine to not normalize it - espree was used as a 3rd-party parser in this case.

will update as @nzakas suggested (if no objections).

mdjermanovic:
> Can we compare that object to `require("espree")` to determine this? Or compare the absolute path to `require.resolve("espree")`?

Sounds good to me!

> I just thought a very edge case: someone might want to use a different version of espree.
> `parser: "./node_modules/espree"`
> 
> ```
> node_modules
> ├── espree
> └── eslint
>     └── node_modules
>         └── espree
> ```
> 
> But I think it's fine to not normalize it - espree was used as a 3rd-party parser in this case.

I agree. If someone explicitly installs another version of espree, then it's like they're using a third-party parser with its own defaults.

Though this might also happen unintentionally when another dependency has espree as a dependency, and the other version of espree ends up in the top `node_modules`. But that setup seems already broken. Looks like it generally isn't a good idea to explicitly set `"parser": "espree"` in eslint config unless the user really wants a different version, because `"parser": "espree"` doesn't guarantee that the espree version specified in eslint's package.json will be used.

nzakas:
I had a realization: `Linter` is not supposed to use Node.js-specific functionality because it is also used in the browser, so I don't think using `require()` make much sense.

So, at this point, I think the path of least resistance is to leave the Espree default `ecmaVersion` as 5. We can still support "latest" in ESLint but don't have to worry about the additional complexity we're running into now. With flat config, we will have an opportunity to choose a different default `ecmaVersion` for ESLint and it will be much easier to special-case Espree because `Linter` will have a direct reference to it.

Thoughts>

aladdin-add:
> I had a realization: Linter is not supposed to use Node.js-specific functionality because it is also used in the browser, so I don't think using require() make much sense.

do you mean using `require.resolve()`? But  in the PR, I didn't use it, I just used `parser === espree` to check it, so it seems not an issue.

nzakas (approved):
Ah ok. LGTM. 👍

tests/lib/linter/linter.js
```diff
@@ -3493,6 +3493,22 @@ var a = "test2";
...
+            it("supports ECMAScript version 'latest'", () => {
+                const messages = linter.verify("let x = 5 ** 7;", {
+                    parserOptions: {
+                        ecmaVersion: "latest"
+                    }
+                });
+
+                assert.strictEqual(messages.length, 0);
+            });
+
```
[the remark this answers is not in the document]
mdjermanovic: And maybe one test to confirm that we'll pass normalized version (not `"latest"`) to `eslint-scope` (first choice from https://github.com/eslint/espree/issues/495#issuecomment-847142140):

```js
let blockScope = null;

linter.defineRule("block-scope", context => ({
    BlockStatement() {
        blockScope = context.getScope();
    }
}));

linter.verify("{}", {
    rules: {
        "block-scope": 2
    },
    parserOptions: {
        ecmaVersion: "latest"
    }
});

assert.strictEqual(blockScope.type, "block");
```

`blockScope.type` would be `"global"` if `eslint-scope` gets `"latest"`, < 6, or `undefined` as `ecmaVersion`.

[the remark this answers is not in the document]
aladdin-add: addressed in https://github.com/eslint/eslint/pull/14622/commits/2b82badb2a6b9acbc329b44d0c07279cd5c3b372.

mdjermanovic (approved):
LGTM, thanks!

nzakas:
I just realized that this can be included before v8, do we want to merge now?

mdjermanovic:
> I just realized that this can be included before v8, do we want to merge now?

Yes, I also think this doesn't depend on https://github.com/eslint/espree/commit/b068cea907ef59fe5681f6acedd30b932d496e0a. We can merge this for an ESLint v7 minor version.

aladdin-add:
yes, and https://github.com/eslint/espree/commit/b068cea907ef59fe5681f6acedd30b932d496e0a depends on this one(set ecmaVersion defaults to 5) :)

PR #14779, opened 2021-07-08
Update: improving the no-console error message (fixes #14782)

### LINKED ISSUE #14782: Forgetting "use strict" results in the entire file being underlined
### Overview

With the rule configuration `strict: ['warn', 'global']`

If you forget to include the `'use strict'` directive at top of the file, it highlights the entire file, rather than just the first line.

image

Expected behavior would be to only show the warning on the first line.

strict

Yes

new option

With the rule configuration `strict: ['warn', 'global']`

```js
let x = 1;
let y = 2;
let z = 3;
```

The entire file is underlined. Making it difficult to see/count other issues.

It will only underline the first line.

Yes.

paulsmithkc:
#### Prerequisites checklist

[ ] Documentation update
[ ] Add autofixing to a rule
[ ] Add a CLI option
[ ] Add something to the core
[ ] Other, please explain:

no-console

no

new default behavior

```js
console.log('we got here');
```

It shows the error message: `Unexpected console statement.`

It shows the error message: `console statements are used for debugging, and should generally be stripped before the code is pushed to production.`

Expanded the error message to include the recommendations found in the docs. 

Should the error message specify that this is a `console.log()` or `console.error()` statement?

### DIFF
--- a/lib/rules/no-console.js
+++ b/lib/rules/no-console.js
@@ -44,7 +44,7 @@ module.exports = {
         ],
 
         messages: {
-            unexpected: "Unexpected console statement."
+            unexpected: "console statements are used for debugging, and should generally be stripped before the code is pushed to production."
         }
     },
 


nzakas:
Thanks for the suggestion. We generally like to keep the messages ESLint produces as short as possible, otherwise, the formatting doesn't look great on a console. As is, I think your suggested messages is too specific to the browser use case, as `console.log()` can be used in Node.js as part of a CLI program.

So, while we appreciate the suggestion, I don't think this is a worthwhile change.

paulsmithkc:
@nzakas Considering that eslint has some idea of the context of the code. Would it be possible to provide different errors for browser / node environments?

I know for myself I use the "debug" package over console.log in Node.js. So some way for me to indicate this in the error message would be helpful.

Does this maybe need an option to configure the error message based on context? Similar to no-restricted-globals?

PR #14798, opened 2021-07-14
Update: add end location to report in `consistent-return` (refs #12334)

snitin315:
#### Prerequisites checklist

[X] Other, please explain:

refs #12334

This PR adds `loc.end` to reports in `consistent-return`

Just a small change from `loc.start` to `loc`

No

### DIFF
--- a/tests/lib/rules/consistent-return.js
+++ b/tests/lib/rules/consistent-return.js
@@ -52,7 +52,11 @@ ruleTester.run("consistent-return", rule, {
                 {
                     messageId: "missingReturnValue",
                     data: { name: "Function 'foo'" },
-                    type: "ReturnStatement"
+                    type: "ReturnStatement",
+                    line: 1,
+                    column: 46,
+                    endLine: 1,
+                    endColumn: 53
                 }
             ]
         },
@@ -63,7 +67,11 @@ ruleTester.run("consistent-return", rule, {
                 {
                     messageId: "missingReturnValue",
                     data: { name: "Arrow function" },
-                    type: "ReturnStatement"
+                    type: "ReturnStatement",
+                    line: 1,
+                    column: 47,
+                    endLine: 1,
+                    endColumn: 54
                 }
             ]
         },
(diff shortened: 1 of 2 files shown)


tests/lib/rules/consistent-return.js
```diff
@@ -255,6 +316,7 @@ ruleTester.run("consistent-return", rule, {
                     messageId: "missingReturn",
                     data: { name: "program" },
                     type: "Program",
+                    line: 1,
                     column: 1
```
btmills, line 320: Every other invalid test case asserts end location information, but we don't have an end location for `Program` reports. Do you think it's worth adding these to demonstrate to future readers that this omission is intentional?

```suggestion
                    line: 1,
                    column: 1,
                    endLine: void 0,
                    endColumn: void 0 
```

tests/lib/rules/consistent-return.js
```diff
@@ -207,7 +256,10 @@ ruleTester.run("consistent-return", rule, {
                     messageId: "missingReturn",
                     data: { name: "function" },
                     type: "FunctionExpression",
-                    column: 3
+                    line: 1,
+                    column: 3,
+                    endLine: 1,
+                    endColumn: 37
```
btmills, line 262: Reporting the entire function including its body seems inconsistent here. Everywhere else we have missing returns, the rule selects the function name, `function` keyword, or `=>` token. I'd expect this case to stop at the end of the `function` keyword like the rest.

btmills (changes requested):
Thanks for taking on the rest of these end location reports! I see one test where the end location doesn't match my expectation. The start location matches what I'd expect, so this looks like the rule needs a new special case where it was previously able to use a fallback when it reported just start locations.

nzakas (approved):
LGTM

tests/lib/rules/consistent-return.js
snitin315: Hmm, should stop at tge end of `function`, I will update 👍

tests/lib/rules/consistent-return.js
snitin315: Makes sense, I will add these.

btmills (approved):
LGTM. Thanks @snitin315!

PR #14809, opened 2021-07-18
Chore: assertions on reporting loc in `unicode-bom` (refs #12334)

snitin315:
#### Prerequisites checklist

[X] Other, please explain:

Refs #12334

This PR adds `loc.end` to reports in `unicode-bom` rule.

no

### DIFF
--- a/lib/rules/unicode-bom.js
+++ b/lib/rules/unicode-bom.js
@@ -43,8 +43,13 @@ module.exports = {
             Program: function checkUnicodeBOM(node) {
 
                 const sourceCode = context.getSourceCode(),
-                    location = { column: 0, line: 1 },
+                    firstToken = sourceCode.getFirstToken(node),
+                    location = { start: { column: 0, line: 1 } },
                     requireBOM = context.options[0] || "never";
+
+                if (firstToken) {
+                    location.end = firstToken.loc.end;
+                }
 
                 if (!sourceCode.hasBOM && (requireBOM === "always")) {
                     context.report({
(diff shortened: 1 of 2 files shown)


lib/rules/unicode-bom.js
```diff
@@ -43,9 +43,14 @@ module.exports = {
...
 
                 const sourceCode = context.getSourceCode(),
-                    location = { column: 0, line: 1 },
+                    firstToken = sourceCode.getFirstToken(node),
+                    location = { start: { column: 0, line: 1 } },
                     requireBOM = context.options[0] || "never";
 
+                if (firstToken) {
+                    location.end = firstToken.loc.end;
+                }
```
mdjermanovic, line 52: I'm not sure if it makes sense to highlight a potentially large range because of the missing/required BOM character.

For example, this is how it would look like on the rule source file itself:

image

Thoughts about highlighting only the first character, as proposed in https://github.com/eslint/eslint/issues/12334#issuecomment-655085810?

mdjermanovic: Also, this is basically the same situation (reporting the beginning of the file) as [one case](https://github.com/eslint/eslint/blob/6641d88e17d952a8e51df5e0d3882a842d4c3f35/tests/lib/rules/consistent-return.js#L312) in the `consistent-return` rule where we've decided to keep reporting only the start loc?

snitin315: I think it's better with keeping only the start loc (same as `consistent-return`)

snitin315: I updated the test cases to assert on reporting location.

nzakas (approved):
LGTM

mdjermanovic (approved):
LGTM, thanks!

PR #14886, opened 2021-08-06
Chore: Code path analysis for class fields (fixes #14343)

### LINKED ISSUE #14343: support Class Fields

nzakas:
#### Prerequisites checklist

[ ] Documentation update
[ ] Add autofixing to a rule
[ ] Add a CLI option
[ ] Add something to the core
[x] Other, please explain:

I added tests for code path analysis to show that analysis is working correctly for class fields. No other changes were necessary as I believe the standard traversal gives the code path analyzer all of the information it needs.

I started by adding a test that showed how we currently traverse an object literal with a property that is initialized by a conditional statement because this is the closest analog to how code path analysis should work with class fields.

I then added a test showing that a class field initialized with a conditional statement creates the correct code path, as it basically matches what happened in the object literal case.

Am I missing anything?

### DIFF
--- a/tests/fixtures/code-path-analysis/class-field--conditional.js
+++ b/tests/fixtures/code-path-analysis/class-field--conditional.js
@@ -1,0 +1,20 @@
+/*expected
+initial->s1_1->s1_2->s1_4;
+s1_1->s1_3->s1_4->final;
+*/
+
+class x { a = b ? c : d }
+
+/*DOT
+digraph {
+    node[shape=box,style="rounded,filled",fillcolor=white];
+    initial[label="",shape=circle,style=filled,fillcolor=black,width=0.25,height=0.25];
+    final[label="",shape=doublecircle,style=filled,fillcolor=black,width=0.25,height=0.25];
+    s1_1[label="Program:enter\nClassDeclaration:enter\nIdentifier (x)\nClassBody:enter\nPropertyDefinition:enter\nIdentifier (a)\nConditionalExpression:enter\nIdentifier (b)"];
+    s1_2[label="Identifier (c)"];
+    s1_4[label="ConditionalExpression:exit\nPropertyDefinition:exit\nClassBody:exit\nClassDeclaration:exit\nProgram:exit"];
+    s1_3[label="Identifier (d)"];
+    initial->s1_1->s1_2->s1_4;
+    s1_1->s1_3->s1_4->final;
+}
+*/
--- a/tests/fixtures/code-path-analysis/object-literal--conditional.js
+++ b/tests/fixtures/code-path-analysis/object-literal--conditional.js
@@ -1,0 +1,20 @@
+/*expected
+initial->s1_1->s1_2->s1_4;
+s1_1->s1_3->s1_4->final;
+*/
+
+x = { a: b ? c : d }
+
+/*DOT
+digraph {
+    node[shape=box,style="rounded,filled",fillcolor=white];
+    initial[label="",shape=circle,style=filled,fillcolor=black,width=0.25,height=0.25];
+    final[label="",shape=doublecircle,style=filled,fillcolor=black,width=0.25,height=0.25];
+    s1_1[label="Program:enter\nExpressionStatement:enter\nAssignmentExpression:enter\nIdentifier (x)\nObjectExpression:enter\nProperty:enter\nIdentifier (a)\nConditionalExpression:enter\nIdentifier
+    (b)"];
+    s1_2[label="Identifier (c)"];
+    s1_4[label="ConditionalExpression:exit\nProperty:exit\nObjectExpression:exit\nAssignmentExpression:exit\nExpressionStatement:exit\nProgram:exit"];
+    s1_3[label="Identifier (d)"];
+    initial->s1_1->s1_2->s1_4;
+    s1_1->s1_3->s1_4->final;
+}*/
--- a/tests/lib/linter/code-path-analysis/code-path-analyzer.js
+++ b/tests/lib/linter/code-path-analysis/code-path-analyzer.js
@@ -561,11 +561,11 @@ describe("CodePathAnalyzer", () => {
                     }
                 }));
                 const messages = linter.verify(source, {
-                    parserOptions: { ecmaVersion: 2021 },
+                    parserOptions: { ecmaVersion: 2022 },
                     rules: { test: 2 }
                 });
 
-                assert.strictEqual(messages.length, 0);
+                assert.strictEqual(messages.length, 0, "Unexpected linting error in code.");
                 assert.strictEqual(actual.length, expected.length, "a count of code paths is wrong.");
 
                 for (let i = 0; i < actual.length; ++i) {


tests/fixtures/code-path-analysis/class-field--conditional.js
```diff
@@ -0,0 +1,20 @@
...
+    initial[label="",shape=circle,style=filled,fillcolor=black,width=0.25,height=0.25];
+    final[label="",shape=doublecircle,style=filled,fillcolor=black,width=0.25,height=0.25];
+    s1_1[label="Program:enter\nClassDeclaration:enter\nIdentifier (x)\nClassBody:enter\nPropertyDefinition:enter\nIdentifier (a)\nConditionalExpression:enter\nIdentifier (b)"];
+    s1_2[label="Identifier (c)"];
+    s1_4[label="ConditionalExpression:exit\nPropertyDefinition:exit\nClassBody:exit\nClassDeclaration:exit\nProgram:exit"];
+    s1_3[label="Identifier (d)"];
+    initial->s1_1->s1_2->s1_4;
+    s1_1->s1_3->s1_4->final;
+}
+*/
```
mdjermanovic, line 20: In object literals, the initialization of properties happens in the same code path. Class field initializers are not executed in the same code path, but while running the constructor, for each instance.

```js
const { log } = console;

log(1);

var obj = { 
  foo: log(2) 
};

log(3);

class C {
  foo = log(4);
}

log(5);

new C();

// logs: 1 2 3 5 4
```

Code path analysis should already work well _inside_ class field initializers. The update would be to create  a separate [code path](https://github.com/eslint/eslint/blob/master/docs/developer-guide/code-path-analysis.md#codepath) on entering the initializer, i.e. to treat them as functions.

nzakas: Ah ok. Do you have any idea of how that could be implemented? I just don’t understand enough about how code path analysis to works to even have a theory.

nzakas:
So I took a look at the code path analysis for a similar pattern:

```js
function Foo() {
    this.a = b ? c : d
};

new Foo()
```

The result is:

```
initial->s2_1->s2_2->s2_4;
s2_1->s2_3->s2_4->final;
initial->s1_1->final;
```

Which results in the following dot graphs:

graphviz(1)
graphviz

Similar to class fields, `this.a` isn't initialized until `new Foo()` is called, but we don't tie those two calls together currently. If that's the case, it seems like the same would apply to class fields in an example like this:

```js
class Foo {
    a = b ? c : d;
}

new Foo();
```

So do we have a bug in our existing code path analysis? Or are the class fields different in some way that I'm not seeing?

(Again, I really don't understand how any of this works, so please feel free to point out how I'm misunderstanding this.)

mdjermanovic:
A separate code path (which we currently create only for functions, incl. methods) indicates that the code inside isn't executed in the same flow as the surrounding code, but we don't track function calls to connect code paths. That info isn't provided by our code path analysis, so `new Foo()` isn't tied to the constructor of `class Foo`. Our code path analysis only splits the code into unconnected code paths (Program and functions) and then splits code paths into code path segments (which do provide detailed info about the flow in each code path).

```js
A;

class B extends C {
    [D] = E + F;
}

G;
```

A -> B -> C -> D -> G is one execution flow, they're all evaluated one after another. AST traversal also goes through `E + F` between D and G, but it isn't evaluated between D and G, so it's not part of the same flow.

The specification defines initializers as functions https://tc39.es/ecma262/#sec-runtime-semantics-classfielddefinitionevaluation.

For example:

```js
class Foo {
   x = 0;  
   y = this.x + 1;
}
```

is roughly the same as:

```js
function xInitializer() {
    return 0;
}

function yInitializer() {
    return this.x + 1;
}

class Foo {
    constructor() {
        Object.defineProperty(this, "x",  { value: xInitializer.call(this) /*, ... */ });
        Object.defineProperty(this, "y",  { value: yInitializer.call(this) /*, ... */ });
    }
}
```

I don't think our code path analysis can provide too advanced info about the execution flow of class fields, like the one proposed in https://github.com/eslint/eslint/issues/14317, but it seems reasonable to at least create a separate code path for each class field initializer, i.e. to treat `E + F` as a function.

nzakas:
Gotcha. That makes sense, thanks for explaining. I'm still not sure I know how to do that, but I'll dig in a bit.

nzakas:
A note for myself: The challenge here is to treat the right side of `PropertyDefinition` as a `FunctionExpression` even though it isn't one. Otherwise, normal traversal applies.

nzakas:
So here is what I have so far. For this code:

```js
class Foo {
    [a] = b;
}
```

graphviz(3)

graphviz(4)

I know this isn't correct, but how close does it look?

mdjermanovic:
This definitely looks like the right direction!

It could be actually already correct although it doesn't appear so. Maybe just `debug-helpers` don't handle this example well so it doesn't look good by looking at the graphs (note that `FunctionDeclaration` in https://github.com/eslint/eslint/pull/14886#issuecomment-894582711 also appears both in its own code path and the outer code path; `Identifier` doesn't have child nodes, [this code](https://github.com/eslint/eslint/blob/8c350660e61284c41a5cc1a5955c858db53c516b/lib/linter/code-path-analysis/debug-helpers.js#L77-L78) optimizes the output in such cases, but now when a lone `Identifier` represents a separate code path it produces a weird output with `:enter` in the outer code path).

nzakas:
Okay, I think I've got it! I still don't 100% understand everything that's going on, but this commit seems to work the way (I think) we're expecting.

graphviz(5)
graphviz(6)

mdjermanovic (approved):
LGTM! This works as I would expect it to.

The only thing that might be confusing and difficult to handle in rules is that for code such as  `class Foo { a = () => {} }` the `onCodePathStart` is raised twice with the same `node` argument (the arrow function), so we might want to add `CodePath#type` property which could be used to make a distinction between the initializer code path and function's own code path.

nzakas:
> so we might want to add CodePath#type property which could be used to make a distinction between the initializer code path and function's own code path.

Can you explain this a bit more? I'm not understanding what you're suggesting here.

mdjermanovic:
My suggestion is that we add `type` property to `CodePath` objects. Value would be one of  `"program"`, `"function"`, `"class-field-initializer"`.

Rules are using code path analysis through [events](https://eslint.org/docs/developer-guide/code-path-analysis#events), like `onCodePathStart(codePath, node)`.

Before introducing class fields, it was clear what each code path refers to, by `node.type`. If `node.type === "Program"`, then the code path represents top-level code, otherwise `node` is a function node and the code path represents that function's code.

With class fields initializers, which don't have a dedicated wrapper node in the AST, `onCodePathStart` will be ambiguous when `PropertyDefinition.value` is a function node.

```js
class Foo { 
    a = () => {};
}
```

For the above code, `onCodePathStart(codePath, node)` will be called two times with the same ArrowFunctionExpression node, first time representing the initializer's code path, second time representing function's own code path, but the rule can't easily know which is which. For that reason, `codePath.type` feels like helpful additional info for rules.

nzakas:
Ah gotcha, thanks for explaining. That sounds like a good way to differentiate the types of code paths. I’ll work on that.

PR #14933, opened 2021-08-15
Build: Force prerelease peer dep for Node 16 in CI

btmills:
#### Prerequisites checklist

[ ] Documentation update
[ ] Add autofixing to a rule
[ ] Add a CLI option
[ ] Add something to the core
[x] Other, please explain: Fix CI peerdep issues during prereleases

Several dev dependencies have a peer dependency on the local ESLint. Prereleases don't satisfy regular version range constraints, so the new peer dependency resolution algorithm in Node 16 rejects `v8.0.0-beta.0` as a valid peer dependency. Until we're done with prereleases, we can use the `--force` flag to force npm to accept that the prerelease satisfies the peer dependency.

Is there a better way to do this? Yarn has resolutions that might allow fixing this, but I don't see any equivalent for npm.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -42,6 +42,7 @@ jobs:
         node-version: ${{ matrix.node }}
     - name: Install Packages
       run: npm install
+      # TODO(btmills): Remove Node 16 --force branching after releasing v8.0.0 final.
       if: ${{ !startswith(matrix.node, '16') }}
     - name: Install Packages
       run: npm install --force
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -42,6 +42,10 @@ jobs:
         node-version: ${{ matrix.node }}
     - name: Install Packages
       run: npm install
+      if: ${{ !startswith(matrix.node, '16') }}
+    - name: Install Packages
+      run: npm install --force
```
snitin315, line 47: I think its ok to use `--force`, also we can add a comment here `remove it after v8.0.0 stable release`

nzakas (approved):
LGTM. I’d suggest opening an issue to remind us to revert this change when the final release happens.

.github/workflows/ci.yml
btmills: Done in 972238c

PR #14993, opened 2021-08-29
Update: add configurable comment behavior to no-empty

saadflow:
#### Prerequisites checklist

[ ] Documentation update
[ ] Add autofixing to a rule
[ ] Add a CLI option
[ ] Add something to the core
[ ] Other, please explain:

`no-empty`

more

A new option, whose default is the current behavior.
Currently, an empty block containing only comments doesn't produce any warning. 
This is fine most of the time, but if for example you want to make sure that catch block don't contain `// todo: handle error`, it can be useful to make this behavior configurable.
This change adds an option to do just that, and if not provided, its default is to allow empty blocks with comments, so that the current behavior remains the default.

```js
for(const foo of bar){
 // Keep me in the loop
}
```
```js
try {
   throwError()
} catch {
  // TODO: Handle this error.
}
```

Currently it doesn't produce any error / warning.

Now, it will do the same, unless the `allowEmptyWithComments` option is set to false, in which case it will produce an error / warning

I implemented the new option in `lib/rules/no-empty.js`
I wrote additional tests in `tests/lib/rules/no-empty.js`
I updated the documentation in `docs/rules/no-empty.md`

Their well being !

### DIFF
--- a/docs/rules/no-empty.md
+++ b/docs/rules/no-empty.md
@@ -61,6 +61,8 @@ This rule has an object option for exceptions:
 
 * `"allowEmptyCatch": true` allows empty `catch` clauses (that is, which do not contain a comment)
 
+* `"allowEmptyWithComments": false` makes a block who only contains comments invalid.
+
 ### allowEmptyCatch
 
 Examples of additional **correct** code for this rule with the `{ "allowEmptyCatch": true }` option:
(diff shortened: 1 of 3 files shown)


nzakas:
We did not accept https://github.com/eslint/eslint/issues/15006 so closing.

PR #15032, opened 2021-09-06
Docs: LintMessage.line and column are possibly undefined

btmills:
#### Prerequisites checklist

[x] Documentation update
[ ] Add autofixing to a rule
[ ] Add a CLI option
[ ] Add something to the core
[ ] Other, please explain:

In `lib/linter/linter.js`, the `parse()` function's exception handler translates the exception's message to a `LintMessage`. It also passes through the exception's `lineNumber` and `column`, but the exception is not guaranteed to have those properties. In that case, `LintMessage.line` and `column` can be `undefined`.

https://github.com/eslint/eslint/blob/f966fe6286b6f668812f5155b79d4ee2a8b584b3/lib/linter/linter.js#L693-L711

In eslint/eslint-plugin-markdown#191, the processor would crash when attempting to map messages without a line.

Normally widening a type like this would be a breaking change, but since this is only updating the docs to reflect reality, I think this can be a semver-patch change.

For the JSDoc type, I used `{number|undefined} line` instead of `{number} [line]` because `line` and `column` are always present but may have the value `undefined`.

### DIFF
--- a/docs/developer-guide/nodejs-api.md
+++ b/docs/developer-guide/nodejs-api.md
@@ -376,9 +376,9 @@ The `LintMessage` value is the information of each linting error. The `messages`
   `true` if this is a fatal error unrelated to a rule, like a parsing error.
 * `message` (`string`)<br>
   The error message.
-* `line` (`number`)<br>
+* `line` (`number | undefined`)<br>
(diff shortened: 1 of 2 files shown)


nzakas (approved):
LGTM

PR #15140, opened 2021-10-07
feat: report only lines that exceed the limit in max-lines-per-function

The-x-Theorist:
#### Prerequisites checklist

[ ] Documentation update
[ ] Add autofixing to a rule
[ ] Add a CLI option
[ ] Add something to the core
[ ] Other, please explain:

Updated the max-lines-per-function rule to report only the count of those lines which exceed the specified max-lines limit, instead of reporting total lines from the start of the function to the end.

nope

### DIFF
--- a/lib/rules/max-lines-per-function.js
+++ b/lib/rules/max-lines-per-function.js
@@ -205,10 +205,7 @@ module.exports = {
                         line: node.loc.start.line + maxLines + (comments + blankLines),
                         column: 0
                     },
-                    end: {
-                        line: node.loc.end.line,
-                        column: node.loc.end.column
-                    }
+                    end: node.loc.end
                 };
 
                 context.report({
--- a/lib/rules/max-lines-per-function.js
+++ b/lib/rules/max-lines-per-function.js
@@ -79,7 +79,7 @@ module.exports = {
             OPTIONS_OR_INTEGER_SCHEMA
         ],
         messages: {
-            exceed: "{{name}} has exceeded the limit of lines allowed by ({{linesExceed}}). Maximum allowed is {{maxLines}}."
+            exceed: "{{name}} has exceeded the limit of lines allowed by {{linesExceed}}. Maximum allowed number of lines per function is {{maxLines}}."
         }
     },
 
--- a/lib/rules/max-lines-per-function.js
+++ b/lib/rules/max-lines-per-function.js
@@ -85,7 +85,10 @@ module.exports = {
 
     create(context) {
         const sourceCode = context.getSourceCode();
-        const lines = sourceCode.lines;
+        const lines = sourceCode.lines.map((text, i) => ({
+            lineNumber: i + 1,
+            text
+        }));
 
         const option = context.options[0];
         let maxLines = 50;
@@ -171,7 +174,7 @@ module.exports = {
             let lineCount = 0;
 
             for (let i = node.loc.start.line - 1; i < node.loc.end.line; ++i) {
-                const line = lines[i];
+                const line = lines[i].text;
 
                 if (skipComments) {
                     if (commentLineNumbers.has(i + 1) && isFullLineComment(line, i + 1, commentLineNumbers.get(i + 1))) {
@@ -191,9 +194,21 @@ module.exports = {
             if (lineCount > maxLines) {
                 const name = upperCaseFirst(astUtils.getFunctionNameWithKind(funcNode));
                 const linesExceed = lineCount - maxLines;
+                const srcCodeLinesLength = [...sourceCode.lines].pop().length;
+
+                const loc = {
+                    start: {
+                        line: lines[maxLines].lineNumber,
+                        column: 0
+                    },
+                    end: {
+                        line: sourceCode.lines.length,
+                        column: srcCodeLinesLength
+                    }
+                };
 
                 context.report({
-                    node,
+                    loc,
                     messageId: "exceed",
                     data: { name, linesExceed, maxLines }
                 });
--- a/lib/rules/max-lines-per-function.js
+++ b/lib/rules/max-lines-per-function.js
@@ -79,7 +79,7 @@ module.exports = {
             OPTIONS_OR_INTEGER_SCHEMA
         ],
         messages: {
-            exceed: "{{name}} has too many lines ({{lineCount}}). Maximum allowed is {{maxLines}}."
+            exceed: "{{name}} has exceeded the line limit by ({{linesExceed}}). Maximum allowed is {{maxLines}}."
         }
     },
 
@@ -190,11 +190,12 @@ module.exports = {
 
             if (lineCount > maxLines) {
                 const name = upperCaseFirst(astUtils.getFunctionNameWithKind(funcNode));
+                const linesExceed = lineCount - maxLines;
 
                 context.report({
                     node,
                     messageId: "exceed",
-                    data: { name, lineCount, maxLines }
+                    data: { name, linesExceed, maxLines }
                 });
             }
         }
--- a/lib/rules/max-lines-per-function.js
+++ b/lib/rules/max-lines-per-function.js
@@ -172,21 +172,26 @@ module.exports = {
                 return;
             }
             let lineCount = 0;
-            let linesTotal = 0;
+            let comments = 0;
+            let blankLines = 0;
 
             for (let i = node.loc.start.line - 1; i < node.loc.end.line; ++i) {
                 const line = lines[i].text;
 
-                linesTotal++;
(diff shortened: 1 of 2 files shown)


lib/rules/max-lines-per-function.js
```diff
@@ -190,11 +193,24 @@ module.exports = {
...
             if (lineCount > maxLines) {
                 const name = upperCaseFirst(astUtils.getFunctionNameWithKind(funcNode));
+                const linesExceed = lineCount - maxLines;
+                const srcCodeLinesLength = [...sourceCode.lines].pop().length;
+
+                const loc = {
+                    start: {
+                        line: lines[maxLines].lineNumber,
+                        column: 0
+                    },
```
mdjermanovic, line 203: This would always be the same line in the file. For example, if the max is 5, it will report line 6.

image

We should find the first line that goes above the max in the function (line 13 in this screenshot).

lib/rules/max-lines-per-function.js
```diff
@@ -190,11 +193,24 @@ module.exports = {
...
+                    end: {
+                        line: sourceCode.lines.length,
+                        column: srcCodeLinesLength
+                    }
```
mdjermanovic, line 207:
```suggestion
                    end: node.loc.end
```

Similar to the previous, `loc.end` would always be the last position in the file, while we probably want the last position in the function.

lib/rules/max-lines-per-function.js
```diff
@@ -190,11 +193,24 @@ module.exports = {
...
+                };
 
                 context.report({
-                    node,
```
mdjermanovic, line 195: We can keep `node` here, and just add `loc`. If both `node` and `loc` are present, the reported location will be `loc`.

https://eslint.org/docs/developer-guide/working-with-rules#contextreport

`node` will provide additional info (e.g., `nodeType`) for lint messages.

mdjermanovic (changes requested):

The-x-Theorist (commented):
I will improve the changes in next commit.

lib/rules/max-lines-per-function.js
```diff
@@ -190,11 +193,21 @@ module.exports = {
 
             if (lineCount > maxLines) {
                 const name = upperCaseFirst(astUtils.getFunctionNameWithKind(funcNode));
+                const linesExceed = lineCount - maxLines;
+
+                const loc = {
+                    start: {
+                        line: maxLines + 1,
+                        column: 0
+                    },
```
mdjermanovic, line 202: Sorry if my [previous comment](https://github.com/eslint/eslint/pull/15140#discussion_r740432372) wasn't clear enough.

`line: maxLines + 1` is incorrect, because it reports a line that isn't related to the function:

image

In the example from the screenshot, reported line should be 14, because that's the 6th line of the function:

image

It would be helpful to add some tests cases where the function doesn't start on the first line.

Also, if the rule is configured to ignore blank lines, then a blank line shouldn't be marked as the start of lines that are above the limit, because it doesn't contribute to the line count. We should find the first line that goes above the limit of those that are counted. Same for comments. An example of how this is implemented is [max-lines](https://github.com/eslint/eslint/blob/c415c041912a3abbf106cc5713bdcf4ef42590ac/lib/rules/max-lines.js) rule.

mdjermanovic (changes requested):

lib/rules/max-lines-per-function.js
The-x-Theorist: It should report only those start and end lines in loc property which start after the maxLine or It should just report the start and end line in loc property from the start and end of the whole function not related to maxLine?

mdjermanovic: It should highlight extra lines in the function. Those are lines that start after the last allowed line in the function.

For example, if the function has 7 lines, and the max is 5, it should highlight the 6th and the 7th line of the function. So, `loc.start.line` should be the 6th line of the function (not the 6th line of the whole file). This is how it should look like:

image

```js
loc: {
    start: {
        line: 14,
        column: 0
    },
    end: {
        line: 15,
        column: 1
    }
}
```

It's a bit more complicated if the rule is configured to skip blank lines and/or comments:

> Also, if the rule is configured to ignore blank lines, then a blank line shouldn't be marked as the start of lines that are above the limit, because it doesn't contribute to the line count. We should find the first line that goes above the limit of those that are counted. Same for comments. An example of how this is implemented is [max-lines](https://github.com/eslint/eslint/blob/c415c041912a3abbf106cc5713bdcf4ef42590ac/lib/rules/max-lines.js) rule.

The-x-Theorist:
Is this change works good with skipBlankLines and skipComments too? @mdjermanovic

mdjermanovic:
> Is this change works good with skipBlankLines and skipComments too?

It doesn't look good:

image

It should highlight lines starting from the first that is above the max, of those that are not skipped:

image

The-x-Theorist:
@mdjermanovic Now, I have changed the way skipComments and skipBlankLines was handled for loc property.

lib/rules/max-lines-per-function.js
```diff
@@ -79,13 +79,16 @@ module.exports = {
...
         }
     },
 
     create(context) {
         const sourceCode = context.getSourceCode();
-        const lines = sourceCode.lines;
+        const lines = sourceCode.lines.map((text, i) => ({
+            lineNumber: i + 1,
+            text
+        }));
```
mdjermanovic, line 91: Is this array necessary? It doesn't seem that it provides much value for the rest of the code, so can we just use `sourceCode.lines`?

nzakas:
@The-x-Theorist are you still working on this?

The-x-Theorist:
@nzakas No, I have completed the work it's ready to be merged.

lib/rules/max-lines-per-function.js
```diff
@@ -79,13 +79,16 @@ module.exports = {
             OPTIONS_OR_INTEGER_SCHEMA
         ],
         messages: {
-            exceed: "{{name}} has too many lines ({{lineCount}}). Maximum allowed is {{maxLines}}."
+            exceed: "{{name}} has exceeded the limit of lines allowed by ({{linesExceed}}). Maximum allowed is {{maxLines}}."
```
mdjermanovic, line 82:
```suggestion
            exceed: "{{name}} has exceeded the limit of lines allowed by {{linesExceed}}. Maximum allowed number of lines per function is {{maxLines}}."
```

lib/rules/max-lines-per-function.js
mdjermanovic, line 91:
```suggestion
        const lines = sourceCode.lines;
```

lib/rules/max-lines-per-function.js
```diff
@@ -169,18 +172,26 @@ module.exports = {
                 return;
             }
             let lineCount = 0;
+            let comments = 0;
+            let blankLines = 0;
 
             for (let i = node.loc.start.line - 1; i < node.loc.end.line; ++i) {
-                const line = lines[i];
+                const line = lines[i].text;
```
mdjermanovic, line 179:
```suggestion
                const line = lines[i];
```

lib/rules/max-lines-per-function.js
```diff
@@ -190,11 +201,24 @@ module.exports = {
 
             if (lineCount > maxLines) {
                 const name = upperCaseFirst(astUtils.getFunctionNameWithKind(funcNode));
+                const linesExceed = lineCount - maxLines;
+
+                const loc = {
+                    start: {
+                        line: (lines[node.loc.start.line].lineNumber - 1) + maxLines + (comments + blankLines),
```
mdjermanovic, line 208:
```suggestion
                        line: node.loc.start.line + maxLines + (comments + blankLines),
```

lib/rules/max-lines-per-function.js
```diff
@@ -190,11 +201,24 @@ module.exports = {
...
+                        column: 0
+                    },
+                    end: {
+                        line: node.loc.end.line,
+                        column: node.loc.end.column
+                    }
```
mdjermanovic, line 214:
```suggestion
                    end: node.loc.end
```

mdjermanovic (changes requested):

mdjermanovic (approved):
LGTM, thanks!

nzakas (approved):
LGTM. Thanks!

ljharb:
This seems like a bit of a breaking change.

It's causing my `eslint-disable` comments - that have always been positioned at the top of the function - to be considered unused, and instead the random line that happens to break the threshold is reported on.

Issue #14588, opened 2021-05-14
reconsider the defaults of eslint config


aladdin-add:
things has changed since eslint was created in 2013. It is time to reconsider the defaults of eslint config.
I would suggest the changes:

1. sourceType: "script" => "module"

* native esm has been supported by latest modern browsers and node.js
* a large number of devs are writing esm, and using something like babel/ts to support older browser/node.js versions.

2. ecmaVersion: 5 => "latest"
we have often got issues like "unexpected token...." just because not config `ecmaVersion`. e.g. https://github.com/eslint/eslint/issues/11643
the change should be good for users & maintainers.

3. --ext: ".js" => ".js,.mjs,.cjs"

it is a breaking change, but it's very easy to restore the old behaviour
```js
{
  "parserOptions": {"ecmaVersion": 5, "sourceType": "script",}
}
```

if we agreed to the change, document & `eslint --init` should be updated to reflect the change, too.

nzakas:
We can consider this when we move to flat config. Changing this for eslintrc is likely to break a lot of existing users’ configurations.

btmills:
I like @nzakas' idea. Flat config gives us a chance to update defaults in a way that likely _reduces_ migration difficulty.

aladdin-add:
agreed. I will update the OP later.

nzakas:
I’ll handle this as I’m working on the new config.

nzakas:
Fixed in #15185

Issue #4728, opened 2015-12-17
CommonJS recommendation + `global` variable


isiahmeadows:
In http://eslint.org/docs/user-guide/configuring#specifying-environments, it specifies to use `commonjs`, not `node` for Browserify/Webpack/etc. The most common CommonJS implementations have a `global` variable, even though it's not mandated in the CommonJS spec, mostly following Node's lead. In browsers, `global` is an alias for `window`.

Could this recommendation be either fixed/clarified on the website or patched in ESLint proper?

(cross-posted in https://github.com/eslint/eslint.github.io/issues/169, as it applies there, also)

eslintbot:
@isiahmeadows Thanks for the issue! If you're reporting a bug, please be sure to include:

1. The version of ESLint you are using (run `eslint -v`)
2. What you did (the source code and ESLint configuration)
3. The actual ESLint output complete with numbers
4. What you expected to happen instead

Requesting a new rule? Please see [Proposing a New Rule](http://eslint.org/docs/developer-guide/contributing/new-rules) for instructions.

nzakas:
I'm not sure what you're requesting here. The `commonjs` environment was created specifically for this purpose, are you saying it's missing `global`?

isiahmeadows:
Yes

nzakas:
So the solution would be to add `global` to the CommonJS environment rather than change the documentation. We can fix that.

michaelficarra:
No, we shouldn't do this. It's easy to add a global named global. It should not be implicitly added for everyone using CommonJS.

nzakas:
@michaelficarra why not? If every CommonJS environment provides `global`, why should we force people to always add it?

michaelficarra:
A CommonJS environment is not required to have it. One may drop it, or a new one without it may become popular. Given that System.global will likely soon be a thing, I don't think this is too unlikely.

nzakas:
The fact is that commonjs environments that support `global` currently dominate the landscape. Given the state of current environments, it seems more likely that any new ones would have to support `global` for compatibility (even of that just means `global = System.global`).

I'd rather help people today rather than worry about a future where something becomes popular and doesn't support it.

nzakas:
The `globals` module has been updated. This only needs the package to be updated for the fix.

PR #15185, opened 2021-10-18
feat: Flat config support in Linter (refs #13481)

nzakas:
#### Prerequisites checklist

[ ] Documentation update
[ ] Add autofixing to a rule
[ ] Add a CLI option
[x] Add something to the core
[ ] Other, please explain:

Updated `Linter` to accept `FlatConfigArray` as an argument. To do so, I added a new `configType` option that must be set to `"flat"`. When set to `"flat"`, the `defineRule`, `defineRules`, and `defineParser` methods all throw an error because there is no good way to mimic this functionality with flat config. (People can instead define a plugin directly in the config.) Ultimately, we'll end up removing these methods.

I've only added enough tests to verify each of the initial paths for flat config. In a future pull request I'll work on converting the existing tests to work with flat config. I just don't want to wait too long before merging these changes in order to avoid conflicts.

Note: These changes are a prerequisite for the `ESLint` class changes.

Task list:

- [x] Linting both with and without processors
- [x] Environments do not work when `configType` is `"flat"`
- [x] Globals added via comments are correctly added to the global scope
- [x] Autofixing
- [ ] ~~`context.parserPath` compatibility~~
- [x] `context.languageOptions` doesn't exist
- [x]  `languageOptions.ecmaVersion` is year-based (2015 instead of 6)
- [x] `languageOptions.ecmaVersion` translates to `parserOptions.ecmaVersion`
- [x] `languageOptions.ecmaVersion` adds globals for the appropriate version
- [x]  `languageOptions.ecmaVersion: 3` only has ES3 globals
- [x]  `languageOptions.sourceType: "commonjs"` adds Node.js globals
- [x]  `languageOptions.sourceType: "commonjs"` switches parser options to parse global `return`
- [x] Convert all eslintrc tests into flat config tests
- [x] Update defaults (#14588)
- [x] All tests passing

Is there any danger in merging this?


lib/config/default-config.js
```diff
@@ -46,7 +46,16 @@ exports.defaultConfig = [
             ".git/**"
         ],
         languageOptions: {
-            parser: "@/espree"
+            ecmaVersion: "latest",
+            sourceType: "module",
+            parser: "@/espree",
+            parserOptions: {}
+        }
```
nzakas, line 53: Updating the defaults as briefly discussed in #14588. I think it makes sense to assume people will be using ESM and always want the latest `ecmaVersion`.

lib/config/default-config.js
```diff
@@ -46,7 +46,16 @@ exports.defaultConfig = [
...
+    },
+    {
+        files: ["**/*.cjs"],
+        languageOptions: {
+            sourceType: "commonjs"
```
nzakas, line 58: For `.cjs` files, I think it makes sense to opt them into "commonjs" sourceType.

nzakas:
Latest update fixes #15284 as well.

nzakas:
@mdjermanovic @btmills this is ready for review. This PR is just focused on getting `Linter` working with flat config. I have purposely not documented these changes, in accordance with our "if it's not documented, use at your own risk" mantra. If we can agree on and implement https://github.com/eslint/espree/issues/519, at that point I believe it will be useful to document everything as a "Developer Preview" of sorts to let people start playing with `Linter` to get some early feedback on what is working/not working.

lib/linter/linter.js
```diff
@@ -566,6 +633,30 @@ function resolveParserOptions(parser, providedOptions, enabledEnvironments) {
...
+function createLanguageOptions({ globals: configuredGlobals, parser, parserOptions }) {
+
+    const {
+        ecmaVersion,
+        sourceType
+    } = parserOptions;
+
+    return {
+        globals: configuredGlobals,
+        ecmaVersion: normalizeEcmaVersionForLanguageOptions(ecmaVersion),
```
mdjermanovic, line 653: If `parserOptions.ecmaVersion` wasn't specified in the configuration, this will set `languageOptions.ecmaVersion` to `NaN`.

If `parserOptions.ecmaVersion` is a string, for example `"next"`, this will set `languageOptions.ecmaVersion` to `"next2019"`.

conf/globals.js
```diff
@@ -0,0 +1,143 @@
...
+
+//-----------------------------------------------------------------------------
+// Globals
+//-----------------------------------------------------------------------------
+
+const commonjs = {
+    require: false,
+    exports: false,
+    module: false
+};
```
mdjermanovic, line 16: There are two differences between this and [globals.commonjs](https://github.com/sindresorhus/globals/blob/332404584ae6721500cfcddb034a3fae26618666/globals.json#L1220). Is it intentional?

* `global` is missing here.
* `exports` is readonly here.

`global` was added to `globals.commonjs` per https://github.com/eslint/eslint/issues/4728.

Assignment to `exports` is usually a bug, but there's [this style](https://github.com/mysticatea/eslint-plugin-node/blob/master/docs/rules/exports-style.md#allowbatchassign).

lib/shared/types.js
```diff
@@ -25,6 +25,14 @@ module.exports = {};
  * @property {"script"|"module"} [sourceType] The source code type.
  */
 
+/**
+ * @typedef {Object} LanguageOptions
+ * @property {number|"latest"} [ecmaVersion] The ECMAScript version (or revision number).
+ * @property {"script"|"module"} [sourceType] The source code type.
+ * @property {string|Object} [parser] The parser to use.
+ * @property {Object} [parserOptions] The parser options to use.
```
mdjermanovic, line 33: This typedef is missing `LanguageOptions.globals` property.

lib/linter/linter.js
```diff
@@ -1270,15 +1391,43 @@ class Linter {
...
+                 * Because of how Webpack packages up the files, we can't
+                 * compare directly to `FlatConfigArray` using `instanceof`
+                 * because it's not the same `FlatConfigArray` as in the tests.
+                 * So, we work around it by assuming an array is, in fact, a
+                 * `FlatConfigArray` if it has a `getConfig()` method.
+                 */
+                let configArray = config;
+
+                if (!Array.isArray(config) || typeof config.getConfig !== "function") {
+                    configArray = new FlatConfigArray(config || {});
```
mdjermanovic, line 1414:
```suggestion
                    configArray = new FlatConfigArray(config);
```

At this point, `config` is certainly truthy?

mdjermanovic:
Should we maybe create a lint warning if `/* eslint-env ... */` comment is found in `configType: "flat"` mode?

conf/globals.js
nzakas: Not intentional. I'll update it. Thanks.

lib/linter/linter.js
nzakas: Good point.

lib/shared/types.js
nzakas: Nice catch. Added.

lib/linter/linter.js
nzakas: Wow, amazing catch!

nzakas:
> Should we maybe create a lint warning if /* eslint-env ... */ comment is found in configType: "flat" mode?

yes, I’ll do that in a future PR

mdjermanovic:
In `configType: "flat"` mode, rule configuration directives are still validated using eslintrc's ConfigValidator:

https://github.com/eslint/eslint/blob/2536a6b0ebe9370c6d9cfbd8ddfb1b4601bcbf93/lib/linter/linter.js#L445

Is this intentional for the first version?

nzakas:
> In configType: "flat" mode, rule configuration directives are still validated using eslintrc's ConfigValidator:

Yes. It's a little hard to pull this apart while we support two modes.

lib/linter/linter.js
```diff
@@ -1289,6 +1438,215 @@ class Linter {
...
+        // evaluate arguments
+        if (typeof textOrSourceCode === "string") {
+            slots.lastSourceCode = null;
+            text = textOrSourceCode;
+        } else {
+            slots.lastSourceCode = textOrSourceCode;
+            text = textOrSourceCode.text;
+        }
+
+        const languageOptions = config.languageOptions;
```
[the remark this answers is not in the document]
nzakas: That is weird. I'll ensure the behavior is consistent with the eslintrc approach.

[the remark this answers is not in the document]
nzakas: We should probably have this be an error in both Espree and eslint-scope, but that would be a breaking change, so maybe we can think about it later.

nzakas:
Okay, I think I've got everything

nzakas:
I just upgraded Espree and eslint-scope to the versions that support `sourceType:commonjs` and eliminated a bunch of the code that managed that conversion. Should be all set now.

lib/linter/linter.js
```diff
@@ -460,6 +497,38 @@ function normalizeEcmaVersion(parser, ecmaVersion) {
...
+            }
+    }
+
+    /*
+     * We default to the latest supported ecmaVersion for everything else.
+     * Remember, this is for languageOptions.ecmaVersion, which sets the version
+     * that is used for a number of processes inside of ESLint. It's normally
+     * safe to assume people want the latest unless otherwise specified.
+     */
+    return espree.latestEcmaVersion + 2009;
```
mdjermanovic, line 529: Since `languageOptions.ecmaVersion` is an ESLint option, and it is used not only for parsing, shouldn't ESLint maintain its own `latestEcmaVersion` instead of using `espree.latestEcmaVersion`?

For example, we could release support for ES2023 in Espree's minor version, which would be then in range of ESLint versions that don't support ES2023. In that case, `languageOptions: { ecmaVersion: "latest" }` in user's configuration will result in `2023`, but as there is no `es2023` environment there will be no ES globals at all.

lib/linter/linter.js
```diff
@@ -460,6 +497,38 @@ function normalizeEcmaVersion(parser, ecmaVersion) {
...
+ * @param {number} ecmaVersion ECMAScript version from the initial config
+ * @returns {number} normalized ECMAScript version
+ */
+function normalizeEcmaVersionForLanguageOptions(ecmaVersion) {
+
+    switch (ecmaVersion) {
+        case 3:
+            return 3;
+
+        // void 0 = now ecmaVersion specified so use the default
```
mdjermanovic, line 512:
```suggestion
        // void 0 = no ecmaVersion specified so use the default
```

typo

lib/linter/linter.js
nzakas: That might be a good idea. Can you open a separate issue to discuss? I don’t want to couple that decision with this PR.

mdjermanovic:
> Is there any danger in merging this?

I could find only 3 differences with eslintrc configs before and after this change.

-----

1. If configured `parserOptions` doesn't have `sourceType` property, what will be passed to parsers as `sourceType` property in the options object that is the second argument of `parse()`/`parseForESLint()`.
 
* Before this change:  there will be no `sourceType` property in the options object.
* After this change: `sourceType: undefined`.

 In theory, this could affect parsing, depending on how the handling of options is implemented in a custom parser, although it does not seem likely. Thoughts? I checked espree, [`@typescript-eslint/parser`](https://github.com/typescript-eslint/typescript-eslint/blob/853d799428a061d9bf6a2e74b01dc49a1e4f3134/packages/parser/src/parser.ts#L92-L94) and [`@babel/eslint-parser`](https://github.com/babel/babel/blob/23fec0c8c99a70b5739124dc2087a24f291b021e/eslint/babel-eslint-parser/src/configuration.cjs#L2-L9), and It doesn't seem that this change will affect the behavior.

----

2. Possible `ecmaVersion` values passed to `eslint-scope`.

* Before this change: 3, 5, 6, 7, 8, 9, 10, 11, 12, 13.
* After this change: 3, 5, 2015, 2016, 2017, 2018, 2019, 2020, 2021, 2022.

`eslint-scope` expects `ecmaVersion` normalized to 6 ... 13, not 2015 ... 2022. Nonetheless, this change doesn't really affect the scope analysis, because the only conditions related to `ecmaVersion` in `eslint-scope` are `ecmaVersion >= 5` and  `ecmaVersion >= 6`, so they'll work the same with 2015 ... 2022 instead of 6 ... 13. However, there are no tests in `eslint-scope` that would make sure that 2015 and others work the same as 6, all tests are based on the 6 ... 13 normalization.

----

3. `ecmaVersion` passed to `eslint-scope` when a custom parser is used and `ecmaVersion` is not a number.

* Before this change: 6 (per https://github.com/eslint/eslint/issues/15256).
* After this change: 2022.

Similar to the previous point, this doesn't really change how `eslint-scope` works.

nzakas:
1. I don’t think it’s worth addressing this now. While it’s technically a change, it shouldn’t affect anything. And if it does affect something, it will be easier to solve when we have a real world use case. 

2. Can you open an issue on eslint-scope for that? We can certainly add tests. 

3. This is intentional. The new default ecmaVersion with flat config is latest.

mdjermanovic:
> 2\. Can you open an issue on eslint-scope for that? We can certainly add tests.

https://github.com/eslint/eslint-scope/issues/82

lib/linter/linter.js
mdjermanovic: >  Can you open a separate issue to discuss?

https://github.com/eslint/eslint/issues/15366

lib/linter/linter.js
```diff
@@ -1324,6 +1691,46 @@ class Linter {
...
+                { ...options, filename, disableFixes, postprocess, preprocess },
+                configArray
+            );
+        }
+
+
+        // check for options-based processing
+        if (options.preprocess || options.postprocess) {
+            return this._verifyWithFlatConfigArrayAndProcessor(textOrSourceCode, config, options);
+        }
```
mdjermanovic, line 1729: In the `if (config.processor)` path, `_verifyWithFlatConfigArray` calls `_verifyWithFlatConfigArrayAndProcessor` with `options.preprocess`, and then `_verifyWithFlatConfigArrayAndProcessor` calls `_verifyWithFlatConfigArray` with the same `options.preprocess`, so this will run `config.processor.preprocess()` again on its output:

```js
const { Linter } = require("eslint");

const linter = new Linter({ configType: "flat" });

const config = {
    files: ["*.md"],
    processor: {
        preprocess(text, filename) {
            console.log(`Called with text = '${text}', filename = '${filename}'`);

            return [{ text: "bar", filename: "0.js" }];
        },
        postprocess() {
            return [];
        }
    }
};

linter.verify("foo", config, "a.md");

// logs:
//       Called with text = 'foo', filename = 'a.md'
//       Called with text = 'bar', filename = 'a.md/0_0.js'
```

lib/linter/linter.js
```diff
@@ -1289,6 +1448,214 @@ class Linter {
...
+
+                parserOptions.sourceType = languageOptions.sourceType;
+
+                if (
+                    parserOptions.sourceType === "module" &&
+                    parserOptions.ecmaFeatures &&
+                    parserOptions.ecmaFeatures.globalReturn
+                ) {
+                    parserOptions.ecmaFeatures.globalReturn = false;
+                }
```
mdjermanovic, line 1558: `_verifyWithoutProcessors` (`resolveParserOptions`) is overwriting `globalReturn` regardless of the parser, but `_verifyWithFlatConfigArrayAndWithoutProcessors` will do that only if the parser is Espree?

nzakas: Yes. I don’t think we should be messing around with other parsers’ options.

lib/linter/linter.js
nzakas: Huh, nice catch. I'll look at that.

Just FYI: It was a little bit difficult to understand that you were pointing out a problem here. When you state the behavior you're seeing, it would help me out a lot if you'd say something like, "This behavior is unexpected and I don't see it when using eslintrc." This is probably just the brain fog, but little hints like saying "unexpected" make it a lot easier for me to decipher descriptions.

mdjermanovic (approved):
LGTM, thanks!

nzakas:
@btmills can you give this a once-over before I merge?

btmills (approved):
I only found one potential functional change where this branch is currently passing `void 0` as `parserName`. I left a couple inline suggestions, but if the `void 0` question is answered, this can be merged as-is.

I'm very excited to see this go into the release, even if undocumented. Great work!

lib/linter/linter.js
```diff
@@ -837,16 +945,16 @@ const BASE_TRAVERSAL_CONTEXT = Object.freeze(
  * @param {SourceCode} sourceCode A SourceCode object for the given text
  * @param {Object} configuredRules The rules configuration
  * @param {function(string): Rule} ruleMapper A mapper function from rule names to rules
- * @param {Object} parserOptions The options that were passed to the parser
  * @param {string} parserName The name of the parser in the config
```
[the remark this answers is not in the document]
nzakas: I want to break them. :) Passing an empty string or other string value will be a more difficult error to catch that having it be undefined. Because this is only in the flat config path, it will help us identify rules that rely on this early.

PR #15198, opened 2021-10-21
Chore: drop minimum node 14 version to 14.16.1

ErisDS:
refs: https://github.com/eslint/eslint/pull/14592#issuecomment-948567452

- this will allow the Ghost project to update their version of ESLint without impacting on all of our users

#### Prerequisites checklist

[ ] Documentation update
[ ] Add autofixing to a rule
[ ] Add a CLI option
[ ] Add something to the core
[x] Other, please explain:

I've changed the minimum version of Node 14 that ESLint supports from 14.17.0 back one version to 14.16.1.

The underlying reason for this is that in https://github.com/eslint/eslint/pull/14592 the change that intended to drop support for Node 10, 13 & 15 also set the minimum version of Node 14 that ESLint supports to 14.17.0.

For most projects this shouldn't be a problem, of course everyone should be on the latest and greatest Node!.

However, for Ghost - which is an application, not a library - we will have to change our own node engines from 14.16.1 to 14.17.0 in order to install the latest ESLint. ESLint is a dev dependency that we use when building Ghost, it is not relevant to our end-users. 

But if we change our node engines, then all the end users who run Ghost will have to upgrade their Node version in order to get the next version of Ghost.

It doesn't make sense for us to force this just to get a new version of a dev dependency that has no impact on a production install or an end user.

By pure chance, we're just 1 Node 14 version behind the version you have pinned - we're pinned to 14.16.1 rather than 14.17.0 and therefore I was hoping you would be willing to accept this minor low impact change in order to help us out.

I understand this is an odd request and very specific to one project, however:
- This would be extremely helpful for us
- I think this is an unintended side effect of the change to drop 10/13/15
- This change doesn't affect the intention of the original change

I do not believe there is a specific reason for the choice of 14.17.0, but if I'm wrong no worries.

Thanks for reviewing this PR & considering this change🙏


bmish:
Dropping ESLint's required Node version below the version that officially supports ESM (14.17.0) is a pretty big downside for ESLint, so I'm strongly against that. It's better for us to be more strict/safe about the versions we support. It's also generally easy for consumers to perform minor Node version upgrades (e.g. 14.16.1 to 14.17.0) to solve the problem.

> However, for Ghost - which is an application, not a library - we will have to change our own node engines from 14.16.1 to 14.17.0 in order to install the latest ESLint. ESLint is a dev dependency that we use when building Ghost, it is not relevant to our end-users.

I've run into this situation myself many times--it's a common situation with dependencies, but it's not specific to ESLint. The easiest solution is to just wait until you bump your project's Node version requirement in a future major release, assuming the upgrade isn't urgent. If you're using yarn, you could also try something like `yarn install --ignore-engines` when installing dependencies / running linting internally.

PR #15263, opened 2021-11-04
feat!: Make line breaks invalid for key-spacing strict mode

mathiasvr:
#### Prerequisites checklist

[ ] Documentation update
[ ] Add autofixing to a rule
[ ] Add a CLI option
[ ] Add something to the core
[ ] Other, please explain:

key-spacing

More warnings

New default behavior

With `key-spacing` rule set in default strict mode:

```json
"key-spacing": [ "error", { "afterColon": true, "beforeColon": true, "mode": "strict" } ]
```

Following code is currently valid, but will became invalid with this change since line breaks is treated whitespace:
```js
const obj = {
	key :
	'value'
};

const obj = {
	key
	:
	'value'
};
```

Following code is valid for this PR:
```js
// 'key-spacing': [ 'error', { 'afterColon': true, 'beforeColon': true, 'mode': 'strict' } ]
const obj = { key : 'value' };
const obj = {
	key : 'value'
};
```

For `mode` set to "minimum" the current behaviour is the same and first example is still valid.

Changed key-spacing rule and tests.

I believe this is preferred behaviour, but if not maybe a new option should be added?

### DIFF
--- a/lib/rules/key-spacing.js
+++ b/lib/rules/key-spacing.js
@@ -15,16 +15,6 @@ const astUtils = require("./utils/ast-utils");
 //------------------------------------------------------------------------------
 
 /**
- * Checks whether a string contains a line terminator as defined in
- * http://www.ecma-international.org/ecma-262/5.1/#sec-7.3
- * @param {string} str String to test.
- * @returns {boolean} True if str contains a line terminator.
- */
-function containsLineTerminator(str) {
-    return astUtils.LINEBREAK_MATCHER.test(str);
-}
-
-/**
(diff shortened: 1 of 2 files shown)


nzakas:
Thanks for the input. As [stated in the README](https://github.com/eslint/eslint#stylistic-rule-updates), we actually aren't making any further updates to rules that enforce stylistic preferences. If you have a strong preference for such a change, you can always copy the existing rule into a custom rule and make it behave the way you want.

mathiasvr:
@nzakas Ah sorry, I must have missed that.

Issue #12714, opened 2019-12-27
Expose ast-utils in the api


jimjenkins5:
**The version of ESLint you are using.**
6.x

**The problem you want to solve.**
In order to customize some rules for our specific use cases, the eslint team has at times suggested forking a core rule or creating our own. Most of the core rules us the ast-utils module internally. 

In order to fork these rules, we either need to fork ast-utils.js as well, or require the module by reaching deep into eslint codebase. In order to keep maintenance down on our plugin, we decided to go with the second route and require the file from the eslint code. 

When upgrading to eslint v6.x from v5 we found that our plugin broke because the file was moved internally.  

In addition to the problem with forking rules, this module would be very useful for creating custom rules, as the team put a lot of work into these very useful helper functions.

**Your take on the correct solution to problem.**

If these, and possibly other utilities, were exposed in the api export itself we could simply require the eslint module and use the `ast-utlis` property. (i.e. `require('eslint').astUtils` or similar)

**Are you willing to submit a pull request to implement this change?**
Yes

jimjenkins5:
Here's an example of a rule we had to customize that required forking the core rule and is resulting in the described problem:
https://github.com/eslint/eslint/issues/11217

kaicataldo:
Apologies that we missed this.

I do see the value in this. One thing that the team will need to decide is if we're okay taking on the extra maintenance burden of making this a public API, which limits the kinds of changes we can make in a release and requires proper documentation. @eslint/eslint-tsc Thoughts on this? This feels like something the TSC should evaluate.

kaicataldo:
Alternatively, some kind of 3rd party package of useful Estree AST utils could be useful beyond just ESLint!

not-an-aardvark:
I don't think we should make ast-utils itself part of the public API, since a lot of its methods weren't designed with long-term stability in mind. It has basically been a place where we throw utility methods that we need to share between rules, without needing to put too much thought into API design.

If there are specific methods within ast-utils that would be useful for third-party rules, then it seems like it would be worth discussing those. For the use case described in the original issue (forked rules having dependencies on ast-utils), I think forking the needed methods from ast-utils is probably the most stable solution since it's unlikely to break, even if ESLint's version of ast-utils changes in a major release.

ilyavolodin:
I agree. ast-utils were never meant to be public. I would highly suggest somebody copy and refactor them into external package if needed, since this package might be useful for other purposes as well (babel, etc.)

[...]

PR #15301, opened 2021-11-12
feat: Export ast-utils for use in custom rules

mathiasvr:
#### Prerequisites checklist

[ ] Documentation update
[ ] Add autofixing to a rule
[ ] Add a CLI option
[ ] Add something to the core
[x] Other, please explain:

Exports the built-in ast-utils so that customized rules can be made without reimplementing this functionality.

Some eslint rules starts by importing `ast-utils` like so:
```js
```

This change allows custom rules to use the same library through eslint as a peer dependency:
```js
const astUtils = require("eslint/ast-utils");
```

The [eslint-utils](https://www.npmjs.com/package/eslint-utils) module doesn't have all methods of ast-utils so cannot always be used as an alternative.
Exposing `ast-utils` through eslint makes it easier to customize rules and requires no extra dependencies since eslint is expected to be a peer dependency in projects using the rule.

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,6 +10,7 @@
   "exports": {
     "./package.json": "./package.json",
     ".": "./lib/api.js",
+    "./ast-utils": "./lib/rules/utils/ast-utils.js",
     "./use-at-your-own-risk": "./lib/unsupported-api.js"
   },
   "scripts": {


nzakas:
Duplicate of https://github.com/eslint/eslint/issues/12714

While we understand the reasoning, we don't want to expose this as part of the public API.

PR #15353, opened 2021-11-25
feat: redesign options for the `quotes` rule

snitin315:
#### Prerequisites checklist

[ ] Documentation update
[ ] Add autofixing to a rule
[ ] Add a CLI option
[ ] Add something to the core
[ ] Other, please explain:

Fix https://github.com/eslint/eslint/issues/12156

Redesign options for the `quotes` rule as per https://github.com/eslint/eslint/issues/12156#issuecomment-601512762

`/* eslint quotes: ['error', { precedence: ['single', 'double', 'backtick'], avoidEscape: true } ]*/`

### DIFF
--- a/lib/rules/quotes.js
+++ b/lib/rules/quotes.js
@@ -68,8 +68,6 @@ QUOTE_SETTINGS.backtick.convert = function(str) {
     }) + newQuote;
 };
 
-const AVOID_ESCAPE = "avoid-escape";
-
 //------------------------------------------------------------------------------
 // Rule Definition
 //------------------------------------------------------------------------------
@@ -87,30 +85,21 @@ module.exports = {
 
         fixable: "code",
 
-        schema: [
-            {
-                enum: ["single", "double", "backtick"]
-            },
-            {
-                anyOf: [
-                    {
-                        enum: ["avoid-escape"]
-                    },
-                    {
-                        type: "object",
-                        properties: {
-                            avoidEscape: {
-                                type: "boolean"
-                            },
-                            allowTemplateLiterals: {
-                                type: "boolean"
-                            }
-                        },
-                        additionalProperties: false
+        schema: [{
+            type: "object",
+            properties: {
+                precedence: {
+                    type: "array",
+                    minItems: 1,
+                    items: {
+                        enum: ["single", "double", "backtick"]
                     }
-                ]
+                },
+                avoidEscape: {
+                    type: "boolean"
+                }
             }
-        ],
+        }],
 
         messages: {
             wrongQuotes: "Strings must use {{description}}."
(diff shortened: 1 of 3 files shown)


aladdin-add:
are the old-style options allowed after the change?  I think they can be auto translated to the new-style options -- to avoid breaking changes.

snitin315:
I was considering this as a candidiate for v9 because of the breaking changes.

But, yes, we should keep the old-style schema for avoiding breaking changes if we want to ship this feature with v8.

And then we can completely remove the old syntax in v9.

nzakas:
@snitin315 are you still working on this? There’s linting and test errors.

Also, breaking changes need to be flagged as “feat!:”, and need to be discussed separately. It’s a bit early to start sending PRs for v9 :)

snitin315:
Should we keep the old syntax as well to avoid breaking changes? So that we can ship this new syntax with v8. We could log a deprecation warning when when using the old syntax.

aladdin-add:
For a smoother UX, I would prefer to also allow the old options if it does not add too many maintenances burden.
and, we don't have a rule options deprecation process (though good to have IMO :)).

nzakas:
Yeah, we don’t really have a way to deprecate old config formats. We need to continue supporting the old way going forward, just mark it as deprecated in the docs.

noahsark769:
This is awesome! @snitin315 happy to help with this if I can, also agreed about keeping the old syntax compatible

snitin315:
This feature is definitely going to be a big breaking change. Even if we allow the old syntax by converting the old syntax options to new syntax internally, the working of this rule will be changed and it will produce more warnings.

### Consider the following example with old syntax - 

```js
/*eslint quotes: ["error", "double", { "allowTemplateLiterals": true }]*/

let double = "double"; // ok
double = `double`; // ok
double = `has ' single`; // ok
double = `has " double`; // ok
```

### Possible conversion to new syntax as per https://github.com/eslint/eslint/issues/12156#issuecomment-601512762 - 

```js
/*eslint quotes: ["error", { precedence: ["double", "backtick"] }]*/

let double = "double"; // ok
double = `double`; // error,  fixed to "double". 
double = `has ' single`; // error,  fixed to "has ' single". 
double = `has " double`; // error,  fixed to "has \" double". 
```

```js
/*eslint quotes: ["error", { precedence: ["double", "backtick"], avoidEscape: true }]*/

let double = "double"; // ok
double = `double`; // fixed to "double".
double = `has ' single`; // error,  fixed to "has ' single". 
double = `has " double`; // ok
```

snitin315:
I am closing this PR for now since it's too early for v9 PRs. I will reopen when we are close to v9.

PR #15371, opened 2021-11-29
chore: Use default Chromium binary in M1 Mac tests

btmills:
#### Prerequisites checklist

[ ] Documentation update
[ ] Add autofixing to a rule
[ ] Add a CLI option
[ ] Add something to the core
[x] Other, please explain: Update test config to work automatically on M1 Macs.

On M1 Macs, we're able to use the Chromium binary that Puppeteer downloads as part of its install step. Prior to this change, running `npm test` on an M1 Mac would fail with "Cannot start ChromeHeadless" because it couldn't find the binary.

Should I do `os.platform !== "darwin"` instead? As is, this changes behavior for all ARM-based non-Linux environments. Windows on ARM wouldn't work with the previous version, and it might with this change, but I don't have a way to confirm.

### DIFF
--- a/karma.conf.js
+++ b/karma.conf.js
@@ -2,7 +2,7 @@
 const os = require("os");
 const NodePolyfillPlugin = require("node-polyfill-webpack-plugin");
 
-if (os.arch() === "arm64") {
+if (os.platform === "linux" && os.arch() === "arm64") {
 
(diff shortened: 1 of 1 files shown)


nzakas (approved):
I think this is fine to start. Let’s see what happens.

PR #15404, opened 2021-12-08
docs: consistency changes

BBosman:
#### Prerequisites checklist

[X] Documentation update
[ ] Add autofix to a rule
[ ] Add a CLI option
[ ] Add something to the core
[X] Other, please explain:

I've made a bunch of consistency changes to the rules documentation, including:
* Fixed header casing (i.e. `options` -> `Options`)
* Fixed header levels (i.e. `## allow` -> `### allow`)
* Removed duplicate headers (i.e. multiple `## Rule Details` headers)
* Made all inter-rule links internal links if they weren't already (i.e. `[object-shorthand](https://eslint.org/docs/rules/object-shorthand)` -> `[object-shorthand](object-shorthand.md)`)
* Made unordered list style consistent (i.e. `- ` -> `* `). Almost everywhere `* ` was already used so fixed the few places it wasn't. I also added an entry to `markdownlint.yml` to ensure future consistency. This also resulted in a couple of changes outside of the `rules` directory.
* Made the order of sections consistent. The order is now always `"Rule Details", "Options", "Environments", "Examples", "Known Limitations", "When Not To Use It", "Related Rules", "Compatibility", "Further Reading"`. Especially those last three had quite a few permutations. The order is arbitrary, but made sense to me.
* Added a test to `Makefile.js` to validate those header names and their order
* Some typo fixes and similar minor changes.

This is my first PR to `eslint` and I think I followed all guidelines, but I'm happy to make improvements to the PR if needed.

### DIFF
--- a/.github/ISSUE_TEMPLATE/NEW_SYNTAX.md
+++ b/.github/ISSUE_TEMPLATE/NEW_SYNTAX.md
@@ -32,12 +32,14 @@ assignees: ''
 
 Please check off all items that have already been completed. Be sure to paste the pull request URLs next to each item so we can verify the work as done.
 
-* [ ] Ecma262 update: <!-- paste PR URL for this syntax here -->
-* [ ] ESTree update: <!-- paste PR URL for this ESTree update here -->
-* [ ] Acorn update: <!-- paste PR URL for this syntax here -->
-* [ ] `eslint-visitor-keys` update: <!-- paste PR URL for this syntax here -->
-* [ ] `espree` update: <!-- paste PR URL for this syntax here -->
-* [ ] `eslint-scope` update: <!-- paste PR URL for this syntax here -->
-* [ ] `eslint` update: <!-- paste PR URL for this syntax here -->
+- [ ] Ecma262 update: <!-- paste PR URL for this syntax here -->
+- [ ] ESTree update: <!-- paste PR URL for this ESTree update here -->
+- [ ] Acorn update: <!-- paste PR URL for this syntax here -->
+- [ ] `eslint-visitor-keys` update: <!-- paste PR URL for this syntax here -->
+- [ ] `espree` update: <!-- paste PR URL for this syntax here -->
+- [ ] `eslint-scope` update: <!-- paste PR URL for this syntax here -->
+- [ ] `eslint` update: <!-- paste PR URL for this syntax here -->
 
 **Are you willing to submit a pull request to implement this syntax?**
+
+<!-- markdownlint-disable-file MD004 -->
--- a/.github/ISSUE_TEMPLATE/NEW_SYNTAX.md
+++ b/.github/ISSUE_TEMPLATE/NEW_SYNTAX.md
@@ -20,7 +20,7 @@ assignees: ''
 
 **Please provide the TC39 URL for the syntax proposal:**
 
-
+<!-- for example, <https://github.com/tc39/proposal-top-level-await> -->
 
 **Please provide some example code for the new syntax:**
 
@@ -32,12 +32,12 @@ assignees: ''
 
 Please check off all items that have already been completed. Be sure to paste the pull request URLs next to each item so we can verify the work as done.
 
-- [ ] Ecma262 update: <!-- paste PR URL for this syntax here -->
-- [ ] ESTree update: <!-- paste PR URL for this ESTree update here -->
-- [ ] Acorn update: <!-- paste PR URL for this syntax here -->
-- [ ] `eslint-visitor-keys` update: <!-- paste PR URL for this syntax here -->
-- [ ] `espree` update: <!-- paste PR URL for this syntax here -->
-- [ ] `eslint-scope` update: <!-- paste PR URL for this syntax here -->
-- [ ] `eslint` update: <!-- paste PR URL for this syntax here -->
+* [ ] Ecma262 update: <!-- paste PR URL for this syntax here -->
+* [ ] ESTree update: <!-- paste PR URL for this ESTree update here -->
+* [ ] Acorn update: <!-- paste PR URL for this syntax here -->
+* [ ] `eslint-visitor-keys` update: <!-- paste PR URL for this syntax here -->
+* [ ] `espree` update: <!-- paste PR URL for this syntax here -->
+* [ ] `eslint-scope` update: <!-- paste PR URL for this syntax here -->
+* [ ] `eslint` update: <!-- paste PR URL for this syntax here -->
 
 **Are you willing to submit a pull request to implement this syntax?**
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -6,7 +6,7 @@
 
 #### Prerequisites checklist
 
-* [ ] I have read the [contributing guidelines](https://github.com/eslint/eslint/blob/HEAD/CONTRIBUTING.md).
+- [ ] I have read the [contributing guidelines](https://github.com/eslint/eslint/blob/HEAD/CONTRIBUTING.md).
 
 #### What is the purpose of this pull request? (put an "X" next to an item)
 
@@ -43,3 +43,5 @@
 #### What changes did you make? (Give an overview)
 
 #### Is there anything you'd like reviewers to focus on?
+
+<!-- markdownlint-disable-file MD004 -->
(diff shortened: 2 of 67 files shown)


.github/ISSUE_TEMPLATE/NEW_SYNTAX.md
```diff
@@ -20,7 +20,7 @@ assignees: ''
 
 **Please provide the TC39 URL for the syntax proposal:**
 
-
+<!-- for example, <https://github.com/tc39/proposal-decorators> -->
```
aladdin-add, line 23: good to replace it with a finished proposal - in general, we do not accept stage<4 features. it's a bit confusing as decorator is still in stage-2.

aladdin-add (approved):
LGTM, just a minor suggestion.

.github/ISSUE_TEMPLATE/NEW_SYNTAX.md
BBosman: Valid point. I changed it to the stage 4 `top-level-await` proposal.

aladdin-add (approved):
LGTM, thanks!

.github/ISSUE_TEMPLATE/NEW_SYNTAX.md
```diff
@@ -32,12 +32,12 @@ assignees: ''
 
 Please check off all items that have already been completed. Be sure to paste the pull request URLs next to each item so we can verify the work as done.
 
-- [ ] Ecma262 update: <!-- paste PR URL for this syntax here -->
```
nzakas, line 35: I’m pretty sure this needs to be a dash in order to create a checklist in the issue. This isn’t just a regular bullet list.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -6,7 +6,7 @@
 
 #### Prerequisites checklist
 
-- [ ] I have read the [contributing guidelines](https://github.com/eslint/eslint/blob/HEAD/CONTRIBUTING.md).
+* [ ] I have read the [contributing guidelines](https://github.com/eslint/eslint/blob/HEAD/CONTRIBUTING.md).
```
nzakas, line 9: Same. Pretty sure this needs to be a dash to make the checkbox we want.

Makefile.js
```diff
@@ -853,6 +884,12 @@ target.checkRuleFiles = function() {
                 console.error("Missing id in the doc page's title of rule %s", basename);
                 errors++;
             }
+
+            // check for proper doc headers
+            if (!hasKnownHeaders(basename)) {
```
nzakas, line 889: It looks like this function doesn’t accept any arguments?

nzakas (changes requested):
Thanks, this is some nice work. Just a handful of changes I don’t think we can make but overall looks good.

.github/ISSUE_TEMPLATE/NEW_SYNTAX.md
BBosman: Looking at the GitHub docs for checklists I think you're right. But that's going to conflict with the changes I made in `.markdownlint.yml` as `* ` is now the required way, so there are a couple of ways I could "fix" this:

1. Revert the changes in `.markdownlint.yml `. This would reinstate the old behaviour of consistency within single files, but not across the repo. So this PR brings them all in line, but they could drift apart again in the future.
1. Add `NEW_SYNTAX.md` and `PULL_REQUEST_TEMPLATE.md` to `.markdownlintignore`. But this would also suppress other (future) linting issues in those files.
1. Add `<!-- markdownlint-disable-file MD004 -->` to the top/bottom of those two files. This would probably show up in the GitHub UI when creating a PR though.
1. Wrap those specific lines with `<!-- markdownlint-disable MD004 -->` and `<!-- markdownlint-enable MD005 -->` in those two files. This would probably show up in the GitHub UI when creating a PR though.

All of them have some negative side-effects, so @nzakas which one do you prefer that I implement?

Makefile.js
BBosman: Originally I had the function do the logging and it used this argument for that, but I refactored it to do the logging outside the function here, to bring it in line with the others checks and I forgot to remove the argument here. Will fix this when I make the other changes.

.github/ISSUE_TEMPLATE/NEW_SYNTAX.md
nzakas: I think option 3 makes the most sense for now. We should really convert this file into YAML format anyway.

BBosman:
@nzakas, I think I addressed all your comments.

Makefile.js
```diff
@@ -790,7 +791,9 @@ target.checkRuleFiles = function() {
         const basename = path.basename(filename, ".js");
         const docFilename = `docs/rules/${basename}.md`;
         const docText = cat(docFilename);
+        const docMarkdown = marked.lexer(docText, { gfm: true, silent: false });
         const ruleCode = cat(filename);
+        const knownHeaders = ["Rule Details", "Options", "Environments", "Examples", "Known Limitations", "When Not To Use It", "Related Rules", "Compatibility", "Further Reading"];
```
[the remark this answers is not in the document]
BBosman: It wasn't in that order in `guard-for-in.md`, but that was indeed the only one. The order of "Related Rules" and "Compatibility" has more inconsistencies.

My reasoning for the current order of `"Related Rules", "Compatibility", "Further Reading"` in the PR was "Internal rule links", "External rule links" and "Other external links", so progressing from internal links to external links.

But as I said in my initial PR it's a bit of an arbitrary choice, so I'm willing to change it to `"Further Reading", "Related Rules", "Compatibility"` or even something else if that's the consensus.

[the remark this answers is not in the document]
nzakas: Further Reading is meant to always be last because it’s intended to link outside of the ESLint site. We don’t want to send people off site in the middle of the page. 

I like the order of Related Rules, Compatiblity, then Further Reading.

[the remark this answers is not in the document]
mdjermanovic: `"Further Reading"` is usually more related to the rule than the related rules are. It provides more info on the context of the problem the rule aims to solve, while `"Related Rules"` is more like leaving the rule to read about another rule. I think that's the reason why the convention was to have `"Further Reading"` before `"Related Rules"`, even though they are external links.

"Further Reading", "Related Rules", "Compatibility" makes sense to me. @nzakas what do you think about this?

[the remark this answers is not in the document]
mdjermanovic: > Further Reading is meant to always be last because it’s intended to link outside of the ESLint site. We don’t want to send people off site in the middle of the page.
> 
> I like the order of Related Rules, Compatiblity, then Further Reading.

For some reason, I didn't see this post while I was writing my previous one. Okay, if that was the intent then we should follow it.

docs/rules/accessor-pairs.md
```diff
@@ -273,14 +273,15 @@ var o = {
 
 The code above creates an object with just a setter for the property `"a"`.
 
-See [no-dupe-keys](no-dupe-keys.md) if you also want to disallow duplicate keys in object literals.
-
-See [no-dupe-class-members](no-dupe-class-members.md) if you also want to disallow duplicate names in class definitions.
-
```
mdjermanovic, line 279: I think this should stay here, as it's in the context of the limitations explained in this section.

mdjermanovic (commented):
Very nice work, thanks! I left only two small suggestions.

docs/rules/no-negated-in-lhs.md
```diff
@@ -6,8 +6,6 @@ This rule was **deprecated** in ESLint v3.3.0 and replaced by the [no-unsafe-neg
 
 Just as developers might type `-a + b` when they mean `-(a + b)` for the negative of a sum, they might type `!key in object` by mistake when they almost certainly mean `!(key in object)` to test that a key is not in an object.
 
-## Rule Details
-
```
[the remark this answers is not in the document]
BBosman: Fixed.

docs/rules/accessor-pairs.md
BBosman: Changed.

mdjermanovic (approved):
LGTM, thanks!

snitin315 (approved):
Nice enhancement. Thank you for your contribution.

nzakas (approved):

PR #15436, opened 2021-12-18
feat: allow to define `eslint-disable-next-line` in multiple lines

snitin315:
#### Prerequisites checklist

[X] Other, please explain: Allow `eslint-disable-next-line` directive to be of multiple lines

Fixes https://github.com/eslint/eslint/issues/14284

```js
/* eslint-disable-next-line global-require --
   description on why this exception is seen as appropriate but past a 
   comfortable reading line length */
const fizz = require(buzz); // ignored

/* eslint-disable-next-line 
    no-alert
*/
alert("buzz"); // ignored
```

None

### DIFF
--- a/lib/linter/apply-disable-directives.js
+++ b/lib/linter/apply-disable-directives.js
@@ -261,17 +261,22 @@ function applyDirectives(options) {
     const processed = processUnusedDisableDirectives(unusedDisableDirectivesToReport);
 
     const unusedDisableDirectives = processed
-        .map(({ description, fix, position }) => ({
-            ruleId: null,
-            message: description
-                ? `Unused eslint-disable directive (no problems were reported from ${description}).`
-                : "Unused eslint-disable directive (no problems were reported).",
-            line: position.parentComment.commentToken.loc.start.line,
-            column: position.parentComment.commentToken.loc.start.column + 1,
-            severity: options.reportUnusedDisableDirectives === "warn" ? 1 : 2,
-            nodeType: null,
-            ...options.disableFixes ? {} : { fix }
-        }));
+        .map(({ description, fix, position }) => {
+            const { parentComment, type, line, column } = position;
+            const { commentToken } = parentComment;
+
+            return {
+                ruleId: null,
+                message: description
+                    ? `Unused eslint-disable directive (no problems were reported from ${description}).`
+                    : "Unused eslint-disable directive (no problems were reported).",
+                line: type === "disable-next-line" && commentToken.loc ? commentToken.loc.start.line : line,
+                column: type === "disable-next-line" && commentToken.loc ? commentToken.loc.start.column + 1 : column,
+                severity: options.reportUnusedDisableDirectives === "warn" ? 1 : 2,
+                nodeType: null,
+                ...options.disableFixes ? {} : { fix }
+            };
+        });
 
     return { problems, unusedDisableDirectives };
 }
--- a/lib/linter/apply-disable-directives.js
+++ b/lib/linter/apply-disable-directives.js
@@ -266,8 +266,8 @@ function applyDirectives(options) {
             message: description
                 ? `Unused eslint-disable directive (no problems were reported from ${description}).`
                 : "Unused eslint-disable directive (no problems were reported).",
-            line: position.line,
-            column: position.column,
+            line: position.parentComment.commentToken.loc.start.line,
+            column: position.parentComment.commentToken.loc.start.column + 1,
             severity: options.reportUnusedDisableDirectives === "warn" ? 1 : 2,
             nodeType: null,
             ...options.disableFixes ? {} : { fix }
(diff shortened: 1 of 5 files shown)


lib/linter/linter.js
```diff
@@ -305,7 +305,11 @@ function createDisableDirectives(options) {
 
         // push to directives, if the rule is defined(including null, e.g. /*eslint enable*/)
         if (ruleId === null || !!ruleMapper(ruleId)) {
-            result.directives.push({ parentComment, type, line: commentToken.loc.start.line, column: commentToken.loc.start.column + 1, ruleId });
+            if (type === "disable-next-line") {
+                result.directives.push({ parentComment, type, line: commentToken.loc.end.line, column: commentToken.loc.end.column + 1, ruleId });
```
mdjermanovic, line 309: A problem with changing these `line` and `column` is that they're also used as the position to report the directive as unused, so it would look like this:

image

We could modify the code that creates unused disable directive errors to use `loc` from `parentComment.commentToken` instead of these `line` and `column`, or maybe even consider entirely removing these `line` and `column` and using only `parentComment.commentToken`  in `apply-disable-directives.js` for everything, since the logic of where disable directives apply is in `apply-disable-directives.js` anyway.

mdjermanovic:
It would be good to update this section in the docs with multiline `eslint-disable-next-line` comments:

https://eslint.org/docs/user-guide/configuring/rules#using-configuration-comments-1

lib/linter/linter.js
snitin315: fixed.

lib/linter/apply-disable-directives.js
```diff
@@ -261,17 +261,21 @@ function applyDirectives(options) {
...
+        .map(({ description, fix, position }) => {
+            const { parentComment, type, line, column } = position;
+
+            return {
+                ruleId: null,
+                message: description
+                    ? `Unused eslint-disable directive (no problems were reported from ${description}).`
+                    : "Unused eslint-disable directive (no problems were reported).",
+                line: type === "disable-next-line" ? parentComment.commentToken.loc.start.line : line,
+                column: type === "disable-next-line" ? parentComment.commentToken.loc.start.column + 1 : column,
```
mdjermanovic, line 273: Can we always use `parentComment.commentToken.loc` to simplify?

```suggestion
                line: type === parentComment.commentToken.loc.start.line,
                column: type === parentComment.commentToken.loc.start.column + 1,
```

mdjermanovic: If this would break tests in `tests/lib/linter/apply-disable-directives.js`, then we can leave it as is. In my opinion, those tests require an overly complicated and error-prone setup, and are very difficult to maintain, and it's questionable how useful they are, so I think we could consider dropping them at some point in favor of equivalent tests in `tests/lib/linter/linter.js`.

snitin315: We will have to refactor the `tests/lib/linter/apply-disable-directives.js` to include `loc` - https://github.com/eslint/eslint/blob/eaa08d3055b195bce59cc96bb63ac29038cd7c7d/tests/lib/linter/apply-disable-directives.js#L26-L31

mdjermanovic: That looks too complicated and unmaintainable, so let's leave this as is in the current version of this PR. I think we should (in another PR) entirely delete `tests/lib/linter/apply-disable-directives.js` and convert all those tests to `tests/lib/linter/linter.js` tests.

lib/linter/apply-disable-directives.js
```diff
@@ -261,17 +261,21 @@ function applyDirectives(options) {
...
-                ? `Unused eslint-disable directive (no problems were reported from ${description}).`
-                : "Unused eslint-disable directive (no problems were reported).",
-            line: position.line,
-            column: position.column,
-            severity: options.reportUnusedDisableDirectives === "warn" ? 1 : 2,
-            nodeType: null,
-            ...options.disableFixes ? {} : { fix }
-        }));
+        .map(({ description, fix, position }) => {
+            const { parentComment, type, line, column } = position;
```
mdjermanovic, line 265: It looks confusing to get `parentComment` and `type` from a property named `position`. Since that object is actually the original directive object passed in by the linter, maybe we could rename the property to `unprocessedDirective` (as that's how it is called in the rest of the code)?

snitin315: Done.

mdjermanovic (approved):
LGTM, thanks!

Since this is a change in the core, I'll leave it open for a few days in case someone else wants to review it, and then merge.

thernstig:
@snitin315 some of us prefer to reserver `/*` just for JSDoc comments and instead use multi-line `//` for the rest of the comments.

Is there any chance support could be added for this? I assume it was not as part of this as I cannot see it in the docs:
```js
// eslint-disable-next-line global-require -- Some text here as well
// description on why this exception is seen as appropriate but past a 
// comfortable reading line length
```

snitin315:
@thernstig please open a new issue for this.

PR #15508, opened 2022-01-11
feat: lint files concurrently

### LINKED ISSUE #3565: Lint multiple files in parallel [$500]
This is a discussion issue for adding ability to run eslint in parallel for multiple files.

The idea is that ESLint is mostly CPU bound, not IO bound, so creating multiple threads (for machine with multiple cores) might (and probably will) increase performance in a meaningful way. The downside is that currently ESLint's codebase is synchronous. So this would require rewriting everything up to and including eslint.js to be asynchronous, which would be a major effort.

I played with this a little while ago and found a few libraries for Node that handle thread pool, including detection of number of cores available on the machine.
- [Node-threads-a-gogo](https://github.com/xk/node-threads-a-gogo) - seems pretty good, but looks dead.
- [nPool](https://github.com/inh3/nPool) - seems actively in development, but has native components (C++)
- [Node WebWorkers](https://github.com/audreyt/node-webworker-threads) - seems pretty dead too.
- [Parallel](https://github.com/adambom/parallel.js) - seems dead, and no pool implementation.
- [Node Clusters](https://nodejs.org/api/cluster.html) - not stable yet, and probably isn't going to be available on Node v0.10
- WebWorkers - seems that they are only implemented in io.js
  And there are a ton of other libraries out there for this.

If anyone had any experience writing multithreaded applications for node.js and would like to suggest alternatives or comment on the above list, please feel free.

P.S. https://www.airpair.com/javascript/posts/which-async-javascript-libraries-should-i-use
## <bountysource-plugin>

Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/26284182-lint-multiple-files-in-parallel?utm_campaign=plugin&utm_content=tracker%2F282608&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F282608&utm_medium=issues&utm_source=github).
</bountysource-plugin>

paulbrimicombe:
added a conncurency option that sets the number of threads to be used when linting files.

fixes #3565

#### Prerequisites checklist

[ ] Documentation update
[ ] Add autofix to a rule
[ ] Add a CLI option
[ ] Add something to the core
[x] Other, please explain:

I have added a `concurrency` option that sets the number of worker threads used when linting files. This defaults to `1`.

How the concurrency works:

- When `concurrency > 1`, all of the files that are to be linted are fetched up-front.
- A pool of `concurrency` worker threads is created, each containing an `ESLint` instance with `concurrency = 1`.
- Chunks of files to lint are passed to these worker threads until all files have been linted.
- When the workers finish, they pass the result array back to the main thread where the results are combined.

This is a fairly simple solution to check that the maintainers are happy with the general approach. I'm sure that there are lots of things that I've missed — please point these out!

Performance:

Below is a table indicating how the linting time for the `eslint` repository varies with concurrency. Results will probably vary considerably according to the number of files to be linted, the machine spec, the plugins used, etc.).

| Concurrency | Time to run `eslint **/*.js` (s) |
|------|------|
| 1 | 65 |
| 2 | 34 |
| 3 | 26 |
| 4 | 28 |
| 5 | 27 |
| 6 | 29 |

This is all based on running on my local machine (8 cores). Clearly there is great benefit in `concurrency > 1` but there are diminishing returns above `concurrency = 3`, presumably because of resource contention.

Caching:

One thing I do know is that caching doesn't quite work correctly. With the current implementation, workers read from the cache when they start linting a batch of files and then write back when the batch is finished. Any updates to the cache in the meantime (e.g. from another thread) are lost. This means that to populate the cache fully you have to run `eslint` multiple times.  With `concurrency = 5` the cache is still only ~85% filled after 10 runs.

Some options to fix this include:

- Read the cache file contents again immediately before writing, merge the current results and then write. This decreases the window of time during which changes will be lost.
- Say that concurrency and caching cannot be used together.
- Refactor the cache code so that we read the cache once before we start, accumulate all of the changes in all of the different threads and then write the result at the end.

Some guidance on the best approach to take would be appreciated.

Testing:

I have added tests for the new CLI argument but I haven't added any tests specifically for concurrent linting. Ideally I'd like to run pretty much all of the tests with and without concurrency but some guidance on how that might be achieved would be welcome. I have manually adjusted the code to force a concurrency of 2 for a test run and all but three tests pass — two of these are because they `spy` on function interactions that are now in workers. Running the tests with concurrency is also slower than without because of the overhead of continually spawning and terminating workers.

Notes:

- NodeJS worker threads seem to be supported in all of the NodeJS versions listed in the `package.json` `engines` field.
- You can only pass data that can be serialised using the [Structured Clone algorithm](https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Structured_clone_algorithm) to worker threads. There are some properties of `ESLintOptions` that can be functions (e.g. the `plugins` and `fix` properties). When the options cannot be cloned the current implementation falls back to linting in a single thread.

- Advice on how to deal with the caching issues described above.
- Advice on how / where to test this.

### DIFF
--- a/docs/developer-guide/nodejs-api.md
+++ b/docs/developer-guide/nodejs-api.md
@@ -140,6 +140,8 @@ The `ESLint` constructor takes an `options` object. If you omit the `options` ob
   Default is `[]`. An array of paths to directories to load custom rules from.
 * `options.useEslintrc` (`boolean`)<br>
   Default is `true`. If `false` is present, ESLint doesn't load configuration files (`.eslintrc.*` files). Only the configuration of the constructor options is valid.
+* `options.concurrency` (`number`)<br>
+  Default is 1. Sets the concurrency level to use when linting files.
 
 ##### Autofix
 
--- a/docs/user-guide/command-line-interface.md
+++ b/docs/user-guide/command-line-interface.md
@@ -86,6 +86,7 @@ Miscellaneous:
   -h, --help                      Show help
   -v, --version                   Output the version number
   --print-config path::String     Print the configuration for the given file
+  --concurrency                   The concurrency level to use when linting files
 ```
 
 Options that accept array values can be specified by repeating the option or with a comma-delimited list (other than `--ignore-pattern` which does not allow the second style).
@@ -493,6 +494,10 @@ This option outputs the configuration to be used for the file passed. When prese
 Example:
 
     eslint --print-config file.js
+
+#### `--concurrency`
+
+This option sets the concurrency level when linting files. Linting may be faster for large when this option is greater than 1.
 
 ## Ignoring files from linting
 
--- a/lib/cli-engine/cli-engine.js
+++ b/lib/cli-engine/cli-engine.js
@@ -57,6 +57,7 @@ const validFixTypes = new Set(["directive", "problem", "suggestion", "layout"]);
 /** @typedef {import("../shared/types").Rule} Rule */
 /** @typedef {ReturnType<CascadingConfigArrayFactory.getConfigArrayForFile>} ConfigArray */
 /** @typedef {ReturnType<ConfigArray.extractConfig>} ExtractedConfig */
+/** @typedef {import("./file-enumerator").FileEnumerator} FileEnumerator */
 
 /**
  * The options to configure a CLI engine with.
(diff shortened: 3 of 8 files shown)


nzakas:
Oh wow, thanks for digging into this. For changes this large, we do require an [RFC](https://github.com/eslint/rfcs) be created before we consider merging. So while we can take a look at this as inspiration, we can't move forward at this point.

As it is, we are in the process of removing `CLIEngine`, so we don't want to be adding a lot of new functionality into it at this point. We are also working on a [new config system](https://github.com/eslint/eslint/issues/13481) that will change how the core works, so we really need to wait for these pieces to be finished before working on parallel linting.

paulbrimicombe:
> Oh wow, thanks for digging into this. For changes this large, we do require an [RFC](https://github.com/eslint/rfcs) be created before we consider merging. So while we can take a look at this as inspiration, we can't move forward at this point.

Ah. Thanks for getting in touch @nzakas . I hadn't spotted that there was an RFC in progress. Is it worth me linking this PR to the existing RFC?

> As it is, we are in the process of removing `CLIEngine`, so we don't want to be adding a lot of new functionality into it at this point. We are also working on a [new config system](https://github.com/eslint/eslint/issues/13481) that will change how the core works, so we really need to wait for these pieces to be finished before working on parallel linting.

Hmm. OK. The approach taken in this PR should work with whatever changes you make provided that:

- A list of the files to be linted can be generated (could be a generator or just an array as it is in this PR).
- The configuration options for `ESLint` remain serialisable by Structure Clone for most users.

Given that the `ESLintConcurrent` class just uses `ESLint` but forces `concurrency = 1` that part of the implementation would remain largely unchanged.

Is there any reason to think that changing the configuration system is going to cause any problems with this pattern given that I'm not actually interacting with the configuration values at all, just passing them verbatim into `ESLint`?

nzakas:
You can definitely review the RFC and see if your approach aligns with what’s in there. It’s been a while since I read it over so I can’t tell you off the top of my head.

In the new config system, the configs aren’t serializable, so that will definitely have an effect on the design. It’s worth reviewing that RFC as well: https://github.com/eslint/rfcs/tree/main/designs/2019-config-simplification

nzakas:
Closing because we need to wait for the new config system.

PR #15849, opened 2022-05-07
build: add gh action simeg/urlsup-action

aladdin-add:
it is to validate the links in the docs are up.:)
Run at 00:00 UTC every Sunday.

#### Prerequisites checklist

[ ] Documentation update
[ ] Add autofix to a rule
[ ] Add a CLI option
[ ] Add something to the core
[ x] Other, please explain:

### DIFF
--- a/.github/workflows/urlsup.yml
+++ b/.github/workflows/urlsup.yml
@@ -1,0 +1,17 @@
+name: urlsup
+on:
+  schedule:
+    # Run at 00:00 UTC every Sunday
+    - cron: '0 0 * * */7'
+env:
+  FORCE_COLOR: 1
+jobs:
+  urlsup:
+    runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


nzakas:
This doesn’t seem like a good idea. Why do we care if links are up? That doesn’t seem important for building the website.

aladdin-add:
I noticed some dead links on our website, and I'd like to set a scheduled task to find them, so we can update/remove them. but surely it won't block the building and deployment. thoughts?

nzakas:
I think a better approach would be to have an npm script that checks URLs, so you can run it locally if you want.

I don’t like this GitHub action approach because it always checks every file, even if you didn’t edit it, meaning it will create a lot of noise for contributions. And running periodically isn’t helpful because it won’t notify anyone.

PR #15902, opened 2022-05-20
docs: Add rules meta info to rule pages

nzakas:
#### Prerequisites checklist

[x] Documentation update
[ ] Add autofix to a rule
[ ] Add a CLI option
[ ] Add something to the core
[ ] Other, please explain:

This PR makes several changes to the rules doc pages:

1. It inserts the description from the JavaScript rule source file under the header.
2. It shows panels for recommended, fixable, and suggestions at the top of the pages.
3. It updates Makefile.js to include functionality to generate a rules meta data JSON file for use in the docs site. This will be done automatically during site generation of the current eslint.org so it will be updated with each release.

Don't worry too much about the styling right now...I still need to fix that. 

In the new site, the description tends to be duplicated because each rule's first line of content is the rule description. We can fix that once we move over to the new site.

Otherwise, does everything appear in the correct spot?

### DIFF
--- a/Makefile.js
+++ b/Makefile.js
@@ -181,9 +181,13 @@ function generateFormatterExamples(formatterInfo, prereleaseVersion) {
  * @returns {void}
  */
 function generateRuleIndexPage() {
-    const outputFile = "../website/_data/rules.yml",
+    const legacyWebsiteOutputFile = "../website/_data/rules.yml",
+        docsSiteOutputFile = "docs/src/_data/rules.json",
+        docsSiteMetaOutputFile = "docs/src/_data/rules_meta.json",
         ruleTypes = "conf/rule-type-list.json",
         ruleTypesData = JSON.parse(cat(path.resolve(ruleTypes)));
+
+    const meta = {};
 
     RULE_FILES
         .map(filename => [filename, path.basename(filename, ".js")])
@@ -192,6 +196,17 @@ function generateRuleIndexPage() {
             const filename = pair[0];
             const basename = pair[1];
             const rule = require(path.resolve(filename));
+
+            /*
+             * Eleventy interprets the {{ }} in messages as being variables,
+             * which can cause an error if there's syntax it doesn't expect.
+             * Because we don't use this info in the website anyway, it's safer
+             * to just remove it.
+             */
+            meta[basename] = {
+                ...rule.meta,
+                messages: void 0
+            };
 
             if (rule.meta.deprecated) {
                 ruleTypesData.deprecated.rules.push({
@@ -219,9 +234,12 @@ function generateRuleIndexPage() {
     // `.rules` will be `undefined` if all rules in category are deprecated.
     ruleTypesData.types = ruleTypesData.types.filter(ruleType => !!ruleType.rules);
 
-    const output = yaml.dump(ruleTypesData, { sortKeys: true });
+    JSON.stringify(ruleTypesData, null, 4).to(docsSiteOutputFile);
+    JSON.stringify(meta, null, 4).to(docsSiteMetaOutputFile);
 
-    output.to(outputFile);
+    const legacyOutput = yaml.dump(ruleTypesData, { sortKeys: true });
+
+    legacyOutput.to(legacyWebsiteOutputFile);
 }
 
 /**
@@ -757,6 +775,8 @@ target.gensite = function(prereleaseVersion) {
 
     echo("Done generating eslint.org");
 };
+
+target.generateRuleIndexPage = generateRuleIndexPage;
 
 target.webpack = function(mode = "none") {
     exec(`${getBinFile("webpack")} --mode=${mode} --output-path=${BUILD_DIR}`);
--- a/docs/src/_includes/layouts/doc.html
+++ b/docs/src/_includes/layouts/doc.html
@@ -3,6 +3,7 @@ layout: base.njk
 ---
 
 {% include "partials/docs-header.html" %}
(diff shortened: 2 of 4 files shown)


docs/src/_includes/layouts/doc.html
```diff
@@ -35,9 +37,20 @@ <h2 id="further-reading">Further Reading</h2>
...
+    {% set rule_meta = rules_meta[title] %}
+
     <div class="docs-content">
         <main id="main" tabindex="-1" class="docs-main">
             <h1>{{ title }}</h1>
+            {% if rule_meta %}
+                <p>{{ rule_meta.docs.description | markdown | safe }}</p>
+                {{ ruleCategories({
+                recommended: rule_meta.docs.recommended,
+                fixable: rule_meta.fixable,
```
btmills, line 49: The fixable blurb isn't showing on e.g. [`arrow-body-style`](https://deploy-preview-15902--docs-eslint.netlify.app/rules/arrow-body-style/), perhaps because the `ruleCategories` macro [expects this to be a boolean](https://github.com/eslint/eslint/blob/901ce0f1e32ea1e9e10ce4d8b37c0d750007a3c5/docs/src/_includes/components/rule-categories.macro.html#L12), but `rule_meta.fixable` will be either `"code"` or `"whitespace"`.

btmills (changes requested):

Makefile.js
```diff
@@ -181,10 +181,14 @@ function generateFormatterExamples(formatterInfo, prereleaseVersion) {
  * @returns {void}
  */
 function generateRuleIndexPage() {
-    const outputFile = "../website/_data/rules.yml",
+    const legacyWebsiteOutputFile = "../website/_data/rules.yml",
+        docsSiteOutputFile = "docs/src/_data/rules.json",
+        docsSiteMetaOutputFile = "docs/src/_data/rules_meta.json",
```
mdjermanovic, line 186: I believe the changes in these two files will not be committed during the release. Our release process commits only [changelog](https://github.com/eslint/eslint-release/blob/198c7e700e501f2762f41396635b74fdb7d04e06/lib/release-ops.js#L381-L382) and [version](https://github.com/eslint/eslint-release/blob/198c7e700e501f2762f41396635b74fdb7d04e06/lib/release-ops.js#L385) (`npm version` creates a version commit and tag).

docs/src/_includes/layouts/doc.html
nzakas: Oops, good call.

Makefile.js
nzakas: Good call. Maybe we should update eslint-release to run an `npm build` command and then commit the result?

mdjermanovic: > Good call. Maybe we should update eslint-release to run an `npm build` command and then commit the result?

Maybe we could add this as an option in [`generateRelease()`](https://github.com/eslint/eslint-release/blob/198c7e700e501f2762f41396635b74fdb7d04e06/lib/release-ops.js#L365)? Some projects don't have a `build` script (for example, [`eslint-plugin-markdown`](https://github.com/eslint/eslint-plugin-markdown/blob/87c2b536fd80b15e134766d92b90048ae45cbe1f/package.json#L22-L32)), and some project have a `build` script that should not be run for this purpose (for example, [`espree`](https://github.com/eslint/espree/blob/c98a7e0be7e21f28a00868660d66bda9d323f1d3/package.json#L70)).

mdjermanovic: Alternatively, we could update and commit these files before running [`ReleaseOps.generateRelease()`](https://github.com/eslint/eslint/blob/ce035e5fac632ba8d4f1860f92465f22d6b44d42/Makefile.js#L267). A downside is that the commit would be included in the changelog.

Or add a `preversion` or `version` script ([npm version](https://docs.npmjs.com/cli/v8/commands/npm-version)). A downside is that this would bundle the changes into the version commit, I think.

nzakas: I’ll dig into this more.

mdjermanovic: > we can add the new files and amend the commit after we generate the release.

Version tag will still point to the original version commit, which won't be in the history of the `main` branch anymore. I think that's not a common practice.

nzakas: Hmm good point. I suppose the next logical step is just to update the tag, too.

nzakas:
Updated the git tag after modifying the commit and rebased on top of main.

Note: I noticed a UI bug related to the displaying of fixable/suggestions/recommended on rules. I've opened https://github.com/eslint/eslint/issues/15977 to track separately.

nzakas:
@btmills can you take a look at this to clear your review?

btmills (approved):
I commented on #15977 as well, but I'm seeing all three categories displaying correctly now!

Makefile.js
```diff
@@ -271,6 +292,10 @@ function generateRelease() {
     target.gensite();
     generateBlogPost(releaseInfo);
     commitSiteToGit(`v${releaseInfo.version}`);
+
+    echo("Updating commit with docs data");
+    exec("git add docs/src/_data && git commit --amend --no-edit");
+    exec(`git tag -f v${releaseInfo.version}`);
```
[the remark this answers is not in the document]
nzakas: I’ll have to check if the annotation remains after this or if it’s lost.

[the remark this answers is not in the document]
nzakas: It looks like everything is lost, so I updated to create a new annotated tag with the version as a message (to duplicate what `npm version` does).

mdjermanovic (approved):
LGTM, thanks!

PR #15926, opened 2022-05-26
chore: reduce cumulative layout shifts

amareshsm:
#### Prerequisites checklist

[ ] Documentation update
[ ] Add autofix to a rule
[ ] Add a CLI option
[ ] Add something to the core
[x] Other, please explain:
     Reduced CLS in new docs site.

Reduced CLS.

### DIFF
--- a/docs/src/_includes/components/docs-index.html
+++ b/docs/src/_includes/components/docs-index.html
@@ -19,7 +19,33 @@
             <path id="ham-bottom" d="M30,63 L70,63 Z" stroke="currentColor"></path>
         </svg>
     </button>
-    <ul class="docs-index__list" id="js-docs-index-list" data-open="false">
+    <noscript>
+        <style>
+            #js-docs-index-toggle{
+                display: none;
+            }
+        </style>
+        <button class="docs-index-toggle" aria-label="Index"  hidden>
+            Index
+            <svg width="20" height="20" viewBox="20 20 60 60">
+                <path id="ham-top" d="M30,37 L70,37 Z" stroke="currentColor"></path>
+                <path id="ham-middle" d="M30,50 L70,50 Z" stroke="currentColor"></path>
+                <path id="ham-bottom" d="M30,63 L70,63 Z" stroke="currentColor"></path>
+            </svg>
+        </button>
+    </noscript>
+
+    <ul class="docs-index__list" id="js-docs-index-list">
         {%- for entry in navPages %}{{ renderNavListItem(entry) }}{%- endfor -%}
     </ul>
+    <noscript>
+        <style>
+            #js-docs-index-list{
+                display: none;
+            }
+        </style>
+    <ul class="docs-index__list" >
+        {%- for entry in navPages %}{{ renderNavListItem(entry) }}{%- endfor -%}
+    </ul>
+</noscript>
 </nav>
--- a/docs/src/_includes/components/docs-index.html
+++ b/docs/src/_includes/components/docs-index.html
@@ -10,8 +10,8 @@
 </li>
 {%- endmacro %}
 
-<nav class="docs-index" id="docs-index">
-    <button class="docs-index-toggle" aria-label="Index" id="js-docs-index-toggle" hidden>
+<nav class="docs-index" id="docs-index" >
+    <button class="docs-index-toggle" aria-label="Index" id="js-docs-index-toggle">
         Index
         <svg width="20" height="20" viewBox="20 20 60 60">
             <path id="ham-top" d="M30,37 L70,37 Z" stroke="currentColor"></path>
@@ -19,7 +19,7 @@
             <path id="ham-bottom" d="M30,63 L70,63 Z" stroke="currentColor"></path>
         </svg>
     </button>
-    <ul class="docs-index__list" id="js-docs-index-list">
+    <ul class="docs-index__list" id="js-docs-index-list" data-open="false">
         {%- for entry in navPages %}{{ renderNavListItem(entry) }}{%- endfor -%}
     </ul>
 </nav>
--- a/docs/src/_includes/components/docs-index.html
+++ b/docs/src/_includes/components/docs-index.html
@@ -19,33 +19,7 @@
             <path id="ham-bottom" d="M30,63 L70,63 Z" stroke="currentColor"></path>
         </svg>
     </button>
-    <noscript>
-        <style>
-            #js-docs-index-toggle{
-                display: none;
-            }
-        </style>
-        <button class="docs-index-toggle" aria-label="Index"  hidden>
-            Index
-            <svg width="20" height="20" viewBox="20 20 60 60">
-                <path id="ham-top" d="M30,37 L70,37 Z" stroke="currentColor"></path>
-                <path id="ham-middle" d="M30,50 L70,50 Z" stroke="currentColor"></path>
-                <path id="ham-bottom" d="M30,63 L70,63 Z" stroke="currentColor"></path>
-            </svg>
-        </button>
-    </noscript>
-
     <ul class="docs-index__list" id="js-docs-index-list">
         {%- for entry in navPages %}{{ renderNavListItem(entry) }}{%- endfor -%}
     </ul>
-    <noscript>
-        <style>
-            #js-docs-index-list{
-                display: none;
-            }
-        </style>
-    <ul class="docs-index__list" >
-        {%- for entry in navPages %}{{ renderNavListItem(entry) }}{%- endfor -%}
-    </ul>
-</noscript>
 </nav>
--- a/docs/src/_includes/components/nav-version-switcher.html
+++ b/docs/src/_includes/components/nav-version-switcher.html
@@ -1,7 +1,7 @@
 <div class="version-switcher">
-    <noscript>
-            <a href="/versions/" class="switcher-fallback">Versions</a>
-    </noscript> 
+
+    <a href="/versions/" class="switcher-fallback">Versions</a>
+
     <div  role="region" class="switcher switcher--version" aria-labelledby="nav-version-switcher-label" id="nav-version-switcher">
         <span id="nav-version-switcher-label" hidden>Language Switcher</span>
         <div class="infobox visually-hidden" id="nav-version-infobox">
--- a/docs/src/_includes/components/nav-version-switcher.html
+++ b/docs/src/_includes/components/nav-version-switcher.html
@@ -1,6 +1,4 @@
 <div class="version-switcher">
-    <a href="/versions/" class="switcher-fallback">Versions</a>
-
     <div  role="region" class="switcher switcher--version" aria-labelledby="nav-version-switcher-label" id="nav-version-switcher">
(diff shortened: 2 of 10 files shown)


docs/src/_includes/components/version-switcher.html
```diff
@@ -1,7 +1,7 @@
 <div class="version-switcher">
-    <a href="/versions/" class="switcher-fallback">Versions</a>
+    <a hidden href="/versions/" class="switcher-fallback">Versions</a>
 
-    <div hidden role="region" class="switcher switcher--version" aria-labelledby="version-switcher-label" id="version-switcher">
+    <div  role="region" class="switcher switcher--version" aria-labelledby="version-switcher-label" id="version-switcher">
```
mdjermanovic, line 4: The fallback link is for users who have disabled JavaScript in their browsers. We have a [script](https://github.com/eslint/eslint/blob/10249ad1a961463b6b347be71c074951ab8e2652/docs/src/assets/js/main.js#L131-L154) that hides fallback links and unhides select elements that need JavaScript to work (execution of the script by itself means that JavaScript is enabled).

If the fallback link is initially hidden and the select element is not, these users will see only the select element which doesn't work without JavaScript.

docs/src/_includes/components/nav-version-switcher.html
```diff
@@ -1,7 +1,7 @@
 <div class="version-switcher">
     <a href="/versions/" class="switcher-fallback">Versions</a>
 
-    <div hidden role="region" class="switcher switcher--version" aria-labelledby="nav-version-switcher-label" id="nav-version-switcher">
+    <div  role="region" class="switcher switcher--version" aria-labelledby="nav-version-switcher-label" id="nav-version-switcher">
```
mdjermanovic, line 4: This fixes a layout shift that happens only on mobile?

docs/src/_includes/components/version-switcher.html
amareshsm: oh okay 👍  I should have checked it.

docs/src/_includes/components/nav-version-switcher.html
amareshsm: yes

mdjermanovic: We'll have the same problem here, users without JavaScript will see a select element that doesn't work.

This is inside [`nav-panel`](https://github.com/eslint/eslint/blob/10249ad1a961463b6b347be71c074951ab8e2652/docs/src/_includes/components/navigation.html#L12), which itself causes a layout shift on mobile. Would it be better to set `data-open="false"` initially on `nav-panel`?

mdjermanovic: > This is inside [`nav-panel`] Would it be better to set `data-open="false"` initially on `nav-panel`?

This actually won't work well when JavaSript is disabled because there will be no way to show the panel.

docs/src/_includes/components/docs-index.html
```diff
@@ -10,16 +10,16 @@
 </li>
 {%- endmacro %}
 
-<nav class="docs-index" id="docs-index">
-    <button class="docs-index-toggle" aria-label="Index" id="js-docs-index-toggle" hidden>
```
mdjermanovic, line 14: This is initially hidden because users without JavaScript shouldn't see this. For those with JavaScript, a script will unhide this.

docs/src/_includes/components/docs-index.html
```diff
@@ -10,16 +10,16 @@
...
+    <button class="docs-index-toggle" aria-label="Index" id="js-docs-index-toggle">
         Index
         <svg width="20" height="20" viewBox="20 20 60 60">
             <path id="ham-top" d="M30,37 L70,37 Z" stroke="currentColor"></path>
             <path id="ham-middle" d="M30,50 L70,50 Z" stroke="currentColor"></path>
             <path id="ham-bottom" d="M30,63 L70,63 Z" stroke="currentColor"></path>
         </svg>
     </button>
-    <ul class="docs-index__list" id="js-docs-index-list">
+    <ul class="docs-index__list" id="js-docs-index-list" data-open="false">
```
mdjermanovic, line 22: This is the opposite, users without JavaScript should see this, so we can't hide it initially.

mdjermanovic:
> Edit: Lighthouse reports for desktop version

What's the result for mobile? I'm getting CLS 0.9 on new.eslint.org/docs

docs/src/_includes/components/nav-version-switcher.html
amareshsm: @mdjermanovic I tried the same `<noscript><style>...</style></noscript>`

mdjermanovic:
This is what I'm getting locally with the latest `main`, for the docs homepage:

Desktop:

image

Mobile:

image

amareshsm:
> This is what I'm getting locally with the latest `main`, for the docs homepage:
> 
> Desktop:
> 
> image
> 
> Mobile:
> 
> image

Exactly the same results I am getting in the `main` branch. 


#### After changes (in `reduce-layout-shifts` branch):
URL - http://localhost:2023/docs/

Desktop:
image

Mobile:
image

docs/src/_includes/components/docs-index.html
```diff
@@ -10,16 +10,42 @@
...
+            }
+        </style>
+        <button class="docs-index-toggle" aria-label="Index"  hidden>
+            Index
+            <svg width="20" height="20" viewBox="20 20 60 60">
+                <path id="ham-top" d="M30,37 L70,37 Z" stroke="currentColor"></path>
+                <path id="ham-middle" d="M30,50 L70,50 Z" stroke="currentColor"></path>
+                <path id="ham-bottom" d="M30,63 L70,63 Z" stroke="currentColor"></path>
+            </svg>
+        </button>
```
mdjermanovic, line 35: This looks like a dead code. It's hidden, and since it is in `<noscript>` there will be no script run to unhide it.

docs/src/_includes/components/docs-index.html
```diff
@@ -10,16 +10,42 @@
...
+    <noscript>
+        <style>
+            #js-docs-index-list{
+                display: none;
+            }
+        </style>
+    <ul class="docs-index__list" >
+        {%- for entry in navPages %}{{ renderNavListItem(entry) }}{%- endfor -%}
+    </ul>
+</noscript>
```
mdjermanovic, line 50: Why do we need another list?

mdjermanovic: I think I got it, the other one should be hidden on mobile until the script sets `data-open = true`, while this one should be visible from the start?

mdjermanovic: Instead of duplicating lists, maybe we could do `#js-docs-index-list{display: block !important}` in the noscript style?

docs/src/_includes/components/docs-index.html
```diff
@@ -10,16 +10,42 @@
...
             <path id="ham-middle" d="M30,50 L70,50 Z" stroke="currentColor"></path>
             <path id="ham-bottom" d="M30,63 L70,63 Z" stroke="currentColor"></path>
         </svg>
     </button>
+    <noscript>
+        <style>
+            #js-docs-index-toggle{
+                display: none;
+            }
+        </style>
```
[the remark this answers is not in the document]
amareshsm: sure. we can move to head.

docs/src/_includes/components/docs-index.html
amareshsm: yes. We don't need this.

mdjermanovic:
Now I noticed we have [`class="no-js"`](https://github.com/eslint/eslint/blob/10249ad1a961463b6b347be71c074951ab8e2652/docs/src/_includes/layouts/base.njk#L2) set on the html tag, and a [script](https://github.com/eslint/eslint/blob/10249ad1a961463b6b347be71c074951ab8e2652/docs/src/_includes/layouts/base.njk#L38-L42) that replaces it with `enhanced`, so that seems to be the intended way to do this.

Can you try with that instead of `noscript`? (e.g., `.no-js #js-docs-index-list { display: block !important }`).

`.enhanced` could hide the fallback links.

I _think_ we should remove `type = "module"` from the script because modules are deferred, while we probably need that to execute as soon as possible.

docs/src/_includes/components/navigation.html
```diff
@@ -22,4 +22,19 @@
...
+                    <a href="{{ item.url }}" {% if item.text == 'Docs' %} aria-current="true"{% endif %}>{{ item.text }}</a>
+                </li>
+                {% endfor %}
+            </ul>
+        </div>
+        <div class="mobile-only">
+            {% include 'components/nav-version-switcher-noscript.html' %}
+            {% include 'components/nav-search.html' %}
+        </div>
+    </noscript>
```
mdjermanovic, line 39: Can we avoid duplicating the docs nav panel, like we avoided duplicating the docs index list? (https://github.com/eslint/eslint/pull/15926#discussion_r883910216)

amareshsm: yes, done.

docs/src/_includes/components/nav-version-switcher.html
```diff
@@ -1,7 +1,8 @@
 <div class="version-switcher">
-    <a href="/versions/" class="switcher-fallback">Versions</a>
-
-    <div hidden role="region" class="switcher switcher--version" aria-labelledby="nav-version-switcher-label" id="nav-version-switcher">
+    <noscript>
+            <a href="/versions/" class="switcher-fallback">Versions</a>
+    </noscript>
```
mdjermanovic, line 4: Could we use the `enhanced` class name (the one that replaces `no-js` on html) to hide all fallback links, and then we won't need to use `<noscript>`?

Something like `.enhanced .switcher-fallback { display: none !important }` in the `<style>` we added to `<head>`.

amareshsm: yes, we can use the `enhanced` class instead of `<noscript>`.

docs/src/_includes/layouts/base.njk
```diff
@@ -34,13 +34,27 @@
...
+            display: none !important 
+        }
+        .no-js #js-docs-index-toggle, .no-js #nav-toggle, .no-js #version-switcher, .no-js #nav-version-switcher {
+            display: none;
+        }
+        .no-js #js-docs-index-list {
+            display: block !important
+        }
+        .no-js #nav-panel {
+            display:grid !important;
```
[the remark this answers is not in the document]
amareshsm: yes, done.

mdjermanovic (approved):
LGTM, thanks!

This looks noticeably better than before, especially on mobile. And the lighthouse report shows the improvement:

CLS before: 0.9
CLS after: < 0.1

I would like more reviews from other team members before merging.

mdjermanovic:
I converted this to draft since we first need to sync these files with the latest version of the prototype (https://github.com/eslint/eslint/pull/15944), as we noticed that the initial version in this repo was not the latest one from the prototype.

amareshsm:
Closing this in favour of #16047

PR #15976, opened 2022-06-08
feat!: allow changing warnIgnored via CLI

domnantas:
#### Prerequisites checklist

[ ] Documentation update
[ ] Add autofix to a rule
[x] Add a CLI option
[ ] Add something to the core
[ ] Other, please explain:

This is related to https://github.com/eslint/rfcs/pull/90 and https://github.com/eslint/eslint/issues/15010

- Added `--no-warning-on-ignored-files` CLI option
- Removed inconsistent `warnIgnored` defaults, replaced with a single `warnIgnored: true` default
- Updated warning messages to inform users about the ability to suppress the warning
- Added tests for `CLIEngine.executeOnText`, `CLIEngine.executeOnFiles`, `ESLint.lintText`, `ESLint.lintFiles` and CLI

- Are there any tests missing?
- Making defaults consistent is technically a breaking change. What should be done in this case?

### DIFF
--- a/lib/cli-engine/cli-engine.js
+++ b/lib/cli-engine/cli-engine.js
@@ -297,11 +297,11 @@ function createIgnoreResult(filePath, baseDir) {
     const isInNodeModules = baseDir && path.relative(baseDir, filePath).startsWith("node_modules");
 
     if (isHidden) {
-        message = "File ignored by default.  Use a negated ignore pattern (like \"--ignore-pattern '!<relative/path/to/filename>'\") to override.";
+        message = "File ignored by default.  Use a negated ignore pattern (like \"--ignore-pattern '!<relative/path/to/filename>'\") to override. Use \"--no-warning-on-ignored-files\" to suppress this warning.";
     } else if (isInNodeModules) {
-        message = "File ignored by default. Use \"--ignore-pattern '!node_modules/*'\" to override.";
+        message = "File ignored by default. Use \"--ignore-pattern '!node_modules/*'\" to override. Use \"--no-warning-on-ignored-files\" to suppress this warning.";
     } else {
-        message = "File ignored because of a matching ignore pattern. Use \"--no-ignore\" to override.";
+        message = "File ignored because of a matching ignore pattern. Use \"--no-ignore\" to override. Use \"--no-warning-on-ignored-files\" to suppress this warning.";
     }
 
     return {
@@ -759,7 +759,8 @@ class CLIEngine {
                 cache,
                 cwd,
                 fix,
-                reportUnusedDisableDirectives
+                reportUnusedDisableDirectives,
+                warnIgnored
             }
         } = internalSlotsMap.get(this);
         const results = [];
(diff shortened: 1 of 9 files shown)


lib/eslint/eslint.js
```diff
@@ -568,7 +573,7 @@ class ESLint {
         }
         const {
             filePath,
-            warnIgnored = false,
```
domnantas, line 571: Removing the inconsistent default, this will fall back to `processOptions` defaults

lib/eslint/eslint.js
```diff
@@ -168,6 +169,7 @@ function processOptions({
     resolvePluginsRelativeTo = null, // ← should be null by default because if it's a string then it suppresses RFC47 feature.
     rulePaths = [],
     useEslintrc = true,
+    warnIgnored = true,
```
domnantas, line 172: This will override `default-cli-options.js`, just like all of the options above this line.

lib/cli-engine/cli-engine.js
```diff
@@ -897,7 +901,7 @@ class CLIEngine {
         // Clear the last used config arrays.
         lastConfigArrays.length = 0;
         if (resolvedFilename && this.isPathIgnored(resolvedFilename)) {
-            if (warnIgnored) {
+            if (warnIgnored ?? engineWarnIgnored) {
```
domnantas, line 904: Using nullish coalescing here because `engineWarnIgnored` should only be used when `warnIgnored` is null or undefined

lib/cli-engine/cli-engine.js
```diff
@@ -785,7 +786,9 @@ class CLIEngine {
         // Iterate source code files.
         for (const { config, filePath, ignored } of fileEnumerator.iterateFiles(patterns)) {
             if (ignored) {
-                results.push(createIgnoreResult(filePath, cwd));
+                if (warnIgnored) {
+                    results.push(createIgnoreResult(filePath, cwd));
+                }
                 continue;
             }
```
domnantas, line 793: This might seem odd, but
if `ignored = false, warnIgnored = false` -> lint
if `ignored = false, warnIgnored = true` -> lint
if `ignored = true, warnIgnored = false` -> do not show warning, and do not lint
if `ignored = true, warnIgnored = true` -> show warning, do not lint

nzakas:
Converting to a draft until the RFC is approved.

markoenix (commented):
I tried on iphone7

Faithfinder:
> Oops! It looks like we lost track of this pull request. What do we want to do here? This pull request will auto-close in 7 days without an update.

Sorry if I'm spamming, but I'd _love_ for this to be merged, so I'm trying to unstale it. Sorry again if that's not the right way!

domnantas:
I think it is ok to close this. The [RFC](https://github.com/eslint/rfcs/pull/90) is still in progress and the PR has accrued a lot of conflicts. It can still be used as an example in the future.

Previously `warnIgnored` default was `true` in some places, `false` in other places. This PR aimed to unify them and make `warnIgnored: true` default everywhere. However, this is considered a breaking change, so we can only include it in a major version release. _However_, ESLint is going though a big change in its config system called FlatESLint (read more about it in this [great blog post](https://eslint.org/blog/2022/08/new-config-system-part-1/) by Nicholas), therefore it is unwise introducing such small changes into its scope.

This doesn't mean this issue will not be fixed, we will just need to rely on workarounds a little longer.

Issue #15726, opened 2022-03-25
Consider using Renovate


nzakas:
This app helps keep dependencies up to date:
https://github.com/marketplace/renovate

What do we think about including this in the repo?

aladdin-add:
just a note: the default commit tag is a liitle different than ours. we may need update `eslint-release` to allow the format `chore(deps): xxx`.

bmish:
Very much in favor of using an automatic dependency updater.

I do like Renovate, although [Dependabot](https://docs.github.com/en/code-security/dependabot/working-with-dependabot/keeping-your-actions-up-to-date-with-dependabot) has become my default automatic dependency updater since GitHub acquired and integrated it. Any reason to choose one over the other?

ljharb:
Renovate is awesome and far more configurable, and dependabot imo is junk, altho it’s slowly getting better over time.

You can configure any commit message convention you like.

nzakas:
I just can’t stand the millions of one-off dependabot update PRs. I’m open to anything that gets us out of that situation.

nzakas:
I’m testing Renovate on my own repos. Will report back with feedback.

nzakas:
**TSC Summary:** This issue proposes setting up Renovate on ESLint repos. Renovate prefers to pin all dependencies and send pull requests to update each dependency individually. It can be a bit noisy, but it would ensure that we are regularly updating dependencies, which is something we don't do very well right now.

**TSC Agenda:** Should we enable Renovate on the org?

ljharb:
Note you can customize the config to group updates if you like, much more granularly than any other alternative I’m aware of.

btmills:
We accepted this change in the [2023-02-09 TSC meeting](https://github.com/eslint/tsc-meetings/blob/main/notes/2023/2023-02-09.md#consider-using-renovate). I'm working on `release-please` first and will try to get to this next, but anyone else who's interested should feel free to take it before me!

rarkins:
Hi, the Renovate team would be happy to help you get onboarded.

Here's what I recommend:
1. An admin of the `eslint` org should go to https://github.com/apps/renovate and Install into this org. Use "Select repositories" and select this repo only
2. You will then see a single onboarding Pull Request like this: https://github.com/renovate-reproductions/eslint/pull/1
[...]

PR #15985, opened 2022-06-11
chore: Included githubactions in the dependabot config

naveensrinivasan:
This should help with keeping the GitHub actions updated on new releases. This will also help with keeping it secure.

Dependabot helps in keeping the supply chain secure https://docs.github.com/en/code-security/dependabot

GitHub actions up to date https://docs.github.com/en/code-security/dependabot/working-with-dependabot/keeping-your-actions-up-to-date-with-dependabot

https://github.com/ossf/scorecard/blob/main/docs/checks.md#dependency-update-tool
Signed-off-by: naveen <172697+naveensrinivasan@users.noreply.github.com>

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,6 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"


snitin315 (changes requested):
Thanks for the PR. We already have an open issue https://github.com/eslint/eslint/issues/15726 where we are evaluating which dependency management tool we should use.

nzakas:
We can include this for now. We can always switch later

PR #16063, opened 2022-06-28
feat: add deprecation warnings for legacy API in `RuleTester`

snitin315:
#### Prerequisites checklist

[ ] Documentation update
[ ] Add autofix to a rule
[ ] Add a CLI option
[x] Add something to the core
[ ] Other, please explain: 

Resolve https://github.com/eslint/eslint/pull/15761
Refers https://github.com/eslint/eslint/issues/14709

Prepares for upcoming breaking changes in ESLint v9 to disallow function-style rules and rules with options that are missing a schema.

Adds the deprecation warnings mentioned in the [RFC](https://github.com/eslint/rfcs/tree/main/designs/2021-schema-object-rules#backwards-compatibility-analysis) (https://github.com/eslint/rfcs/pull/85).

These deprecation warnings will help alert plugin authors that they need to update their rules. They are only visible when running rule tests and not to end-users who just use the rules.

> This lint rule can assist with the conversion: https://github.com/not-an-aardvark/eslint-plugin-eslint-plugin/blob/main/docs/rules/require-meta-schema.md");

I'm not sure if we should add the above line in the deprecation message. Two links in the log can be overwhelming. Maybe we can mention this somewhere in our docs here - https://eslint.org/docs/latest/developer-guide/working-with-rules


tests/fixtures/testers/rule-tester/no-var.js
```diff
@@ -12,7 +12,8 @@
 module.exports = {
 
     meta: {
-        fixable: "code"
+        fixable: "code",
+        shema: []
```
snitin315, line 16: Otherwise, we will see a deprecation log when using this fixture.

lib/rule-tester/rule-tester.js
```diff
@@ -521,6 +521,21 @@ class RuleTester {
             ].concat(scenarioErrors).join("\n"));
         }
 
+        if (typeof rule === "function") {
+            util.deprecate(
+                () => {},
+                `"${ruleName}" rule is using the deprecated function-style format and will stop working in ESLint v9. Please use object-style format: https://eslint.org/docs/developer-guide/working-with-rules`,
+                "DEP_ESLINT_LEGACY_RULE_API"
```
mdjermanovic, line 528: According to `util.deprecate` docs, if we specify a deprecation code, then the warning will be emitted only once for that code, while I think we want the warning to be emitted for each rule that is using the function-style format.

lib/rule-tester/rule-tester.js
```diff
@@ -521,6 +521,21 @@ class RuleTester {
...
+            )();
+        }
+
+        if (typeof rule === "object" && rule.meta && typeof rule.meta.schema === "undefined") {
+            util.deprecate(
+                () => {},
+                `"${ruleName}" rule is using the object-style format but is missing the "meta.schema" property. Please add a schema: https://eslint.org/docs/developer-guide/working-with-rules#options-schemas`,
+                "DEP_ESLINT_MISSING_RULE_SCHEMA"
+            )();
+        }
```
mdjermanovic, line 538: This should be triggered by a test case that specifies `options` property with a non-empty array value. Rules that don't have options are okay because the default schema will be `schema: []` which is correct for those rules, so we shouldn't emit the warning for them. A test case with `options` is a way to detect that the rule has options.

mdjermanovic: I think the message should be something like:

```
`"${ruleName}" rule has options but is missing the "meta.schema" property and will stop working in ESLint v9. Please add a schema: https://eslint.org/docs/developer-guide/working-with-rules#options-schemas`
```

mdjermanovic: Also, I think that https://github.com/eslint/eslint/pull/16063#discussion_r908316909 applies here as well, the warning should be emitted for each rule.

snitin315: @mdjermanovic yes, but we may not want to emit a warning multiple times for the same rule. So, let's add ruleName to the code as well to make it unique?

Screenshot 2022-06-29 at 7 26 33 AM

snitin315: I've updated the PR, please check the new implementation.

tests/fixtures/testers/rule-tester/no-var.js
snitin315: We have some more deprecation logs on `npm run test` now, a few of these might be intentional depending upon the test case. Should we go through them and fix it here or in a separate PR?

Screenshot 2022-06-29 at 8 40 39 AM

lib/rule-tester/rule-tester.js
mdjermanovic: > but we may not want to emit a warning multiple times for the same rule.

We could emit the warning once per `run()`. For example, we can create an instance in `run()`:

```js
const emitMissingSchemaWarning = util.deprecate(
    () => {},
    `"${ruleName}" rule has options but is missing the "meta.schema" property and will stop working in ESLint v9. Please add a schema: https://eslint.org/docs/developer-guide/working-with-rules#options-schemas`
);
```

and then call it in `runRuleForItem()`:

```js
if (/* ... */) {
    emitMissingSchemaWarning();
}
```

Per the `util.deprecate` docs, only the first call will emit the warning.

Or, we could add a flag and check if the warning was already emitted.

mdjermanovic: > Or, we could add a flag and check if the warning was already emitted.

In this case, maybe we could directly use [process.emitWarning](https://nodejs.org/api/process.html#processemitwarningwarning-type-code-ctor)

tests/fixtures/testers/rule-tester/no-var.js
mdjermanovic: If it doesn't seem that a test case is specifically testing how RuleTester works with function-style rules, then I think it's okay to convert the rule that the test case uses to object style.

tests/lib/rule-tester/rule-tester.js
```diff
@@ -2295,6 +2296,123 @@ describe("RuleTester", () => {
...
+            };
+
+            ruleTester.run("rule-with-no-options", ruleWithEmptySchema, {
+                valid: [],
+                invalid: [{ code: "var foo = bar;", errors: 1 }]
+            });
+
+            assert.strictEqual(spy.callCount, 0, "never calls `util.deprecate`");
+        });
+    });
```
mdjermanovic, line 2414: It would be nice to add tests for more cases where the warnings should not be emitted:

* When the rule is an object-style rule, the legacy rule API warning is not emitted.
* When the rule has `meta.schema` and there are test cases with `options`, the missing schema warning is not emitted.
* When the rule does not have `meta`, but there are no test cases with `options`, the missing schema warning is not emitted.
* When the rule has `meta` without `meta.schema`, but there are no test cases with `options`, the missing schema warning is not emitted.
* When the rule has `meta` without `meta.schema`, and some test cases have `options` property but it's an empty array, the missing schema warning is not emitted.

mdjermanovic:
I checked all tests that are using function-style fixture rules, and it seems that only [this one](https://github.com/eslint/eslint/blob/eee43067f635c0ec3b61e416f47849029d12268d/tests/lib/rule-tester/rule-tester.js#L1755-L1778) is specifically testing how RuleTester handles function-style rules, so I think we should convert all other fixture rules to object-style now. That will avoid logging deprecation warnings when we run `npm test`, and also ensure that RuleTester functionalities are tested with object-style rules, which seems more relevant than testing with legacy function-style rules.

Those fixure rules are:

* [no-eval](https://github.com/eslint/eslint/blob/main/tests/fixtures/testers/rule-tester/no-eval.js)
* [no-test-settings](https://github.com/eslint/eslint/blob/main/tests/fixtures/testers/rule-tester/no-test-settings.js)
* [no-test-global](https://github.com/eslint/eslint/blob/main/tests/fixtures/testers/rule-tester/no-test-global.js)
* [no-test-filename](https://github.com/eslint/eslint/blob/main/tests/fixtures/testers/rule-tester/no-test-filename)
* [no-invalid-args](https://github.com/eslint/eslint/blob/main/tests/fixtures/testers/rule-tester/no-invalid-args.js) (we can also add `meta.shema` to this one).
* [no-invalid-schema](https://github.com/eslint/eslint/blob/main/tests/fixtures/testers/rule-tester/no-invalid-schema.js)
* [no-schema-violation](https://github.com/eslint/eslint/blob/main/tests/fixtures/testers/rule-tester/no-schema-violation.js)
* [modify-ast](https://github.com/eslint/eslint/blob/main/tests/fixtures/testers/rule-tester/modify-ast.js)
* [modify-ast-at-first](https://github.com/eslint/eslint/blob/main/tests/fixtures/testers/rule-tester/modify-ast-at-first.js)
* [modify-ast-at-last](https://github.com/eslint/eslint/blob/main/tests/fixtures/testers/rule-tester/modify-ast-at-last.js)

snitin315:
@mdjermanovic Thank you for investigating this, I'll update the PR with your suggestions.

mdjermanovic:
> only [this one](https://github.com/eslint/eslint/blob/eee43067f635c0ec3b61e416f47849029d12268d/tests/lib/rule-tester/rule-tester.js#L1755-L1778) is specifically testing how RuleTester handles function-style rules

As for this one, I think we can just remove that test case + two test cases right before it; in total, we can now remove [these 3 test cases](https://github.com/eslint/eslint/blob/eee43067f635c0ec3b61e416f47849029d12268d/tests/lib/rule-tester/rule-tester.js#L1713-L1778).

These 3 test cases were added to check RuleTester-only errors we introduced in v7.6.0 (https://github.com/eslint/eslint/commit/ecb2b7343a0d14fb57d297a16be6c1b176fb3dbf) as a notice about future Linter errors that were planned and eventually implemented in v8.0.0 (https://github.com/eslint/eslint/commit/4a7aab7d4323ff7027eebca709d4e95a9aaa80bc). Since these are now Linter errors, there's no need to test them in RuleTester.

lib/rule-tester/rule-tester.js
```diff
@@ -305,6 +305,36 @@ function getCommentsDeprecation() {
...
+        util.deprecate(
+            () => {},
+            `"${ruleName}" rule is using the deprecated function-style format and will stop working in ESLint v9. Please use object-style format: https://eslint.org/docs/developer-guide/working-with-rules`
+        )();
+    }
+}
+
+/**
+ * Emit a deprecation warning if rule has options but is missing the "meta.schema" property
+ * @param {string} ruleName Name of the rule.
```
harish-sethuraman, line 325:
```suggestion
 * @param {string} ruleName of the rule.
```

tests/fixtures/testers/rule-tester/no-var.js
harish-sethuraman, line 16: isn't this `schema` ?

snitin315: Yep, typo. Thanks!

tests/lib/rule-tester/rule-tester.js
```diff
@@ -2295,6 +2296,123 @@ describe("RuleTester", () => {
...
+                invalid: [
+                    { code: "var foo = bar;", errors: 1 }
+                ]
+            });
+
+            assert.strictEqual(spy.callCount, 1, "calls `util.deprecate()` once");
+            assert.strictEqual(
+                spy.getCall(0).args[1],
+                "\"function-style-rule\" rule is using the deprecated function-style format and will stop working in ESLint v9. Please use object-style format: https://eslint.org/docs/developer-guide/working-with-rules"
+            );
```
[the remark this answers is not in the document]
snitin315: Yes, using the `process.emitWarning` will simplify be the code here.

tests/lib/rule-tester/rule-tester.js
snitin315: I will add these cases 👍🏻

tests/lib/rule-tester/rule-tester.js
```diff
@@ -2295,6 +2228,123 @@ describe("RuleTester", () => {
         });
     });
 
+    describe("deprecations", () => {
+        let spy;
+
+        beforeEach(() => {
+            spy = sinon.spy(process, "emitWarning");
```
mdjermanovic, line 2235: spy will call the original `process.emitWarning` so we'll get warning logs on `npm test`. Can we use something else that doesn't call the original function, for example `sinon.stub`?

tests/lib/rule-tester/rule-tester.js
```diff
@@ -2295,6 +2228,123 @@ describe("RuleTester", () => {
...
+            ruleTester.run("function-style-rule", functionStyleRule, {
+                valid: [],
+                invalid: [
+                    { code: "var foo = bar;", errors: 1 }
+                ]
+            });
+
+            assert.strictEqual(spy.callCount, 1, "calls `process.emitWarning()` once");
+            assert.strictEqual(
+                spy.getCall(0).args[0],
```
bmish, line 2266: Instead of using `spy.getCall(0).args[0]`, you should actually use `assert.deepEqual(spy.getCall(0).args, ...)` to ensure we test ALL the args and not just the first one.

lib/rule-tester/rule-tester.js
```diff
@@ -578,6 +611,15 @@ class RuleTester {
 
             if (hasOwnProperty(item, "options")) {
                 assert(Array.isArray(item.options), "options must be an array");
+                if (
+                    item.options.length > 0 &&
+                    typeof rule === "object" &&
+                    (
+                        !rule.meta || (rule.meta && typeof rule.meta.schema === "undefined")
```
bmish, line 618: Shouldn't we also check for `null`?

snitin315: @bmish like this?
```suggestion
                        !rule.meta || (rule.meta && (typeof rule.meta.schema === "undefined" || rule.meta.schema === null))
```

tests/lib/rule-tester/rule-tester.js
snitin315: Done.

tests/lib/rule-tester/rule-tester.js
snitin315: `sinon.stub` works 👍🏻

lib/rule-tester/rule-tester.js
bmish: Yes because both null/undefined indicate the lack of a schema.

lib/rule-tester/rule-tester.js
```diff
@@ -305,6 +305,36 @@ function getCommentsDeprecation() {
...
+        );
+    }
+}
+
+/**
+ * Emit a deprecation warning if rule has options but is missing the "meta.schema" property
+ * @param {string} ruleName Name of the rule.
+ * @returns {void}
+ */
+function emitMissingSchemaWarning(ruleName) {
```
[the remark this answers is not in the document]
snitin315: We can refactor, we will just have to create a unique flag every time. If we set `emitWarning[`warned-${ruleName}`] = true` in the helper it may suppress another warning coming from the same rule.

I suppose this can be the helper"

```js
// API
function emitWarning(flag, message) {
    if (!emitWarning[`warned-${flag}`]) {
        emitWarning[`warned-${flag}`] = true;
        process.emitWarning(
            `"${ruleName}" rule has options but is missing the "meta.schema" property and will stop working in ESLint v9. Please add a schema: https://eslint.org/docs/developer-guide/working-with-rules#options-schemas`,
            "DeprecationWarning"
        );
    }
}

// Usage
emitWarning(`legacy-rule-api-${ruleName}`, 'some message...');
emitWarning(`missing-schema-${ruleName}`, 'some message...');
```

lib/rule-tester/rule-tester.js
snitin315: CI is failing now after adding the check🧐

snitin315: resolved.

lib/rule-tester/rule-tester.js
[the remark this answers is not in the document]
snitin315: I feel we can add this helper later when we have requirements for adding warnings in other files too. 
/cc @mdjermanovic WDYT?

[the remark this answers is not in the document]
mdjermanovic: > I feel we can add this helper later when we have requirements for adding warnings in other files too. /cc @mdjermanovic WDYT?

I agree. We could consider refactoring when we add more warnings.

mdjermanovic:
ESLint currently allows `schema` as a top-level property in object-style rules, and it's treated the same as `meta.schema`.

```js
// rule
module.exports = {

    // this will be treated as the options schema
    schema: {
        /* ... */
    },

    create(context) {
        /* ... */
    }
};
```

This is most likely an oversight. [`getRuleOptionsSchema`](https://github.com/eslint/eslintrc/blob/bdce01aad747393d13cd1ae17d8280e54c9c5191/lib/shared/config-validator.js#L50) gets schema from the given rule by `rule.schema || rule.meta && rule.meta.schema`. I believe `rule.schema` was only intended to get schema from function-style rules, but it works for object-style rules as well.

This PR does not use `getRuleOptionsSchema` (it gets the schema directly from `rule.meta.schema`) so it will log the warning as if the schema is not defined. I think this behavior is correct and that we want to drop support for top-level `schema` properties in favor of `meta.schema`, but this wasn't mentioned in the RFC. @eslint/eslint-tsc thoughts?

nzakas:
> I think this behavior is correct and that we want to drop support for top-level schema properties in favor of meta.schema, but this wasn't mentioned in the RFC.

I agree. 👍

bmish:
Good catch, I'd very much like to drop support for top-level `schema` in this coming major release given that it wasn't intended nor documented for object-style rules.

mdjermanovic (approved):
LGTM, thanks!

tests/lib/rule-tester/rule-tester.js
```diff
@@ -2295,6 +2228,200 @@ describe("RuleTester", () => {
         });
     });
 
+    describe("deprecations", () => {
+        let processStub;
+        const ruleWithNoSchema = {
```
bmish, line 2233: There needs to be tests with `schema: null` and `schema: undefined` as well.

nzakas:
Should we also duplicate these changes in FlatRuleTester?

tests/lib/rule-tester/rule-tester.js
snitin315: Added in https://github.com/eslint/eslint/pull/16063/commits/b23b8c13c2ad731b2b30d810b2928d2c84c2b035

bmish:
@nzakas good point, it seems like we have to duplicate these changes/tests in FlatRuleTester, otherwise we'll have divergence and these changes will be lost.

mdjermanovic:
> Should we also duplicate these changes in FlatRuleTester?

I think it isn't necessary since we are planning https://github.com/eslint/rfcs/pull/85 for ESLint v9, and FlatRuleTester will not replace RuleTester before ESLint v9. When we implement https://github.com/eslint/rfcs/pull/85 in linter/config, these deprecation warnings will become errors thrown by linter/config.

bmish (approved):
Thanks for your work on this!

nzakas:
Fair enough.

Issue #16128, opened 2022-07-13
Bug: Anchor link missing on headers


bmish:
### URL(s)

Any page on the new ESLint website. Example: https://eslint.org/docs/latest/developer-guide/working-with-rules

### What did you do?

Hover over section header. For example, the "Rule Basics" section header.

### What did you expect to happen?

An anchor link icon to appear that can be used to directly link to a particular header inside a page. The direct section link https://eslint.org/docs/latest/developer-guide/working-with-rules#rule-basics can only be found in the table of contents now.

### What actually happened?

Nothing.

### Participation

- [ ] I am willing to submit a pull request for this issue.

### Additional comments

From my memory, this feature was present on the old website, but is missing on the new website. It is also a standard convenience feature on GitHub markdown pages and many other documentation sites.

harish-sethuraman:
I believe we were using https://www.npmjs.com/package/anchor-js to add anchors to the heading elements in the archive website. If we want to do the same i'm up for grabs to change it in all docs site :)

nzakas:
@harish-sethuraman yes, please go ahead. 🙏

PR #16134, opened 2022-07-15
docs: add anchors to headings inside docs content

harish-sethuraman:
#### Prerequisites checklist

Fixes https://github.com/eslint/eslint/issues/16128

[x] Documentation update
[ ] Add autofix to a rule
[ ] Add a CLI option
[ ] Add something to the core
[ ] Other, please explain:

Added anchor-js library and add anchors for headings inside docs

check if any other element is getting anchors. I verified that table of content was h2 and was getting the anchor so I've move it to be a paragraph text and added font-weight alone to match the styling. If you are aware of any other elements like that need help. I verified once 😅  and everything seems to work fine

### DIFF
--- a/docs/src/_includes/components/docs-toc.html
+++ b/docs/src/_includes/components/docs-toc.html
@@ -1,6 +1,6 @@
 <nav class="docs-toc c-toc" aria-labelledby="js-toc-label">
     {%- if all_content | toc | safe | length -%}
-    <p class="c-toc__label" id="js-toc-label">Table of Contents</p>
+    <h2 class="c-toc__label" id="js-toc-label">Table of Contents</h2>
     <div class="c-toc__panel" id="js-toc-panel">
         {{ all_content | toc | safe }}
     </div>
--- a/docs/src/_includes/components/docs-toc.html
+++ b/docs/src/_includes/components/docs-toc.html
@@ -1,6 +1,6 @@
 <nav class="docs-toc c-toc" aria-labelledby="js-toc-label">
     {%- if all_content | toc | safe | length -%}
-    <h2 class="c-toc__label" id="js-toc-label">Table of Contents</h2>
+    <p class="c-toc__label" id="js-toc-label">Table of Contents</p>
     <div class="c-toc__panel" id="js-toc-panel">
         {{ all_content | toc | safe }}
     </div>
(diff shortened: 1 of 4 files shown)


docs/src/_includes/layouts/base.html
```diff
@@ -121,6 +121,7 @@
     <script type="module" src="{{ '/assets/js/search.js' | url }}"></script>
     <link rel="stylesheet" type="text/css" href="{{ '/assets/css/styles.css' | url }}">
     <link rel="stylesheet" href="{{ '/assets/css/print.css' | url }}" media="print">
+    <script src="https://unpkg.com/anchor-js@4.3.1/anchor.min.js"></script>
```
bmish, line 124: This was an actual dependency in package.json on the [old documentation site](https://github.dev/eslint/archive-website/blob/master/package.json#L33) so probably it should be here as well?

docs/src/_includes/components/docs-toc.html
```diff
@@ -1,6 +1,6 @@
 <nav class="docs-toc c-toc" aria-labelledby="js-toc-label">
     {%- if all_content | toc | safe | length -%}
-    <h2 class="c-toc__label" id="js-toc-label">Table of Contents</h2>
+    <p class="c-toc__label" id="js-toc-label">Table of Contents</p>
```
nzakas, line 3: This is actually a subheading so I don’t think we should change the html tag here. You should be able to filter this out of the anchorjs results using :not(c-toc__label)

docs/src/_includes/layouts/base.html
harish-sethuraman: Yea? We could do it either way? https://www.npmjs.com/package/anchor-js both are suggested as per their documentation?

bmish: Personally I prefer to take advantage of the package management provided by package.json and specify it there, especially given that we haven't begun using unpkg anywhere.

harish-sethuraman: we have to change the imports as `module` if we want to use imports or add passthrough through eleventy (I haven't tried these yet) anyway the passthrough is just going to import add the files into the assets and we will be importing it from there. I don't think there is going to be much difference using the dep from npm too instead of this approach which seems very straight forward.

nzakas: I think this is fine for now. If we end up needing to upgrade this dependency at any point, then I'd say let's move it to an npm dependency.

nzakas (approved):

PR #16159, opened 2022-07-25
feat: func-style: optionally allow arrow functions with let

jacekkopecky:
Add a new option for func-style when used with the "declaration" style.

The rationale is that in code styles that prefer function declarations and generally don't allow code like `const foo = () => {};`, there needs to be an exception when using `let` - variables that will be reassigned. Currently, the `func-style` rule either disallows both or allows both.

#### Prerequisites checklist

[ ] Documentation update
[ ] Add autofix to a rule
[ ] Add a CLI option
[ ] Add something to the core
[ ] Other, please explain:

`func-style`

fewer

a new option called `allowLetArrowFunctions`

```js
// foo is intended as a variable (it can be reassigned later)
// and we do not want to use a function declaration and reassign that.
let foo = () => {};
```

With the `"declaration"` option, this code will be considered incorrect.

With the `"declaration", { allowLetArrowFunctions: true }` options, the code will be allowed.

I added the new option, tests for incorrect and correct code, and updated the documentation.

If this is forbidden by the freeze of stylistic rules, can you please add guidance on best practices for these types of changes?

### DIFF
--- a/docs/src/rules/func-style.md
+++ b/docs/src/rules/func-style.md
@@ -61,9 +61,10 @@ This rule has a string option:
 * `"expression"` (default) requires the use of function expressions instead of function declarations
 * `"declaration"` requires the use of function declarations instead of function expressions
 
-This rule has an object option for an exception:
+This rule has two object options for exceptions:
 
 * `"allowArrowFunctions"`: `true` (default `false`) allows the use of arrow functions. This option applies only when the string option is set to `"declaration"` (arrow functions are always allowed when the string option is set to `"expression"`, regardless of this option)
+* `"allowLetArrowFunctions"`: `true` (default `false`) allows the use of arrow functions in `let` expressions. This option applies only when the string option is set to `"declaration"` (arrow functions are always allowed when the string option is set to `"expression"`, regardless of this option)
 
 ### expression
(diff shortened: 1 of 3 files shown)


nzakas:
Thanks for contributing, unfortunately we are no longer updating this rule. As [stated in the README](https://github.com/eslint/eslint#stylistic-rule-updates), we actually aren't making any further updates to rules that enforce stylistic preferences. Because you’ve already created a version of the rule that works the way you want, you can use it as a custom rule.

jacekkopecky:
Yes, understood @nzakas 

I'd seen both the readme or the policy before working on my pull request but it wasn't clear to me which rules are stylistic (yes, func-_style_ is a hint but it would be useful to be clearer about the distinction – I was looking for some indication of stylistic/non-stylistic in the rules list).

It would be useful if the policy linked to documentation on how to incorporate custom rules. I can't find anything less arduous than making a whole plugin – is that the way to go? Could the policy say that, with a link to a howto?

Since the eslint project has added this policy (for clear and understandable reasons, too), do you know if anybody has started a project to gather stylistic and other custom rules, where I could contribute this change instead?

PR #16190, opened 2022-08-10
feat: Implement caching for FlatESLint

nzakas:
Refs #13481

#### Prerequisites checklist

[ ] Documentation update
[ ] Add autofix to a rule
[ ] Add a CLI option
[x] Add something to the core
[ ] Other, please explain:

This PR implements caching for `FlatESLint` by making the following changes:

1. Enabling `cache: true` as an option for `FlatESLint`
2. Adding back all of the caching functionality that was removed for the first developer preview
3. Creating a `toJSON()` method on each config object that will normalize it into a form that is more appropriate for storing in the cache

For now, I throw an error when trying to serialize a config object where either the `parser` or the `processor` are objects instead of strings because it's not clear how best to serialize these. I see the following possible paths forward from here:

1. We ask that parsers and processors publish a `name` property that we can use for caching purposes and otherwise throw an error. This seems like the simplest approach.
2. We disallow objects for `processor` (same as eslintrc) and `parser` (different from eslintrc -- problematic for back compat reasons)
3. We figure out some way to hash a value for objects. Maybe for `parser` we could hash to `toString()` of the `parse()` method, and maybe for processor we could hash the `toString()` of both `preprocess()` and `postprocess()`?

### DIFF
--- a/lib/config/flat-config-array.js
+++ b/lib/config/flat-config-array.js
@@ -139,30 +139,71 @@ class FlatConfigArray extends ConfigArray {
     [ConfigArraySymbol.finalizeConfig](config) {
 
         const { plugins, languageOptions, processor } = config;
+        let parserName, processorName;
+        let invalidParser = false,
+            invalidProcessor = false;
 
         // Check parser value
-        if (languageOptions && languageOptions.parser && typeof languageOptions.parser === "string") {
-            const { pluginName, objectName: parserName } = splitPluginIdentifier(languageOptions.parser);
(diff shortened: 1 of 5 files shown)


btmills (approved):
LGTM!

For future serialization, I'm not a big fan of 2, and it sounds like you aren't either. As a variant of option 1, we could call it `cacheKey` and encourage packages to include version as `package-name@1.2.3`, making it even more thorough. As a fallback, option 3 seems like a reasonable best effort.

PR #16377, opened 2022-10-02
chore: use "ci" for Dependabot commit message

nschonni:
#### Prerequisites checklist

[ ] Documentation update
[ ] Add autofix to a rule
[ ] Add a CLI option
[ ] Add something to the core
[x] Other, please explain:

Noticed in https://github.com/eslint/eslint/pull/16350 that the commit message formatting check wasn't happy with the default message that Dependabot uses, so changed it so that it should be compliant for those updates to the GitHub Actions dependencies

https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#commit-message

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -4,3 +4,5 @@ updates:
     directory: "/"
     schedule:
       interval: "weekly"
+    commit-message:
+      prefix: "ci"


snitin315 (approved):
This change makes sense to me. I Would like others to review it as well.

kecrily:
I think dependency changes should be `chore` instead of `ci`.

.github/dependabot.yml
```diff
@@ -4,3 +4,5 @@ updates:
     directory: "/"
     schedule:
       interval: "weekly"
+    commit-message:
+      prefix: "ci"
```
kecrily, line 8: I think dependency changes should be `chore` instead of `ci`.

kecrily, line 8: I think dependency changes should belong to `chore`, aka current value. `ci` should refer to changes to CI configuration (like this PR), not modifications made by CI.

snitin315: Currently I see depandabot PRs for github-actions only. I suppose this is because we don't have a lockfile.

kecrily: Maybe we can try [renovate]()https://docs.renovatebot.com/?

snitin315: We are exploring it - https://github.com/eslint/eslint/issues/15726

nzakas: I agree with @kecrily. “ci” doesn’t seem appropriate for this type of change.

nschonni: From https://eslint.org/docs/latest/developer-guide/contributing/pull-requests `ci - changes to our CI configuration files and scripts.`.
The dependabot updates to GitHub Actions, are only to CI scripts in the `.github/workflows` folder. If Dependabot is later used for other types of updates like NPM, the default of `chore` would still be used, even with this change

nzakas: Ah fair point! I didn't realize this was scoped to just GitHub Actions. In that case, I agree that `ci` makes sense. :+1:

nzakas (approved):

PR #16394, opened 2022-10-07
docs: Document extending plugin with new config

### LINKED ISSUE #16310: docs: Add configs from plugin usage for the new config system.

In the new config system, the specification of plugin is different.

For example, a property `configs` does not have any effect.

How plugin works in the new config system should be described in the docs.

A new page should be introduced, like [this page](https://eslint.org/docs/latest/user-guide/configuring/configuration-files-new) is introduced.

### Participation

### Additional comments

_No response_

bpmutter:
Document how to extend a plugin configuration using the new ESLint configuration file type system.

Note that I included a comment in the example coming how extending a plugin works in the new config files compared to the old ones. I did this b/c I think that there's a significant enough change in behavior here for a critical feature that it's worth an explicit call out.

Fixes #16310

#### Prerequisites checklist

[x] Documentation update
[ ] Add autofix to a rule
[ ] Add a CLI option
[ ] Add something to the core
[ ] Other, please explain:

Document extending plugin for new ESLint config file type (`eslint.config.js`)

### DIFF
--- a/docs/src/user-guide/configuring/configuration-files-new.md
+++ b/docs/src/user-guide/configuring/configuration-files-new.md
@@ -329,7 +329,36 @@ For historical reasons, the boolean value `false` and the string value `"readabl
 
 ### Using plugins in your configuration
 
-Plugins are used to share rules, processors, configurations, parsers, and more across ESLint projects. Plugins are specified in a configuration object using the `plugins` key, which is an object where the name of the plugin is the property name and the value is the plugin object itself. Here's an example:
+Plugins are used to share rules, processors, configurations, parsers, and more across ESLint projects.
+
+#### Using configurations included in plugins
+
+You can use a configuration included in a plugin by adding that configuration
+directly to the `eslint.config.js` configurations array.
+Often, you do this for a plugin's recommended configuration. Here's an example:
+
+```js
+import jsdoc from "eslint-plugin-jsdoc";
+
+export default [
+    jsdoc.configs.recommended,
+    // ... other configuration objects
+];
+/*
+The above code is equivalent to the following configuration using the legacy .eslintrc
+configuration system:
+
+{
+  "extends": ["plugin:jsdoc/recommended"]
+}
+*/
+```
+
+#### Using plugin rules
+
+You can use specific rules included in a plugin. To do this, specify the plugin
+in a configuration object using the `plugins` key. The value for the `plugin` key
+is an object where the name of the plugin is the property name and the value is the plugin object itself. Here's an example:
 
 ```js
 import jsdoc from "eslint-plugin-jsdoc";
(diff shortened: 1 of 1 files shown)


kecrily (approved):
LGTM, thank you

docs/src/user-guide/configuring/configuration-files-new.md
```diff
@@ -329,7 +329,36 @@ For historical reasons, the boolean value `false` and the string value `"readabl
...
+
+```js
+import jsdoc from "eslint-plugin-jsdoc";
+
+export default [
+    jsdoc.configs.recommended,
+    // ... other configuration objects
+];
+/*
+The above code is equivalent to the following configuration using the legacy .eslintrc
```
nzakas, line 348: I don’t think we should refer to the old way of doing this at all because they are not actually equivalent. They accomplish the same thing in a very narrow way but there are subtle differences that I think will confuse people if we say they’re equivalent.

docs/src/user-guide/configuring/configuration-files-new.md
```diff
@@ -329,7 +329,36 @@ For historical reasons, the boolean value `false` and the string value `"readabl
...
+You can use a configuration included in a plugin by adding that configuration
+directly to the `eslint.config.js` configurations array.
+Often, you do this for a plugin's recommended configuration. Here's an example:
+
+```js
+import jsdoc from "eslint-plugin-jsdoc";
+
+export default [
+    jsdoc.configs.recommended,
+    // ... other configuration objects
```
nzakas, line 345: I think it’s important to include an actual config object here as an example. It can be fairly simple, just setting a jsdoc rule to “warn” would suffice.

docs/src/user-guide/configuring/configuration-files-new.md
```diff
@@ -329,7 +329,36 @@ For historical reasons, the boolean value `false` and the string value `"readabl
...
+The above code is equivalent to the following configuration using the legacy .eslintrc
+configuration system:
+
+{
+  "extends": ["plugin:jsdoc/recommended"]
+}
+*/
+```
+
+#### Using plugin rules
```
nzakas, line 357: I’d suggest flipping the order of these two sections so using plugin rules comes first. That way, you can show how you can include a config from a plugin and also override the config’s default rule options with another config object in the array.

nzakas (changes requested):
Thanks for this, I just left a few comments on ways I think we can make things clearer.

bpmutter:
@nzakas incorporated your feedback. ready for re-review!

mdjermanovic (approved):
LGTM, thanks!

nzakas (approved):
LGTM. Thanks!

PR #16466, opened 2022-10-29
docs: enable stylelint declaration-block-no-duplicate-properties

nschonni:
#### Prerequisites checklist

[ ] Documentation update
[ ] Add autofix to a rule
[ ] Add a CLI option
[ ] Add something to the core
[x] Other, please explain:

Enable the stylelint rule to remove duplicate/shadowed property values and auto-fixeed occurances.

Mostly removed duplicate values, but one or 2 were different and may have been some sort of fallback

### DIFF
--- a/docs/src/assets/scss/carbon-ads.scss
+++ b/docs/src/assets/scss/carbon-ads.scss
@@ -15,9 +15,9 @@
     padding: .6em;
     font-size: 1rem;
     overflow: hidden;
-    border-radius: 4px;
     background-color: var(--body-background-color);
     border: 1px solid var(--border-color);
+    border-radius: 4px;
     border-radius: var(--border-radius);
     box-shadow: 0 1px 4px 1px hsla(0, 0%, 0%, 0.1);
 
--- a/docs/src/assets/scss/languages.scss
+++ b/docs/src/assets/scss/languages.scss
@@ -13,10 +13,10 @@
 
     a {
         color: inherit;
-        display: block;
         width: 100%;
         padding: .75rem .1rem;
         text-decoration: none;
+        display: block;
         display: flex;
         align-items: center;
         border-bottom: 1px solid var(--divider-color);
--- a/docs/src/assets/scss/syntax-highlighter.scss
+++ b/docs/src/assets/scss/syntax-highlighter.scss
@@ -36,7 +36,6 @@ pre[class*="language-"] {
     padding: 1.5rem;
     margin: 1.5rem 0;
     overflow: auto;
-    background-color: var(--color-neutral-50);
     border-radius: var(--border-radius);
     background-color: var(--lightest-background-color);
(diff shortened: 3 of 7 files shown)


docs/src/assets/scss/carbon-ads.scss
```diff
@@ -15,7 +15,6 @@
     padding: .6em;
     font-size: 1rem;
     overflow: hidden;
-    border-radius: 4px;
```
harish-sethuraman, line 18: I remember @nzakas  was telling these dups were in place knowingly. Some browsers did not have css variables support https://caniuse.com/css-variables. so it was intentionally added.

nschonni: OK, I allowed the consecutive duplicate properties, although there are variables used throughout, so I'm not sure if IE11 compat is still even possible. That's the only browser that doesn't support them these days

docs/src/assets/scss/languages.scss
```diff
@@ -13,10 +13,10 @@
 
     a {
         color: inherit;
-        display: block;
         width: 100%;
         padding: .75rem .1rem;
         text-decoration: none;
+        display: block;
         display: flex;
```
harish-sethuraman, line 20: interesting 😂  we have both block and flex

harish-sethuraman: Is it possible for us to check which value takes precedence and gets applied?

nschonni: I assumed it might have been another Old IE thing, so I left it on the second pass.

nschonni: I think as a separate discussion, something like PostCSS might be able to calculate those variables as a build step based on a supported browser range. It's been awhile since I looked at that type of tooling though

nzakas:
Yes, so a lot of the duplicate properties were intentional for fallback to older browsers. I’m not sure how safe it is to start removing them.

In CSS, the last valid property is used.

harish-sethuraman:
If we know the browser support list we are targetting we can add auto prefixer to support vendor specific css and also we can try adding support for older browsers?

docs/src/assets/scss/syntax-highlighter.scss
```diff
@@ -36,7 +36,6 @@ pre[class*="language-"] {
     padding: 1.5rem;
     margin: 1.5rem 0;
     overflow: auto;
-    background-color: var(--color-neutral-50);
```
nschonni, line 39: @nzakas this was the only property removed that had a different value, as they both used variables, so I didn't see it as a fallback. The others with unique/fallback values were moved so they're side-by-side and configured to remove that style

nzakas: Gotcha. Thanks for the details.

nzakas (approved):

Issue #16050, opened 2022-06-24
add ESLint init commands instead of installation commands


harish-sethuraman:
### ESLint version

v8.18.0

### What problem do you want to solve?

```console
npm install eslint --save-dev

# or

yarn add eslint --dev
```

we right now have npm and yarn commands to install and get started with ESLint. How about adding pnpm to the docs as well. We do support `pnpm` via `@eslint/config` so it might be nice to provide that as well

### What do you think is the correct solution?

Adding `pnpm` commands equivalents to the getting started page.

### Participation

- [X] I am willing to submit a pull request for this change.

nzakas:
I'd prefer not to start adding every package manager into the docs. I wonder if we can just now say `npm init @eslint/config` rather than instructing people to manually install ESLint?

harish-sethuraman:
That would be better as well. Should we then remove the npm installation commands and replace them with `npm init @eslint/config`?

nzakas:
Seems fine to me. Let’s see what the rest of the team thinks.

nzakas:
@eslint/eslint-tsc this is a quick one, shall we change our docs to recommend `npm init @eslint/config` as the way to install ESLint?

aladdin-add:
sounds good! as the `eslint --init` has been moved to `@eslint/create-config`, it's no longer needed to manually install.

nzakas:
Marking as accepted.

PR #16483, opened 2022-11-01
docs: init config about specifying shared configs

kecrily:
#### Prerequisites checklist

[x] Documentation update
[ ] Add autofix to a rule
[ ] Add a CLI option
[ ] Add something to the core
[ ] Other, please explain:

init config about specifying shared configs

ref https://github.com/eslint/create-config/pull/38

### DIFF
--- a/docs/src/user-guide/getting-started.md
+++ b/docs/src/user-guide/getting-started.md
@@ -47,7 +47,7 @@ npm init @eslint/config -- --config semistandard,standard
 npm init @eslint/config -- --config semistandard --config standard
 ```
 
-**Note:** `npm init @eslint/config` assumes you have a `package.json` file already. If you don't, make sure to run `npm init` beforehand.
+**Note:** `npm init @eslint/config` assumes you have a `package.json` file already. If you don't, make sure to run `npm init` or `yarn init` beforehand.
 
 After that, you can run ESLint on any file or directory like this:
 
--- a/docs/src/user-guide/getting-started.md
+++ b/docs/src/user-guide/getting-started.md
@@ -23,9 +23,40 @@ You can install and configure ESLint using this command:
 
 ```shell
 npm init @eslint/config
+
+
+# or
+
+yarn init @eslint/config
+
+# or
+
+pnpm create @eslint/config
 ```
 
-**Note:** `npm init @eslint/config` assumes you have a `package.json` file already. If you don't, make sure to run `npm init` or `yarn init` beforehand.
+If you want to use the specified shared configs:
+
+```shell
+# use `eslint-config-semistandard` shared config
+# npm 6.x, yarn or pnpm
+npm init @eslint/config --config semistandard
+
+# ⚠️ npm 7+, extra double-dash is needed:
+npm init @eslint/config -- --config semistandard
+
+# or (`eslint-config` prefix is optional)
+npm init @eslint/config -- --config eslint-config-semistandard
+```
+
+The `--config` flag also supports passing in arrays:
+
+```shell
+npm init @eslint/config -- --config semistandard,standard
+# or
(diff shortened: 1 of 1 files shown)


docs/src/user-guide/getting-started.md
```diff
@@ -23,9 +23,40 @@ You can install and configure ESLint using this command:
...
 npm init @eslint/config
+
+
+# or
+
+yarn init @eslint/config
+
+# or
+
+pnpm create @eslint/config
```
harish-sethuraman, line 34: https://github.com/eslint/eslint/issues/16050#issuecomment-1165831664

We talked about it earlier and decided to not add all package managers

> I'd prefer not to start adding every package manager into the docs. I wonder if we can just now say npm init @eslint/config rather than instructing people to manually install ESLint?

nzakas: That's correct. We aren't including every option. When they run `npm init @eslint/config`, it will ask if they want to use yarn or pnpm. There's no need to show multiple options here.

docs/src/user-guide/getting-started.md
```diff
@@ -35,6 +66,10 @@ npx eslint yourfile.js
 # or
 
 yarn run eslint yourfile.js
+
+# or
+
+pnpm eslint yourfile.js
```
nzakas, line 72:
```suggestion
```

nzakas (changes requested):

docs/src/user-guide/getting-started.md
kecrily: Do we need to remove the yarn command here as well?

docs/src/user-guide/getting-started.md
```diff
@@ -25,7 +25,29 @@ You can install and configure ESLint using this command:
 npm init @eslint/config
 ```
 
-**Note:** `npm init @eslint/config` assumes you have a `package.json` file already. If you don't, make sure to run `npm init` or `yarn init` beforehand.
+If you want to use a specific shareable config that is hosted on npm, you can use the `--config` option and specify the package name:
+
+```shell
+# use `eslint-config-semistandard` shared config
+# npm 6.x, yarn or pnpm
```
snitin315, line 32:
```suggestion
# npm 6.x
```

snitin315 (approved):
LGTM, thanks!

docs/src/user-guide/getting-started.md
```diff
@@ -25,7 +25,29 @@ You can install and configure ESLint using this command:
...
+
+The `--config` flag also supports passing in arrays:
+
+```shell
+npm init @eslint/config -- --config semistandard,standard
+# or
+npm init @eslint/config -- --config semistandard --config standard
+```
+
+**Note:** `npm init @eslint/config` assumes you have a `package.json` file already. If you don't, make sure to run `npm init` beforehand.
```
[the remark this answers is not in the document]
kecrily: I went with @nzakas idea of not adding all package managers, either yarn or pnpm to the documentation, so only the npm command was kept.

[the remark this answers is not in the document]
mdjermanovic: This is different from `npm init @eslint/config` which executes `@eslint/create-config` which then asks the user to choose a package manager to install dependencies so it doesn't matter whether it was run through `npm` or another package manager. I wouldn't add more package managers here, but I'd keep yarn init to show that it isn't necessary to run _npm_ init before running `npm init @eslint/config`, and yarn users should indeed run `yarn init` instead of `npm init`.

mdjermanovic (approved):
LGTM, thanks!

nzakas (approved):

nzakas:
LGTM. Let’s just wait until after we verify there’s not a patch release.

PR #16687, opened 2022-12-20
docs: switch language based on current path

kecrily:
#### Prerequisites checklist

[x] Documentation update
[ ] Add autofix to a rule
[ ] Add a CLI option
[ ] Add something to the core
[ ] Other, please explain:

Instead of directing users to homepages in other languages for no reason, they are redirected to corresponding pages in other languages based on the path of the current page

It is an addition to https://github.com/eslint/eslint.org/pull/380


docs/src/_includes/components/language-switcher.html
```diff
@@ -14,7 +14,7 @@
         <select name="language selector" id="language-select" aria-describedby="language-infobox"
             class="c-custom-select switcher__select">
             {% for key, other_site in sites %}
-            <option value="{{ other_site.language.code }}" data-url="https://{{ other_site.hostname }}" {% if
+            <option value="{{ other_site.language.code }}" data-url="https://{{ other_site.hostname }}/docs{{ page.url }}" {% if
```
mdjermanovic, line 17:
```suggestion
            <option value="{{ other_site.language.code }}" data-url="https://{{ other_site.hostname }}/docs{{ page.url | prettyURL }}" {% if
```

I _think_ we should use the [`prettyURL`](https://github.com/eslint/eslint/blob/8bda20e8276c6ba17d31842fcdd63ba65476fbbd/docs/.eleventy.js#L134) filter here, otherwise URLs would end with `.html`. It might be best to rebase after https://github.com/eslint/eslint/pull/16686 is merged so we could test how this works.

Also, when the version is `head`, maybe we should hide languages that don't have a `head` version rather than directing to their `/docs` which will subsequently redirect to the `latest` version?

kecrily: I don't think it should be hidden, because then they would need to switch back to `latest` and do the language switch again.

But we can avoid duplicate redirects and do it in one step.

docs/src/_includes/components/language-switcher.html
```diff
@@ -14,8 +14,11 @@
         <select name="language selector" id="language-select" aria-describedby="language-infobox"
             class="c-custom-select switcher__select">
             {% for key, other_site in sites %}
-            <option value="{{ other_site.language.code }}" data-url="https://{{ other_site.hostname }}" {% if
-                site.language.code==other_site.language.code %} selected {% endif %}>
+            <option
+                value="{{ other_site.language.code }}"
+                data-url="https://{{ other_site.hostname }}/docs{% if process.env.BRANCH !== '/latest' %}/latest{% endif %}{{ page.url | prettyURL }}"
```
mdjermanovic, line 19:
```suggestion
                data-url="https://{{ other_site.hostname }}/docs/latest{{ page.url | prettyURL }}"
```

This conditional seems unnecessary. If we always want a direct link to the latest version in another language, then it should be the same when the currently browsed version is the latest version?

docs/src/_includes/components/language-switcher.html
mdjermanovic: > I don't think it should be hidden, because then they would need to switch back to `latest` and do the language switch again.

My concern is - if someone is browsing `https://eslint.org/docs/head/*` or another non-latest version (for example, a previous version or a prerelease version), and then switches to another language, they will get the `/latest` version in that language but they might think they're still reading the docs for the same version.

amareshsm (approved):
LGTM . I would like someone else to double-check.

docs/src/_includes/components/language-switcher.html
nzakas: I agree with @mdjermanovic -- we should only redirect to the equivalent page. If it doesn't exist, we shouldn't show it as an option.

kecrily: I got the key point. Now there is a question, how do we get the latest version number of the docs of each language? Maintaining the data manually is a dumb solution.

---

Do we have plans to release older versions of the docs? Otherwise, it seems that we will never be able to keep the version numbers based on the docs in each language consistent, meaning that language switching will basically useless

nzakas: We aren't going to re-release old versions of docs, but going forward we will keep the last minor/patch version of each major version.

For example, if we were to release ESLint v9.0.0 next week, we will still keep the docs for v8.31.0 and they would be selectable in the dropdown on the website.

nzakas:
@mdjermanovic can you review this to make sure your concerns were addressed?

mdjermanovic:
I think https://github.com/eslint/eslint/pull/16687#discussion_r1056884797 hasn't been addressed yet.

docs/src/_includes/components/language-switcher.html
```diff
@@ -14,9 +14,12 @@
...
             {% for key, other_site in sites %}
-            <option value="{{ other_site.language.code }}" data-url="https://{{ other_site.hostname }}" {% if
-                site.language.code==other_site.language.code %} selected {% endif %}>
-                {{ other_site.language.flag }} {{ other_site.language.name }}
+            <option
+                value="{{ other_site.language.code }}"
+                data-url="https://{{ other_site.hostname }}/docs/latest{{ page.url | prettyURL }}"
+                {% if site.language.code==other_site.language.code %}selected{% endif %}
+            >
+                {{ other_site.language.flag }} {{ other_site.language.name }} ({{ other_site.footer.language_switcher.latest }})
```
mdjermanovic, line 22: So, the idea is to always show the languages and always switch to the `latest` version in the selected language, but to add `(Latest)` to the text so that users are aware that they'll go to the `latest` version even if they're currently browsing e.g. the `HEAD` version?

docs/src/_includes/components/language-switcher.html
```diff
@@ -14,9 +14,12 @@
         <select name="language selector" id="language-select" aria-describedby="language-infobox"
             class="c-custom-select switcher__select">
             {% for key, other_site in sites %}
-            <option value="{{ other_site.language.code }}" data-url="https://{{ other_site.hostname }}" {% if
-                site.language.code==other_site.language.code %} selected {% endif %}>
-                {{ other_site.language.flag }} {{ other_site.language.name }}
+            <option
+                value="{{ other_site.language.code }}"
```
harish-sethuraman, line 18: This should ideally not work in localhost? Is this expected?

amareshsm:
Bump

docs/src/_includes/components/language-switcher.html
kecrily: Do you mean to hide it locally?

docs/src/_includes/components/language-switcher.html
kecrily: Yes, because currently HEAD always points to the English version.

docs/src/_includes/components/language-switcher.html
harish-sethuraman: Just asking if this would work in local as well?

nzakas:
@amareshsm please don't leave "bump" comments, as they just create noise and don't add anything to the discussion.

If you're waiting for a particular response, please mention that. 

@kecrily @mdjermanovic it looks like we are looking for feedback from you on this?

docs/src/_includes/components/language-switcher.html
mdjermanovic: Okay, that makes sense to me.

mdjermanovic (changes requested):
When browsing the HEAD version, the language dropdown shows "English (US) (Latest)" as the selected item, which would be confusing since the version being browsed is not the latest version:

image

docs/src/_includes/components/language-switcher.html
```diff
@@ -14,9 +14,12 @@
...
+                data-url="https://{{ other_site.hostname }}/docs/latest{{ page.url | prettyURL }}"
+                {% if site.language.code==other_site.language.code %}selected{% endif %}
+            >
+                {{ other_site.language.flag }} {{ other_site.language.name }}{% if site.language.code!=other_site.language.code and not isLatest %} ({{ other_site.footer.language_switcher.latest }}){% endif %}
```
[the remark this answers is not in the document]
mdjermanovic: > `(最新)` still appear is the expected behavior, it means that if you select it you will choose to jump to the latest Chinese site.

Isn't the expected behavior that `(最新)` doesn't appear on `eslint.org/docs/latest`? I thought that was the goal of the `not isLatest` condition.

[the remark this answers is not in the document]
kecrily: > Isn't the expected behavior that `(最新)` doesn't appear on `eslint.org/docs/latest`? I thought that was the goal of the `not isLatest` condition.

@mdjermanovic If the language of the current site is different from the language of the target site, `latest` is always displayed. If the current language and the target language are the same, latest is displayed when the current site is latest.

nzakas:
@mdjermanovic can you re-review this please?

docs/src/_includes/components/language-switcher.html
```diff
@@ -14,9 +14,14 @@
...
-            <option value="{{ other_site.language.code }}" data-url="https://{{ other_site.hostname }}" {% if
-                site.language.code==other_site.language.code %} selected {% endif %}>
+            <option
+                value="{{ other_site.language.code }}"
+                data-url="https://{{ other_site.hostname }}/docs/latest{{ page.url | prettyURL }}"
+                {% if site.language.code==other_site.language.code %}selected{% endif %}
+            >
                 {{ other_site.language.flag }} {{ other_site.language.name }}
+                {% if site.language.code!=other_site.language.code %}({{ other_site.footer.language_switcher.latest }})
+                {% elif GIT_BRANCH=="latest" %}({{ other_site.footer.language_switcher.latest }}){% endif %}
```
mdjermanovic, line 24: Can this be simplified with `or`?

```suggestion
                {% if site.language.code!=other_site.language.code or GIT_BRANCH=="latest" %}({{ other_site.footer.language_switcher.latest }}){% endif %}
```

kecrily: No, because `elif` is executed only if the first condition is `false`. If `or` is used, under `BRANCH=latest` all languages will be `latest`

mdjermanovic (approved):
LGTM, thanks!

PR #16712, opened 2022-12-26
docs: Add function call example for no-undefined

elliot-huffman:
Add a function call example on how to skip parameters the safe way

#### Prerequisites checklist

[X] Documentation update
[ ] Add autofix to a rule
[ ] Add a CLI option
[ ] Add something to the core
[ ] Other, please explain:

Add Examples for dunction parameter skipping

### DIFF
--- a/docs/src/rules/no-undefined.md
+++ b/docs/src/rules/no-undefined.md
@@ -63,7 +63,7 @@ function bar(baz, qux) {
     // ...
 }
 
-// Using function with improper refrence to the global undefined value to attempt to skip the first parameter
+// Using a function with improper reference to the undefined global value to attempt to skip the first parameter
 bar(undefined, "lorem");
 ```
 
--- a/docs/src/rules/no-undefined.md
+++ b/docs/src/rules/no-undefined.md
@@ -58,11 +58,6 @@ function foo(undefined) {
     // ...
 }
 
-// Function creation with parameter
-function bar(baz, qux) {
-    // ...
-}
-
 // Using a function with improper reference to the undefined global value to attempt to skip the first parameter
 bar(undefined, "lorem");
 ```
@@ -86,12 +81,7 @@ if (typeof foo === "undefined") {
 
 global.undefined = "foo";
 
-// Function creation with parameter
-function bar(baz, qux) {
-    // ...
-}
-
-// Using function without referencing the global undefined
+// Using function without referencing the global undefined value
 bar(void 0, "lorem");
 ```
 
--- a/docs/src/rules/no-undefined.md
+++ b/docs/src/rules/no-undefined.md
@@ -57,6 +57,14 @@ if (foo === undefined) {
 function foo(undefined) {
     // ...
 }
+
+// Function creation with parameter
+function bar(baz, qux) {
+    // ...
+}
+
+// Using function with improper refrence to the global undefined value to attempt to skip the first parameter
+bar(undefined, "lorem");
 ```
 
 :::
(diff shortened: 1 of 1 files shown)


docs/src/rules/no-undefined.md
```diff
@@ -57,6 +57,14 @@ if (foo === undefined) {
 function foo(undefined) {
     // ...
 }
+
+// Function creation with parameter
+function bar(baz, qux) {
+    // ...
+}
+
+// Using function with improper refrence to the global undefined value to attempt to skip the first parameter
```
amareshsm, line 66:
```suggestion
// Using a function with improper reference to the undefined global value to attempt to skip the first parameter
```

docs/src/rules/no-undefined.md
```diff
@@ -77,6 +85,14 @@ if (typeof foo === "undefined") {
 }
 
 global.undefined = "foo";
+
+// Function creation with parameter
+function bar(baz, qux) {
+    // ...
+}
+
+// Using function wihtout refrencing the global undefined
```
amareshsm, line 94:
```suggestion
// Using function without referencing the global undefined
```

mdjermanovic, line 94:
```suggestion
// Skips the first parameter
```

Same here.

elliot-huffman: I agree

docs/src/rules/no-undefined.md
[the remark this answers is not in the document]
elliot-huffman: I agree

elliot-huffman:
Fixed the above :)

elliot-huffman:
Sorry about the double requests, this is my first time pull requesting this way.

amareshsm:
@mdjermanovic suggested removing the definition comment.  just mention `Skips the first parameter`
image

like this
```
// Skips the first parameter
bar(undefined, "lorem");

// Skips the first parameter
bar(void 0, "lorem");

```

elliot-huffman:
Fixed.

amareshsm (approved):
LGTM. Thanks 🚀

docs/src/rules/no-undefined.md
```diff
@@ -57,6 +57,9 @@ if (foo === undefined) {
 function foo(undefined) {
     // ...
 }
+
+// Skips the first parameter
```
nzakas, line 61: This is a bit confusing, what do you mean by "skips"? I think this comment is making it more confusing and we can just remove it.

This is in the "incorrect" section, so this would be flagged as incorrect by `no-undefined`

docs/src/rules/no-undefined.md
```diff
@@ -77,6 +80,9 @@ if (typeof foo === "undefined") {
 }
 
 global.undefined = "foo";
+
+// Skips the first parameter
```
nzakas, line 84: Same comment here. I'm not sure what this is supposed to mean.

nzakas (commented):
Thanks for the submission. I'm not sure what the comments in the code examples are supposed to describe, so please take a look.

docs/src/rules/no-undefined.md
elliot-huffman: Would the original comments (plus spelling fixes) be better?

docs/src/rules/no-undefined.md
elliot-huffman: Would the original comments (plus spelling fixes) be better?

mdjermanovic: I think it would be best to just remove the comments as the intent seems clear from the code.

mdjermanovic (approved):
LGTM, thanks!

PR #16752, opened 2023-01-06
docs: fix theme switcher button

chenxsan:
#### Prerequisites checklist

[ ] Documentation update

When I was browsing the Chinese version https://zh-hans.eslint.org/docs/latest/, I noticed the buttons didn't seem right when there're enough spaces (reproduced under firefox/chrome/safari/edge on macOS):

[ ] Add autofix to a rule
[ ] Add a CLI option
[ ] Add something to the core
[ ] Other, please explain:

Update the `flex` rule with explicit `flex-grow` and `flex-shrink` so the buttons will accommodate respectively.

### DIFF
--- a/docs/src/assets/scss/components/theme-switcher.scss
+++ b/docs/src/assets/scss/components/theme-switcher.scss
@@ -20,8 +20,7 @@
 }
 
 .theme-switcher__button {
-    flex-shrink: 1;
-    flex-grow: 1;
+    flex-wrap: wrap;
     box-shadow: var(--shadow-xs);
     padding: .625rem .875rem;
     display: inline-flex;
--- a/docs/src/assets/scss/components/theme-switcher.scss
+++ b/docs/src/assets/scss/components/theme-switcher.scss
@@ -20,7 +20,8 @@
 }
 
 .theme-switcher__button {
-    flex: 0;
+    flex-shrink: 1;
+    flex-grow: 1;
     box-shadow: var(--shadow-xs);
     padding: .625rem .875rem;
     display: inline-flex;


docs/src/assets/scss/components/theme-switcher.scss
```diff
@@ -20,7 +20,8 @@
 }
 
 .theme-switcher__button {
-    flex: 0;
+    flex-shrink: 1;
+    flex-grow: 1;
```
harish-sethuraman, line 24: Did you try in all device width? I see that this fails for an intermediate device width?

harish-sethuraman: Also I'd rather suggest so that the button wraps the text to the next line instead of pushing the second character sometimes
```suggestion
    flex-wrap: wrap;
```

chenxsan: > Did you try in all device width? I see that this fails for an intermediate device width?

The main point of this pull request is to fix the issue shown in the screenshot I had provided, i.e., when there're enough spaces buttons should take up them equally.

Regarding how to deal with it when there're not enough space, `flex-shrink` is the original implement, `flex-wrap` is another good option - if we all agree on this one, I can update it.

harish-sethuraman: > The main point of this pull request is to fix the issue shown in the screenshot I had provided, i.e., when there're enough spaces buttons should take up them equally.

Yea, But still we aren't supposed to wrap the text? Im not how it works in Chinese (may be if its two word then wrap is okayish else nope) but somewhere in device width around `814` I still see button being this way. Also seems like padding remains the same for large designs and smaller ones which forces the content to either wrap when its long. 
Screenshot 2023-01-08 at 11 07 08 AM

chenxsan: Not sure we're on the same page.

There're two cases here:

1. when there're enough spaces, flex items should grow equally - hopefully we all agree on this.
2. when there aren't enough spaces, to shrink, or to wrap. The original code chose to shrink, and I kept the behavior in this pull request, while you suggest applying `flex-wrap` to wrap, right?

But it seems you thought there were third case here?

kecrily: > Im not how it works in Chinese (may be if its two word then wrap is okayish else nope) but somewhere in device width around `814` I still see button being this way.

They are one word, I think it should wrap, using `flex-wrap: wrap` would be better.

harish-sethuraman (approved):
Thanks 😄

snitin315 (approved):

Issue #16650, opened 2022-12-12
Bug: no-fallthrough interaction with disable comments and "unused disable directive" warnings


bradzacher:
### Environment

ESLint version: 8.29.0

### What parser are you using?

Default (Espree)

### What did you do?

<details>
<summary>Configuration</summary>

```js
module.exports = {
  reportUnusedDisableDirectives: true,
  rules: {
    'no-fallthrough': 'error',
  },
};
```
</details>

```js
switch (test) {
  case a:
    hello();

  // eslint-disable-next-line no-fallthrough
  case b:
}
```

### What did you expect to happen?

The rule report should be hidden and no "unused disable directive" warnings should be reported.

### What actually happened?

An unused disable directive warning is reported
![image](https://user-images.githubusercontent.com/7462525/206941185-9187b66c-f44c-48e3-9df5-047f7fa2251a.png)

### Participation

- [ ] I am willing to submit a pull request for this issue.

### Additional comments

This default behaviour is problematic because if you run `eslint --fix`, then eslint will remove the unused disable comment, causing no-fallthrough error.

I think that this occurs because the default comment regex will match the `no-fallthrough` string
https://github.com/eslint/eslint/blob/ad44344ef6fdeac7217eb83bc54a230382c0da5e/lib/rules/no-fallthrough.js#L11

IMO either the rule should ignore eslint-disable directives, or the default regex should be updated to ignore them to prevent this behaviour.

bradzacher:
It seems like a weird inconsistency in a core rule that you shouldn't disable.

I wouldn't mind if it was something like `eslint-plugin-eslint-comments` reporting on the unused disable - but the fact that eslint itself reports on the unused disable and subsequently "fix deletes" the valid comment to then cause the lint error that the comment was suppressing seems like really broken behaviour.

Esp considering the default for `reportUnusedDisableDirectives` is currently false, but in future it might be true - meaning with the default config eslint would delete a valid disable comment!

-----

For context what I was doing was enabling the rule on a large codebase. I don't have enough context for each of the violations to be able to determine if it's intentional or not so the workflow I've been using is to fix the ones I can, and those I can't add a disable comment like `// eslint-disable-next-line no-fallthrough -- TODO FIXME please review this fallthrough and determine if it was intentional` to explicitly mark the problem area for clean up.

But ofc my comment includes `fallthrough` (twice) which by default stops the rule from reporting, which triggers the bug and causes the disable to be marked as unused and get deleted.

[...]

PR #16757, opened 2023-01-07
fix: ignore directives for no-fallthrough

gfyoung:
#### Prerequisites checklist

[ ] Documentation update
[ ] Add autofix to a rule
[ ] Add a CLI option
[ ] Add something to the core
[ ] Other, please explain:

Fixes https://github.com/eslint/eslint/issues/16650

Fallthrough comment pattern detection was incorrectly matching to `ESLint` directives.

Abstracts logic from main `linter.js` file to detect directives.  That detection now takes precedence over fallthrough comment pattern detection, regardless of whether the default pattern or user-provided pattern is used.

N/A

### DIFF
--- a/lib/rules/no-fallthrough.js
+++ b/lib/rules/no-fallthrough.js
@@ -5,10 +5,26 @@
 "use strict";
 
 //------------------------------------------------------------------------------
+// Requirements
+//------------------------------------------------------------------------------
+
+const { directivesPattern } = require("../shared/directives");
+
+//------------------------------------------------------------------------------
 // Helpers
 //------------------------------------------------------------------------------
 
 const DEFAULT_FALLTHROUGH_COMMENT = /falls?\s?through/iu;
+
+/**
+ * Checks whether or not a given comment string is really a fallthrough comment and not an ESLint directive.
+ * @param {string} comment The comment string to check.
+ * @param {RegExp} fallthroughCommentPattern The regular expression used for checking for fallthrough comments.
+ * @returns {boolean} `true` if the comment string is truly a fallthrough comment.
+ */
+function isFallThroughComment(comment, fallthroughCommentPattern) {
+    return fallthroughCommentPattern.test(comment) && !directivesPattern.test(comment.trim());
+}
 
 /**
  * Checks whether or not a given case has a fallthrough comment.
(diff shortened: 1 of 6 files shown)


lib/rules/no-fallthrough.js
```diff
@@ -4,12 +4,28 @@
...
 const DEFAULT_FALLTHROUGH_COMMENT = /falls?\s?through/iu;
 
+/**
+ * Checks whether or not a given comment string is really a fallthrough comment and not an ESLint directive.
+ * @param {string} comment The comment string to check.
+ * @param {RegExp} fallthroughCommentPattern The regular expression used for checking for fallthrough comments.
+ * @returns {boolean} `true` if the comment string is truly a fallthrough comment.
+ */
+function isFallThroughComment(comment, fallthroughCommentPattern) {
+    return fallthroughCommentPattern.test(comment) && !directivesPattern.test(comment.trim());
```
gfyoung, line 26: Had to use `.trim()` at the end because `directivesPattern` is whitespace-sensitive.

gfyoung:
https://github.com/eslint/eslint/actions/runs/3861076597/jobs/6581823868#step:5:50

This failure looks flaky for reasons unrelated to this PR.  Seeing failure in another PR too:

https://github.com/eslint/eslint/actions/runs/3843668330/jobs/6546117108#step:5:50

tests/lib/rules/no-fallthrough.js
```diff
@@ -297,6 +297,18 @@ ruleTester.run("no-fallthrough", rule, {
                     column: 34
                 }
             ]
+        },
+        {
+            code: "switch (foo) { case 0: a(); \n// eslint-enable no-fallthrough\n case 1: }",
```
snitin315, line 302: Can we add more test cases with other disable directives as well - `eslint-disable`, `eslint-disable-next-line`. Would be great to include the same case mentioned in the issue 

```js
switch (test) {
  case a:
    hello();

  // eslint-disable-next-line no-fallthrough
  case b:
}
```

gfyoung: If I had used a disable directive (instead of enable), the test would have passed without my changes.

I can add them, but I'm not sure I see as much because a disable directive and a comment pattern matching a disable directive both would cause no warning to be emitted.

gfyoung: @snitin315 is there a way to check for unused directives in these tests and report them as errors?

Otherwise, I don't really see how adding the OP case would be of too much help given my earlier comment.

snitin315: @gfyoung The above case will be a part of valid test cases (in this file) in case of no errors. Also, we should add a test case to assert that we don't get unused disable directive warning [here](https://github.com/eslint/eslint/blob/3e34418b31664decfb2337de798feafbf985b66c/tests/lib/linter/linter.js#L3994) like in OP, right?

gfyoung: > Also, we should add a test case to assert that we don't get unused disable directive warning [here](https://github.com/eslint/eslint/blob/3e34418b31664decfb2337de798feafbf985b66c/tests/lib/linter/linter.js#L3994) like in OP, right?

That was exactly my question, and your answer is what I was looking for.  Thanks @snitin315.  Added.

snitin315 (approved):
LGTM, I'll leave it open for others to review.

btmills (approved):
LGTM, thanks!

PR #16758, opened 2023-01-07
chore: sync templates/*.md files with issue templates

gfyoung:
#### Prerequisites checklist

[x] Documentation update
[ ] Add autofix to a rule
[ ] Add a CLI option
[ ] Add something to the core
[ ] Other, please explain:

The issue templates are essentially more advanced versions of the ones that were being (redundantly) maintained in the `templates` directory.  <strike>Used relative paths instead instead of hardcoding `eslint/eslint/HEAD` because that makes it easier to review a branch that makes any direct changes to the referenced documentation files within a file like this one.</strike>

N/A

### DIFF
--- a/templates/rule-change-proposal.md
+++ b/templates/rule-change-proposal.md
@@ -1,8 +1,17 @@
 **What rule do you want to change?**
 
-**Does this change cause the rule to produce more or fewer warnings?**
+**What change do you want to make? (place an "X" next to just one item)?**
 
-**How will the change be implemented? (New option, new default behavior, etc.)?**
+[ ] Generate more warnings
+[ ] Generate fewer warnings
+[ ] Implement autofix
+[ ] Implement suggestions
+
+**How will the change be implemented? (place an "X" next to just one item)?**
+
+[ ] A new option
+[ ] A new default behavior
+[ ] Other
 
 **Please provide some example code that this change will affect:**
 
--- a/templates/rule-change-proposal.md
+++ b/templates/rule-change-proposal.md
@@ -1,13 +1,13 @@
 **What rule do you want to change?**
 
-**What change do you want to make? (place an "X" next to just one item)?**
+**What change do you want to make (place an "X" next to just one item)?**
 
 [ ] Generate more warnings
 [ ] Generate fewer warnings
 [ ] Implement autofix
 [ ] Implement suggestions
 
-**How will the change be implemented? (place an "X" next to just one item)?**
+**How will the change be implemented (place an "X" next to just one item)?**
 
 [ ] A new option
 [ ] A new default behavior
--- a/templates/bug-report.md
+++ b/templates/bug-report.md
@@ -1,10 +1,19 @@
-**Tell us about your environment:**
+**Tell us about your environment (`npx eslint --env-info`):**
 
-* **ESLint Version:**
-* **Node Version:**
-* **npm Version:**
+* **Node version:**
+* **npm version:**
+* **Local ESLint version:**
+* **Global ESLint version:**
+* **Operating System:**
 
-**What parser (default, `@babel/eslint-parser`, `@typescript-eslint/parser`, etc.) are you using?**
+**What parser are you using (place an "X" next to just one item)?**
+
+[ ] `Default (Espree)`
+[ ] `@typescript-eslint/parser`
+[ ] `@babel/eslint-parser`
+[ ] `vue-eslint-parser`
+[ ] `@angular-eslint/template-parser`
+[ ] `Other`
 
 **Please show your full configuration:**
 
--- a/templates/rule-proposal.md
+++ b/templates/rule-proposal.md
@@ -1,13 +1,14 @@
-**Please describe what the rule should do:**
+**What should the new rule do?**
+
+**What new ECMAScript feature does this rule relate to? Note that we only accept new core rules related to new ECMAScript features.**
 
 **What category of rule is this (place an "X" next to just one item)?**
 
-[ ] Enforces code style
-[ ] Warns about a potential error
+[ ] Warns about a potential problem
 [ ] Suggests an alternate way of doing something
-[ ] Other (please specify:)
+[ ] Enforces a formatting/stylistic preference
 
-**Provide 2-3 code examples that this rule will warn about:**
+**Please provide some example JavaScript code that this rule will warn about:**
 
 ```js
 <!-- put your code examples here -->


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -6,7 +6,7 @@
 
 #### Prerequisites checklist
 
-- [ ] I have read the [contributing guidelines](https://github.com/eslint/eslint/blob/HEAD/CONTRIBUTING.md).
+- [ ] I have read the [contributing guidelines](../CONTRIBUTING.md).
```
kecrily, line 9: Using a relative link here will make the link in the last generated PR content inaccessible.

kecrily (changes requested):

.github/PULL_REQUEST_TEMPLATE.md
gfyoung: Good point.  Fixed.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -16,16 +16,16 @@
 -->
 
 [ ] Documentation update
-[ ] Bug fix ([template](https://raw.githubusercontent.com/eslint/eslint/HEAD/templates/bug-report.md))
-[ ] New rule ([template](https://raw.githubusercontent.com/eslint/eslint/HEAD/templates/rule-proposal.md))
-[ ] Changes an existing rule ([template](https://raw.githubusercontent.com/eslint/eslint/HEAD/templates/rule-change-proposal.md))
+[ ] Bug fix ([template](ISSUE_TEMPLATE/bug-report.yml))
+[ ] New rule ([template](ISSUE_TEMPLATE/new-rule.yml))
+[ ] Changes an existing rule ([template](ISSUE_TEMPLATE/rule-change.yml))
```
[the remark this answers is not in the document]
gfyoung: To be honest, I would prefer from a maintainer standpoint that contributors just open an issue corresponding to the template.  That way they can provide the information you guys need **without having to maintain the same templates twice on your end**.  As for copy/paste, contributors can still copy/paste from GitHub's rendered version of the YML.

gfyoung:
https://github.com/eslint/eslint/actions/runs/3863046536/jobs/6584991911#step:5:50

This failure looks flaky for reasons unrelated to this PR. Seeing same failure in another PR:

https://github.com/eslint/eslint/actions/runs/3843668330/jobs/6546117108#step:5:50

This happened in my other PR too, but I can't restart the build to make it ✅.

.github/PULL_REQUEST_TEMPLATE.md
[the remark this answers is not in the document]
harish-sethuraman: I feel like copy pasting the markdown is better than the rendered YML.

> without having to maintain the same templates twice (in YML and MD format) on your end

Also these templates are not changed regularly so it would be better if leave them as suc?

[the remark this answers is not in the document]
gfyoung: Perhaps not, but there is already divergence between these two template file types, which feels problematic already. Separately, I don't see why users shouldn't just file an issue in the first place.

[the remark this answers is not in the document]
snitin315: >Separately, I don't entirely see why users can't/shouldn't just file an issue in the first place.

Sometimes the fix is a one-liner or very simple to implement that contributors directly submit a PR, in such cases this is helpful.

snitin315:
Removed `accepted` and added `evaluating` since we don't have a consensus on this.

.github/PULL_REQUEST_TEMPLATE.md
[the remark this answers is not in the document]
gfyoung: A simple fix does not necessarily mean the problem itself is easy to communicate to maintainers.

IMO in most cases I as a maintainer would prefer to see triage separate from a PR.  Consider the templates that there: should contributors really be opening PRs for new rules or rule changes without opening an issue first?  More often than not, I think there turns out to be more to discuss that would be better suited in an issue rather than a PR.

I'm still not seeing the benefit of maintaining these templates in two different forms when the YML versions appear to be more developed content-wise than the Markdown ones.  If it is truly the case that an issue is trivial to communicate and fix, I don't think you really need a template, let alone this many.

mdjermanovic:
I think the best path forward is to update the markdown templates if there is a difference in content. We do allow PRs without issues, and copy-pasting plain text from rendered GitHub issue templates doesn't seem like a good alternative to markdown.

gfyoung:
> We do allow PRs without issues

@mdjermanovic: My maintainer preference is to discuss issues separately from a PR where possible.  This would especially be applicable in cases of new rules and rule changes.  That said, if that is a not concern for you guys, I can modify this PR to just sync the Markdown content to the YML one.

mdjermanovic:
@gfyoung those are good points, and we may or may not revisit our PR policies in the future. However, the discussion about the policies doesn't seem relevant to this change, because this change does not enforce different policies. If and when we decide to require an issue, we'd probably need different changes to the PR template (for example, a checkbox to confirm that the PR addresses an issue, or a mandatory issue id field). This change as it is now, in my opinion, makes it more difficult for contributors to submit a PR under the current policies. Maintaining duplicate templates isn't ideal, but there doesn't seem to be a better alternative.

gfyoung:
> Maintaining duplicate templates isn't ideal, but there doesn't seem to be a better alternative.

Okay, fair.  I've updated the PR to just sync the content between the two then.

mdjermanovic (approved):
LGTM, thanks! Leaving open for others to review.

templates/rule-change-proposal.md
```diff
@@ -1,8 +1,17 @@
 **What rule do you want to change?**
 
-**Does this change cause the rule to produce more or fewer warnings?**
+**What change do you want to make? (place an "X" next to just one item)?**
```
harish-sethuraman, line 3:
```suggestion
**What change do you want to make? (place an "X" next to just one item)**
```

templates/rule-change-proposal.md
```diff
@@ -1,8 +1,17 @@
...
 
-**How will the change be implemented? (New option, new default behavior, etc.)?**
+[ ] Generate more warnings
+[ ] Generate fewer warnings
+[ ] Implement autofix
+[ ] Implement suggestions
+
+**How will the change be implemented? (place an "X" next to just one item)?**
```
harish-sethuraman, line 10:
```suggestion
**How will the change be implemented? (place an "X" next to just one item)**
```

templates/rule-change-proposal.md
harish-sethuraman: or  
```suggestion
**What change do you want to make (place an "X" next to just one item)?**
```

to maintain consistency at all places

templates/rule-change-proposal.md
harish-sethuraman: or 
```suggestion
**How will the change be implemented (place an "X" next to just one item)?**
```

gfyoung: Good catch!  Fixed.

templates/rule-change-proposal.md
gfyoung: Good catch!  Fixed.

snitin315 (approved):

harish-sethuraman (approved):
Thanks

kecrily (approved):
LGTM

PR #16785, opened 2023-01-14
feat: add `restrictDefaultExports` option to no-restricted-exports rule

snitin315:
#### Prerequisites checklist

[x] Documentation update
[ ] Add autofix to a rule
[ ] Add a CLI option
[ ] Add something to the core
[ ] Other, please explain:

Resolve https://github.com/eslint/eslint/issues/15617

* `"restrictDefaultExports"` is an object option with boolean properties to restrict certain default export declarations. The option works only if the `restrictedNamedExports` option does not contain the `"default"` value. The following properties are allowed:
  * `direct`: restricts `export default` declarations.
  * `named`: restricts `export { foo as default };` declarations.
  * `defaultFrom`: restricts `export { default } from 'foo';`, `export { default as default } from 'foo';` declarations.
  * `namedFrom`: restricts `export { foo as default } from 'foo';` declarations.
  * `namespaceFrom`: restricts `export * as default from 'foo';` declarations.

- #### [Docs preview link](https://deploy-preview-16785--docs-eslint.netlify.app/rules/no-restricted-exports#restrictdefaultexports)

Open to suggestions for the error message, if it should be different for each property.

### DIFF
--- a/docs/src/rules/no-restricted-exports.md
+++ b/docs/src/rules/no-restricted-exports.md
@@ -17,8 +17,16 @@ By default, this rule doesn't disallow any names. Only the names you specify in 
 This rule has an object option:
 
 * `"restrictedNamedExports"` is an array of strings, where each string is a name to be restricted.
+* `"restrictDefaultExports"` is an object option with boolean properties to restrict default exports in different formats. The option works only if the `restrictedNamedExports` option does not contain the `"default"` value. The following properties are allowed:
+  * `direct`: restricts `export default` declarations.
+  * `named`: restricts `export { foo as default };` declarations.
+  * `defaultFrom`: restricts `export { default } from 'foo';` declarations.
+  * `namedFrom`: restricts `export { foo as default } from 'foo';` declarations.
+  * `namespaceFrom`: restricts `export * as default from 'foo';` declarations.
 
-Examples of **incorrect** code for this rule:
+### restrictedNamedExports
+
+Examples of **incorrect** code for the `"restrictedNamedExports"` option:
 
 ::: incorrect
 
@@ -50,7 +58,7 @@ export { "👍" } from "some_module";
 
 :::
 
-Examples of **correct** code for this rule:
+Examples of **correct** code for the `"restrictedNamedExports"` option:
 
 ::: correct
 
@@ -82,11 +90,11 @@ export { "👍" as thumbsUp } from "some_module";
 
 :::
 
-### Default exports
+#### Default exports
 
-By design, this rule doesn't disallow `export default` declarations. If you configure `"default"` as a restricted name, that restriction will apply only to named export declarations.
+By design, the `"restrictedNamedExports"` option doesn't disallow `export default` declarations. If you configure `"default"` as a restricted name, that restriction will apply only to named export declarations.
 
-Examples of additional **incorrect** code for this rule:
+Examples of additional **incorrect** code for the `"restrictedNamedExports": ["default"]` option:
 
 ::: incorrect
 
@@ -110,7 +118,7 @@ export { default } from "some_module";
 
 :::
 
-Examples of additional **correct** code for this rule:
+Examples of additional **correct** code for the `"restrictedNamedExports": ["default"]` option:
 
 ::: correct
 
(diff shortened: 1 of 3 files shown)


lib/rules/no-restricted-exports.js
```diff
@@ -35,19 +35,51 @@ module.exports = {
                         type: "string"
                     },
                     uniqueItems: true
+                },
+                restrictDefaultExports: {
```
mdjermanovic, line 39: Can we update the schema so that a combination `{ restrictedNamedExports: [..., "default", ...], restrictDefaultExports: { ... } }` is an invalid config? I think it would be better than silently ignoring `restrictDefaultExports` when `restrictedNamedExports` contains `"default"`.

mdjermanovic: I think something like this would work, but please double-check:

```js
schema: [{
    type: "object",
    anyOf: [
        {
            properties: {
                restrictedNamedExports: {
                    type: "array",
                    items: {
                        type: "string"
                    },
                    uniqueItems: true
                }
            },
            additionalProperties: false
        },
        {
            properties: {
                restrictedNamedExports: {
                    type: "array",
                    items: {
                        type: "string",
                        pattern: "^(?!default$)"
                    },
                    uniqueItems: true
                },
                restrictDefaultExports: {
                    type: "object",
                    properties: {

                        // Allow/Disallow `export default foo; export default 42; export default function foo() {}` format
                        direct: {
                            type: "boolean"
                        },

                        // Allow/Disallow `export { foo as default };` declarations
                        named: {
                            type: "boolean"
                        },

                        //  Allow/Disallow `export { default } from "mod"; export { default as default } from "mod";` declarations
                        defaultFrom: {
                            type: "boolean"
                        },

                        //  Allow/Disallow `export { foo as default } from "mod";` declarations
                        namedFrom: {
                            type: "boolean"
                        },

                        //  Allow/Disallow `export * as default from "mod"`; declarations
                        namespaceFrom: {
                            type: "boolean"
                        }
                    },
                    additionalProperties: false
                }
            },
            additionalProperties: false
        }
    ]
}],
```

snitin315: This seems to work.

```js
/* eslint no-restricted-exports: ["error", { "restrictedNamedExports": ["default"], restrictDefaultExports: { direct: true, named: false, namespaceFrom: true }  }]  */
```

The above config throws the following error:
 
Screenshot 2023-01-17 at 6 50 08 PM

snitin315: [CI failed](https://github.com/eslint/eslint/actions/runs/3939718373/jobs/6739902001), push a minor refactor in https://github.com/eslint/eslint/pull/16785/commits/037b30c60503c870fd0cdda121a41e5d7e3fa61b

lib/rules/no-restricted-exports.js
```diff
@@ -63,6 +114,33 @@ module.exports = {
                     messageId: "restrictedNamed",
                     data: { name }
                 });
+                return;
+            }
+
+            if (name === "default") {
+                const isSourceSpecified = node.parent.source || node.parent.parent.source;
+                const specifierLocalName = node.parent.local && node.parent.local.name;
```
mdjermanovic, line 122: We should use `astUtils.getModuleExportName` here. For example:

```js
/* eslint no-restricted-exports: ["error", {
    restrictDefaultExports: {
        defaultFrom: false,
        namedFrom: true
    }
}]*/

export { "default" } from "foo"; // false positive; this is `defaultFrom` but was interpreted as `namedFrom`
```

mdjermanovic (changes requested):

lib/rules/no-restricted-exports.js
snitin315: Thanks for the heads up. Fixed in https://github.com/eslint/eslint/pull/16785/commits/4bf73804d94bf8443bcc868febd594f3c1253768

lib/rules/no-restricted-exports.js
```diff
@@ -27,27 +27,78 @@ module.exports = {
...
-            restrictedNamed: "'{{name}}' is restricted from being used as an exported name."
+            restrictedNamed: "'{{name}}' is restricted from being used as an exported name.",
+            restrictedDefault: "Exporting the 'default' value is restricted."
         }
     },
 
     create(context) {
 
         const restrictedNames = new Set(context.options[0] && context.options[0].restrictedNamedExports);
+        const restrictDefaultExports = context.options[0] && !restrictedNames.has("default") && context.options[0].restrictDefaultExports;
```
mdjermanovic, line 101:
```suggestion
        const restrictDefaultExports = context.options[0] && context.options[0].restrictDefaultExports;
```

Now it isn't necessary to check if `!restrictedNames.has("default")` since the schema wouldn't allow it?

lib/rules/no-restricted-exports.js
```diff
@@ -63,6 +114,44 @@ module.exports = {
...
                     data: { name }
                 });
+                return;
+            }
+
+            if (name === "default") {
+                if (node.parent.type === "ExportAllDeclaration") {
+                    const isSourceSpecified = node.parent.source;
+
+                    if (isSourceSpecified && restrictDefaultExports && restrictDefaultExports.namespaceFrom) {
```
mdjermanovic, line 124:
```suggestion

                    if (restrictDefaultExports && restrictDefaultExports.namespaceFrom) {
```

I believe checking if the source is specified is redundant because ExportAllDeclaration must have the source.

lib/rules/no-restricted-exports.js
```diff
@@ -63,6 +114,44 @@ module.exports = {
...
+                        context.report({
+                            node,
+                            messageId: "restrictedDefault"
+                        });
+                    }
+
+                } else { // ExportSpecifier
+                    const isSourceSpecified = node.parent.parent.source;
```
mdjermanovic, line 132:
```suggestion
                    const isSourceSpecified = !!node.parent.parent.source;
```

The variable name sounds like it should have a boolean value.

lib/rules/no-restricted-exports.js
snitin315: Yes, this is not required anymore.

docs/src/rules/no-restricted-exports.md
```diff
@@ -122,6 +130,85 @@ export default function foo() {}
...
+Examples of **incorrect** code for the `"restrictDefaultExports": { "direct": true }` option:
+
+::: incorrect
+
+```js
+/*eslint no-restricted-exports: ["error", { "restrictDefaultExports": { "direct": true } }]*/
+
+export default foo;
+export default 42;
+export default function foo() {};
```
mdjermanovic, line 148:
```suggestion
export default function foo() {}
```

Extra semi, this is a function declaration.

lib/rules/no-restricted-exports.js
```diff
@@ -27,27 +27,78 @@ module.exports = {
...
                 }
-            },
-            additionalProperties: false
+            ]
         }],
 
         messages: {
-            restrictedNamed: "'{{name}}' is restricted from being used as an exported name."
+            restrictedNamed: "'{{name}}' is restricted from being used as an exported name.",
+            restrictedDefault: "Exporting the 'default' value is restricted."
```
mdjermanovic, line 94:
```suggestion
            restrictedDefault: "Exporting 'default' is restricted."
```

`default` is binding, like other exports. I think we can shorten the message to make it more technically correct.

mdjermanovic (commented):
Just two small suggestions, then LGTM.

lib/rules/no-restricted-exports.js
snitin315: Updated 👍🏻

mdjermanovic (approved):
LGTM, thanks!

elyobo:
@snitin315 @mdjermanovic @btmills does the new [validation rules for `no-restricted-exports` options](https://github.com/eslint/eslint/blob/main/lib/rules/no-restricted-exports.js#L51) introduce a backwards compatible break here? Users with `default` in the `restrictedNamedExports` (which is the case in the popular airbnb / airbnb-base config) will be hit with an error in a minor release if I'm reading that correctly?

elyobo:
Thanks, makes sense.

PR #16838, opened 2023-01-30
docs: Custom Processors cleanup and expansion

### LINKED ISSUE #16814: Change Request: Custom Processors clean up and expansion
### ESLint version

latest

clean up and expand Custom Processors page

For Custom Processors page: 

add overview to page
- copy edit

### Participation

### Additional comments

_No response_

bpmutter:
#### Prerequisites checklist

[x] Documentation update
[ ] Add autofix to a rule
[ ] Add a CLI option
[ ] Add something to the core
[ ] Other, please explain:

Clean up and expansion of the Custom Processors page. 

Resolves #16814

### DIFF
--- a/docs/src/extend/custom-processors.md
+++ b/docs/src/extend/custom-processors.md
@@ -105,9 +105,6 @@ See [Specify a Processor](../use/configure/plugins#specify-a-processor) in the P
 
 ## File Extension-named Processor
 
-> TODO: i don't understand this line. is it saying that as long as the file extension named processor
-> is inclueded in the plugin, it runs?
-> Users don't need to specify the file extension-named processors in their config files.
 
 If a custom processor name starts with `.`, ESLint handles the processor as a **file extension-named processor**. ESLint applies the processor to files with that filename extension automatically. Users don't need to specify the file extension-named processors in their config files.
 
--- a/docs/src/extend/custom-processors.md
+++ b/docs/src/extend/custom-processors.md
@@ -124,3 +124,16 @@ module.exports = {
     }
 }
 ```
+
+You can also use the same custom processor with multiple filename extensions. The following example shows using same processor for both `.md` and `.mdx` files:
+
+```js
+module.exports = {
+    // The same custom processor is applied to both
+    // `.md` and `.mdx` files.
+    processors: {
+        ".md": myCustomProcessor,
+        ".mdx": myCustomProcessor
+    }
+}
+```
--- a/docs/src/extend/custom-processors.md
+++ b/docs/src/extend/custom-processors.md
@@ -105,7 +105,6 @@ See [Specify a Processor](../use/configure/plugins#specify-a-processor) in the P
 
 ## File Extension-named Processor
 
-
 If a custom processor name starts with `.`, ESLint handles the processor as a **file extension-named processor**. ESLint applies the processor to files with that filename extension automatically. Users don't need to specify the file extension-named processors in their config files.
 
 For example:
--- a/docs/src/extend/custom-processors.md
+++ b/docs/src/extend/custom-processors.md
@@ -128,6 +128,8 @@ module.exports = {
 You can also use the same custom processor with multiple filename extensions. The following example shows using same processor for both `.md` and `.mdx` files:
 
 ```js
+const myCustomProcessor = { /* processor methods */ };
+
 module.exports = {
     // The same custom processor is applied to both
     // `.md` and `.mdx` files.
--- a/docs/src/extend/custom-processors.md
+++ b/docs/src/extend/custom-processors.md
@@ -49,7 +49,7 @@ module.exports = {
 
 A code block has two properties `text` and `filename`. The `text` property is the content of the block and the `filename` property is the name of the block. The name of the block can be anything, but should include the file extension, which tells the linter how to process the current block. The linter checks the [`--ext` CLI option](../use/command-line-interface#--ext) to see if the current block should be linted and resolves `overrides` configs to check how to process the current block.
 
-It's up to the plugin to decide if it needs to return just one part of the non-JavaScript file or multiple pieces. For example in the case of processing `.html` files, you might want to return just one item in the array by combining all scripts. However, for `.md` files, because each JavaScript block might be independent, you can return multiple items.
+It's up to the plugin to decide if it needs to return just one part of the non-JavaScript file or multiple pieces. For example in the case of processing `.html` files, you might want to return just one item in the array by combining all scripts. However, for `.md` files, you can return multiple items because each JavaScript block might be independent.
 
 **The `postprocess` method** takes a two-dimensional array of arrays of lint messages and the filename. Each item in the input array corresponds to the part that was returned from the `preprocess` method. The `postprocess` method must adjust the locations of all errors to correspond to locations in the original, unprocessed code, and aggregate them into a single flat array and return it.
 
--- a/docs/src/extend/custom-processors.md
+++ b/docs/src/extend/custom-processors.md
@@ -7,11 +7,12 @@ eleventyNavigation:
     order: 5
 
 ---
-You can also create custom processors that tell ESLint how to process files other than JavaScript.
+
+You can also create custom processors that tell ESLint how to process files other than standard JavaScript. For example, you could write a custom processor to extract and process JavaScript from Markdown files ([eslint-plugin-markdown](https://www.npmjs.com/package/eslint-plugin-markdown) includes a custom processor for this).
 
 ## Custom Processor Specification
 
-In order to create a processor, the object that is exported from your module has to conform to the following interface:
+In order to create a custom processor, the object exported from your module has to conform to the following interface:
 
 ```js
 module.exports = {
@@ -46,27 +47,27 @@ module.exports = {
 
 **The `preprocess` method** takes the file contents and filename as arguments, and returns an array of code blocks to lint. The code blocks will be linted separately but still be registered to the filename.
 
-A code block has two properties `text` and `filename`; the `text` property is the content of the block and the `filename` property is the name of the block. Name of the block can be anything, but should include the file extension, that would tell the linter how to process the current block. The linter will check [`--ext` CLI option](../use/command-line-interface#--ext) to see if the current block should be linted, and resolve `overrides` configs to check how to process the current block.
+A code block has two properties `text` and `filename`. The `text` property is the content of the block and the `filename` property is the name of the block. The name of the block can be anything, but should include the file extension, which tells the linter how to process the current block. The linter checks the [`--ext` CLI option](../use/command-line-interface#--ext) to see if the current block should be linted and resolves `overrides` configs to check how to process the current block.
 
-It's up to the plugin to decide if it needs to return just one part, or multiple pieces. For example in the case of processing `.html` files, you might want to return just one item in the array by combining all scripts, but for `.md` file where each JavaScript block might be independent, you can return multiple items.
+It's up to the plugin to decide if it needs to return just one part of the non-JavaScript file or multiple pieces. For example in the case of processing `.html` files, you might want to return just one item in the array by combining all scripts. However, for `.md` files, because each JavaScript block might be independent, you can return multiple items.
 
 **The `postprocess` method** takes a two-dimensional array of arrays of lint messages and the filename. Each item in the input array corresponds to the part that was returned from the `preprocess` method. The `postprocess` method must adjust the locations of all errors to correspond to locations in the original, unprocessed code, and aggregate them into a single flat array and return it.
 
 Reported problems have the following location information:
 
 ```typescript
+// TODO: are these the messages?
 {
     line: number,
     column: number,
-
     endLine?: number,
     endColumn?: number
 }
 ```
 
-By default, ESLint will not perform autofixes when a processor is used, even when the `--fix` flag is enabled on the command line. To allow ESLint to autofix code when using your processor, you should take the following additional steps:
+By default, ESLint does not perform autofixes when a custom processor is used, even when the `--fix` flag is enabled on the command line. To allow ESLint to autofix code when using your processor, you should take the following additional steps:
 
-1. Update the `postprocess` method to additionally transform the `fix` property of reported problems. All autofixable problems will have a `fix` property, which is an object with the following schema:
+1. Update the `postprocess` method to additionally transform the `fix` property of reported problems. All autofixable problems have a `fix` property, which is an object with the following schema:
 
     ```js
     {
@@ -81,7 +82,9 @@ By default, ESLint will not perform autofixes when a processor is used, even whe
 
 2. Add a `supportsAutofix: true` property to the processor.
 
-You can have both rules and processors in a single plugin. You can also have multiple processors in one plugin.
+TODO: i don't understand what this next paragraph means
+
+You can have both rules and custom processors in a single plugin. You can also have multiple processors in one plugin.
 To support multiple extensions, add each one to the `processors` element and point them to the same object.
 
 ## Specifying Processor in Config Files
@@ -102,7 +105,11 @@ See [Specify a Processor](../use/configure/plugins#specify-a-processor) in the P
 
 ## File Extension-named Processor
 
-If a processor name starts with `.`, ESLint handles the processor as a **file extension-named processor** especially and applies the processor to the kind of files automatically. People don't need to specify the file extension-named processors in their config files.
+> TODO: i don't understand this line. is it saying that as long as the file extension named processor
+> is inclueded in the plugin, it runs?
+> Users don't need to specify the file extension-named processors in their config files.
+
+If a custom processor name starts with `.`, ESLint handles the processor as a **file extension-named processor**. ESLint applies the processor to files with that filename extension automatically. Users don't need to specify the file extension-named processors in their config files.
 
 For example:
 
@@ -110,7 +117,7 @@ For example:
 module.exports = {
     processors: {
         // This processor will be applied to `*.md` files automatically.
-        // Also, people can use this processor as "plugin-id/.md" explicitly.
+        // Also, you can use this processor as "plugin-id/.md" explicitly.
         ".md": {
             preprocess(text, filename) { /* ... */ },
             postprocess(messageLists, filename) { /* ... */ }
--- a/docs/src/extend/custom-processors.md
+++ b/docs/src/extend/custom-processors.md
@@ -57,23 +57,24 @@ Reported problems have the following location information in each lint message:
 
 ```typescript
 type LintMessage = {
+
+  /// The 1-based line number where the message occurs.
+  line: number;
+
    /// The 1-based column number where the message occurs.
   column: number;
 
-  /// The 1-based column number of the end location.
-  endColumn: number;
-
   /// The 1-based line number of the end location.
   endLine: number;
+
+  /// The 1-based column number of the end location.
+  endColumn: number;
 
   /// If `true`, this is a fatal error.
   fatal: boolean;
 
   /// Information for an autofix.
   fix: Fix;
-
-  /// The 1-based line number where the message occurs.
-  line: number;
 
   /// The error message.
   message: string;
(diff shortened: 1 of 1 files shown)


docs/src/extend/custom-processors.md
```diff
@@ -102,15 +105,19 @@ See [Specify a Processor](../use/configure/plugins#specify-a-processor) in the P
 
 ## File Extension-named Processor
 
-If a processor name starts with `.`, ESLint handles the processor as a **file extension-named processor** especially and applies the processor to the kind of files automatically. People don't need to specify the file extension-named processors in their config files.
+> TODO: i don't understand this line. is it saying that as long as the file extension named processor
+> is inclueded in the plugin, it runs?
+> Users don't need to specify the file extension-named processors in their config files.
+
+If a custom processor name starts with `.`, ESLint handles the processor as a **file extension-named processor**. ESLint applies the processor to files with that filename extension automatically. Users don't need to specify the file extension-named processors in their config files.
```
bpmutter, line 112: To reviewer: I have trouble understanding this line.

Is it saying that as long as the file extension named processor is included in the plugin, it runs? and therefore users don't need to specify the file extension-named processors in their config files?

would appreciate clarification so i can adjust phrasing accordingly 😄

docs/src/extend/custom-processors.md
```diff
@@ -102,15 +105,19 @@ See [Specify a Processor](../use/configure/plugins#specify-a-processor) in the P
 
 ## File Extension-named Processor
 
-If a processor name starts with `.`, ESLint handles the processor as a **file extension-named processor** especially and applies the processor to the kind of files automatically. People don't need to specify the file extension-named processors in their config files.
+> TODO: i don't understand this line. is it saying that as long as the file extension named processor
+> is inclueded in the plugin, it runs?
+> Users don't need to specify the file extension-named processors in their config files.
```
bpmutter, line 110:
```suggestion
```

docs/src/extend/custom-processors.md
```diff
@@ -102,15 +105,16 @@ See [Specify a Processor](../use/configure/plugins#specify-a-processor) in the P
 
 ## File Extension-named Processor
 
-If a processor name starts with `.`, ESLint handles the processor as a **file extension-named processor** especially and applies the processor to the kind of files automatically. People don't need to specify the file extension-named processors in their config files.
+
```
bpmutter, line 108:
```suggestion
```

docs/src/extend/custom-processors.md
```diff
@@ -102,15 +105,15 @@ See [Specify a Processor](../use/configure/plugins#specify-a-processor) in the P
 
 ## File Extension-named Processor
 
-If a processor name starts with `.`, ESLint handles the processor as a **file extension-named processor** especially and applies the processor to the kind of files automatically. People don't need to specify the file extension-named processors in their config files.
+If a custom processor name starts with `.`, ESLint handles the processor as a **file extension-named processor**. ESLint applies the processor to files with that filename extension automatically. Users don't need to specify the file extension-named processors in their config files.
```
bpmutter, line 108: To reviewer: I have trouble understanding this line.

Is it saying that as long as the file extension named processor is included in the plugin, it runs? and therefore users don't need to specify the file extension-named processors in their config files?

would appreciate clarification so i can adjust phrasing accordingly 😄

docs/src/extend/custom-processors.md
```diff
@@ -81,7 +82,9 @@ By default, ESLint will not perform autofixes when a processor is used, even whe
 
 2. Add a `supportsAutofix: true` property to the processor.
 
-You can have both rules and processors in a single plugin. You can also have multiple processors in one plugin.
+TODO: i don't understand what this next paragraph means
+
+You can have both rules and custom processors in a single plugin. You can also have multiple processors in one plugin.
```
bpmutter, line 87: To reviewer: I have trouble understanding this line. Could you explain more?

docs/src/extend/custom-processors.md
```diff
@@ -81,7 +82,9 @@ By default, ESLint will not perform autofixes when a processor is used, even whe
 
 2. Add a `supportsAutofix: true` property to the processor.
 
-You can have both rules and processors in a single plugin. You can also have multiple processors in one plugin.
+TODO: i don't understand what this next paragraph means
+
```
bpmutter, line 86:
```suggestion
```

docs/src/extend/custom-processors.md
```diff
@@ -46,27 +47,27 @@ module.exports = {
...
 
-It's up to the plugin to decide if it needs to return just one part, or multiple pieces. For example in the case of processing `.html` files, you might want to return just one item in the array by combining all scripts, but for `.md` file where each JavaScript block might be independent, you can return multiple items.
+It's up to the plugin to decide if it needs to return just one part of the non-JavaScript file or multiple pieces. For example in the case of processing `.html` files, you might want to return just one item in the array by combining all scripts. However, for `.md` files, because each JavaScript block might be independent, you can return multiple items.
 
 **The `postprocess` method** takes a two-dimensional array of arrays of lint messages and the filename. Each item in the input array corresponds to the part that was returned from the `preprocess` method. The `postprocess` method must adjust the locations of all errors to correspond to locations in the original, unprocessed code, and aggregate them into a single flat array and return it.
 
 Reported problems have the following location information:
 
 ```typescript
+// TODO: are these the messages?
```
nzakas, line 59: Yes, we should probably include the full message interface rather than just the location information here.

docs/src/extend/custom-processors.md
nzakas: We are trying to say that a single plugin can export rules, processors, or both. In other words, you don't need one plugin to distribute rules and another to distribute processors, you can just put them in the same one.

docs/src/extend/custom-processors.md
nzakas: Originally, processors were enabled automatically when a plugin containing them was loaded. These auto-enabled processors were named according to file extension, such as:

```js
module.exports = {
    processors: {
        ".md": { /* processor */ }
    }
};
```

So, if the name begins with a dot, we assume it should be auto-enabled.

The way we want people to name processors today is with a descriptive name, like so:

```js
module.exports = {
    processors: {
        "markdown": { /* processor */ }
    }
};
```

Because this name doesn't begin with a dot, we know it shouldn't be applied automatically.

nzakas (commented):
Nice cleanup! Hopefully my answers can help even more.

docs/src/extend/custom-processors.md
bpmutter: > To support multiple extensions, add each one to the `processors` element and point them to the same object.

specifically this part of that paragraph i am still having trouble understanding. could you explain a bit more?

docs/src/extend/custom-processors.md
bpmutter: in latest version i included a typescript definition. you think that's ok?

bpmutter: ok, makes sense to me. i think just writing out the TypeScript interface would work. sound good to you?

also, based upon looking at the source code, it seems like there's some properties of this object that we're not documenting here - https://github.com/eslint/eslint/blob/main/lib/shared/types.js#L95 

should those be included in the docs?

bpmutter:
@nzakas implemented your feedback and added a few additional questions in comment threads

docs/src/extend/custom-processors.md
nzakas: I'm actually not sure if postprocess receives that full interface or not. @btmills do you know?

I think including the interface as TypeScript is fine.

docs/src/extend/custom-processors.md
nzakas: Suppose you want the same processor to work on `.md` and `.mdx` files, you might end up configuring your plugin like this:

```js
const yourProcessor = {};

module.exports = {
    processors: {
        ".md": yourProcessor,
        ".mdx": yourProcessor
    }
};
```

bpmutter: ah got it! adding new paragraph and small example

snitin315 (approved):
LGTM

docs/src/extend/custom-processors.md
```diff
@@ -124,3 +124,16 @@ module.exports = {
     }
 }
 ```
+
+You can also use the same custom processor with multiple filename extensions. The following example shows using same processor for both `.md` and `.mdx` files:
+
+```js
+module.exports = {
```
nzakas, line 131: It would probably be helpful to include something like this at the top of this example:

```js
const myCustomProcessor = { /* processor methods */ };
```

Just so `myCustomProcessor` is defined within the same example.

docs/src/extend/custom-processors.md
```diff
@@ -102,19 +101,39 @@ See [Specify a Processor](../use/configure/plugins#specify-a-processor) in the P
...
+            preprocess(text, filename) { /* ... */ },
+            postprocess(messageLists, filename) { /* ... */ }
+        }
+    }
+}
+```
+
+You can also use the same custom processor with multiple filename extensions. The following example shows using same processor for both `.md` and `.mdx` files:
+
+```js
```
bpmutter, line 130:
```suggestion
```js
const myCustomProcessor = { /* processor methods */ };

```

docs/src/extend/custom-processors.md
bpmutter: added

docs/src/extend/custom-processors.md
btmills: It does! In case it helps, here are a couple examples that I just pulled from linting https://github.com/eslint/eslint-plugin-markdown/tree/ef7dcdccfb94ac7e657a6c66998ce8831f3e58fd/examples/typescript:

```js
[
  {
    ruleId: '@typescript-eslint/ban-types',
    severity: 2,
    message: "Don't use `String` as a type. Use string instead",
    line: 1,
    column: 22,
    nodeType: 'Identifier',
    messageId: 'bannedTypeMessage',
    endLine: 1,
    endColumn: 28,
    fix: { range: [ 21, 27 ], text: 'string' }
  },
  {
    ruleId: '@typescript-eslint/no-explicit-any',
    severity: 1,
    message: 'Unexpected any. Specify a different type.',
    line: 5,
    column: 13,
    nodeType: 'TSAnyKeyword',
    messageId: 'unexpectedAny',
    endLine: 5,
    endColumn: 16,
    suggestions: [
      {
        messageId: 'suggestUnknown',
        fix: { range: [ 82, 85 ], text: 'unknown' },
        desc: 'Use `unknown` instead, this will force you to explicitly, and safely assert the type is correct.'
      },
      {
        messageId: 'suggestNever',
        fix: { range: [ 82, 85 ], text: 'never' },
        desc: "Use `never` instead, this is useful when instantiating generic type parameters that you don't need to know the type of."
      }
    ]
  }
]
```

docs/src/extend/custom-processors.md
```diff
@@ -46,29 +47,28 @@ module.exports = {
 
 **The `preprocess` method** takes the file contents and filename as arguments, and returns an array of code blocks to lint. The code blocks will be linted separately but still be registered to the filename.
 
-A code block has two properties `text` and `filename`; the `text` property is the content of the block and the `filename` property is the name of the block. Name of the block can be anything, but should include the file extension, that would tell the linter how to process the current block. The linter will check [`--ext` CLI option](../use/command-line-interface#--ext) to see if the current block should be linted, and resolve `overrides` configs to check how to process the current block.
+A code block has two properties `text` and `filename`. The `text` property is the content of the block and the `filename` property is the name of the block. The name of the block can be anything, but should include the file extension, which tells the linter how to process the current block. The linter checks the [`--ext` CLI option](../use/command-line-interface#--ext) to see if the current block should be linted and resolves `overrides` configs to check how to process the current block.
```
btmills, line 50: The last sentence here about checking `--ext` describes the pre-ESLint v6 behavior. That still happens, but it also checks extensions in `overrides` patterns now. That's described in https://eslint.org/docs/latest/use/configure/plugins#specify-a-processor.

If updating this is out of scope for the change you're intending here, feel free to disregard.

docs/src/extend/custom-processors.md
bpmutter: thanks brandon. i will provide a full interface definition here in a markdown list like we do throughout the docs.

however, like we talked about in https://github.com/eslint/eslint/discussions/16557#discussioncomment-4814143, i think well annotated TS definitions and docs generated from them will be great for replacing manually created/maintained reference docs like this.

btmills: Agreed! It's too easy for the implementation to drift from these docs if they're manual.

docs/src/extend/custom-processors.md
nzakas: Nice catch! I couldn't remember if we had removed that functionality or not.

docs/src/extend/custom-processors.md
bpmutter: added the TS type in [239b615](https://github.com/eslint/eslint/pull/16838/commits/239b615eacef0e4b77fb4a6c23d2e1bb1fab5afa). mapped the [JSDoc type](https://github.com/eslint/eslint/blob/main/lib/shared/types.js#L95) to a TS type. 

however this is slightly different from the examples that you included above, @btmills. any idea what might be going on with this?

docs/src/extend/custom-processors.md
```diff
@@ -46,29 +47,65 @@ module.exports = {
...
 
-It's up to the plugin to decide if it needs to return just one part, or multiple pieces. For example in the case of processing `.html` files, you might want to return just one item in the array by combining all scripts, but for `.md` file where each JavaScript block might be independent, you can return multiple items.
+It's up to the plugin to decide if it needs to return just one part of the non-JavaScript file or multiple pieces. For example in the case of processing `.html` files, you might want to return just one item in the array by combining all scripts. However, for `.md` files, because each JavaScript block might be independent, you can return multiple items.
```
harish-sethuraman, line 52: however followed by because doesn't sound well? is it fine? 
```suggestion
It's up to the plugin to decide if it needs to return just one part of the non-JavaScript file or multiple pieces. For example in the case of processing `.html` files, you might want to return just one item in the array by combining all scripts. However, for `.md` files, you can return multiple items, because each JavaScript block might be independent.
```

bpmutter:
```suggestion
It's up to the plugin to decide if it needs to return just one part of the non-JavaScript file or multiple pieces. For example in the case of processing `.html` files, you might want to return just one item in the array by combining all scripts. However, for `.md` files, you can return multiple items because each JavaScript block might be independent.
```

docs/src/extend/custom-processors.md
```diff
@@ -46,29 +47,65 @@ module.exports = {
...
+  endLine: number;
+
+  /// If `true`, this is a fatal error.
+  fatal: boolean;
+
+  /// Information for an autofix.
+  fix: Fix;
+
+  /// The 1-based line number where the message occurs.
+  line: number;
```
bpmutter, line 76:
```suggestion
type LintMessage = {

  /// The 1-based line number where the message occurs.
  line: number;

   /// The 1-based column number where the message occurs.
  column: number;

  /// The 1-based line number of the end location.
  endLine: number;

  /// The 1-based column number of the end location.
  endColumn: number;

  /// If `true`, this is a fatal error.
  fatal: boolean;

  /// Information for an autofix.
  fix: Fix;
```

docs/src/extend/custom-processors.md
bpmutter: i don't fully understand. would one of you mind providing a bit more info or a draft of what the copy should say here?

bpmutter: thank you for clarifying!

docs/src/extend/custom-processors.md
bpmutter: @btmills, following up on the above question

btmills: We don't (yet) run a type checker on our codebase, so there's currently no automatic enforcement that the implementation does what the type annotations say it will. The JSDoc types are more like documentation at the moment. 

As for what's actually correct, I opened https://github.com/eslint/eslint/issues/16968 to summarize my findings and ask a couple open questions. To enable you to move forward, what you put in 239b615 is good for now. As part of #16968, I can come back and update it to make sure it's fully in sync with whatever changes we make there. But you're right - the real answer is extracting types automatically to prevent drift.

docs/src/extend/custom-processors.md
btmills: To expand on what nzakas said and give an example, with the config below, to use the Markdown plugin prior to ESLint v6, one had to pass `--ext .js,.md`, or ESLint would only look at `.js` files. Now, ESLint will automatically see there's an `overrides` pattern for `.md` files and include them in the file traversal. One can still pass `--ext` to override that automatic behavior, but it in normal cases ESLint will do the right thing automatically.

```
// .eslintrc.js
module.exports = {
    plugins: ["markdown"],
    overrides: [
        {
            files: ["**/*.md"],
        }
    ]
};
```

docs/src/extend/custom-processors.md
```diff
@@ -102,19 +139,41 @@ See [Specify a Processor](../use/configure/plugins#specify-a-processor) in the P
...
 
-If a processor name starts with `.`, ESLint handles the processor as a **file extension-named processor** especially and applies the processor to the kind of files automatically. People don't need to specify the file extension-named processors in their config files.
+If a custom processor name starts with `.`, ESLint handles the processor as a **file extension-named processor**. ESLint applies the processor to files with that filename extension automatically. Users don't need to specify the file extension-named processors in their config files.
 
 For example:
 
 ```js
 module.exports = {
     processors: {
         // This processor will be applied to `*.md` files automatically.
```
nzakas, line 149: @btmills just double-checking: is this still true? I seem to recall we might have removed automatically applied processors? Or maybe I'm just imagining it?

nzakas (approved):
LGTM. Just leaving open to ensure @btmills' comments are incorporated.

harish-sethuraman (approved):
LGTM leaving open for unresolved comments

docs/src/extend/custom-processors.md
btmills: It's at least still true under `.eslintrc` ([implementation](https://github.com/eslint/eslintrc/blob/bd17b5a39b4b81467e6ad2850eab8f63a46afa8a/lib/config-array-factory.js#L1112-L1147), [tests](https://github.com/eslint/eslint/blob/75df535681d15d7d685468d637945a200301f9ee/tests/lib/eslint/eslint.js#L2947)). I don't know for sure that it's not part of flat config, but the tests all [explicitly specify `processor`](https://github.com/eslint/eslint/blob/75df535681d15d7d685468d637945a200301f9ee/tests/lib/eslint/flat-eslint.js#L2624), and I can't find evidence of the back compat logic being carried over.

docs/src/extend/custom-processors.md
```diff
@@ -102,19 +139,41 @@ See [Specify a Processor](../use/configure/plugins#specify-a-processor) in the P
...
+        // in your configuration as "plugin-id/markdown".
+       "markdown": {
+            preprocess(text, filename) { /* ... */ },
+            postprocess(messageLists, filename) { /* ... */ }
+        }
+    }
+}
+```
+
+You can also use the same custom processor with multiple filename extensions. The following example shows using same processor for both `.md` and `.mdx` files:
```
btmills, line 166: Just a quick missing word fix.

```suggestion
You can also use the same custom processor with multiple filename extensions. The following example shows using the same processor for both `.md` and `.mdx` files:
```

docs/src/extend/custom-processors.md
bpmutter: thanks for clarifying. will take no further action here.

docs/src/extend/custom-processors.md
nzakas: I don't think we need anything further.

nzakas (approved):
This looks ready to go. We can always adjust as needed.

PR #16846, opened 2023-02-01
Fixes reportUnusedDisableDirectives example

laurens-dg:
reportUnusedDisableDirectives must be any of "error" | "warn" | "off", true is not a valid option.

#### Prerequisites checklist

- [X ] I have read the [contributing guidelines](https://github.com/eslint/eslint/blob/HEAD/CONTRIBUTING.md).

[ X] Documentation update
[ ] Add autofix to a rule
[ ] Add a CLI option
[ ] Add something to the core
[ ] Other, please explain:

Using the snippet in the documentation, as one does, currently throws an 'Error: Invalid Options'. This change fixes that.

### DIFF
--- a/docs/src/use/configure/rules.md
+++ b/docs/src/use/configure/rules.md
@@ -298,7 +298,7 @@ To report unused `eslint-disable` comments, use the `reportUnusedDisableDirectiv
 ```json
 {
   "rules": {...},
-  "reportUnusedDisableDirectives": true
+  "reportUnusedDisableDirectives": "warn"
 }
 ```
 


snitin315 (changes requested):
> this is actually a valid option, just like the eslint org's shared config: > https://github.com/eslint/eslint/blob/ad38d77102d6fe30cfa92c831174f178bb35c88b/packages/eslint-config-eslint/default.yml#L1

> can you share how did you get the error as more details as possible, e.g what config you are using(eslintrc or flat config)?

Same as https://github.com/eslint/eslint/pull/16847#issuecomment-1411702344

nzakas:
"warn" | "error" | "off" are only valid options at the API level, not at the config level. Inside a config we can only have a boolean value currently.

PR #16929, opened 2023-02-24
feat: Enhance no-tabs allowIndentationTabs option + add autofix

laverdet:
#### Prerequisites checklist

[ ] Documentation update
[X] Add autofix to a rule
[ ] Add a CLI option
[ ] Add something to the core
[ ] Other, please explain:

no-tabs

[ ] Generate more warnings
[X] Generate fewer warnings
[X] Implement autofix
[ ] Implement suggestions

[ ] A new option
[X] A new default behavior
[ ] Other

note: Here "\t" means literally a tab character
```js
// \tconst unused = 1;
```

This rule raises an error since a tab is encountered in the source text. An exception is carved out for indentation tabs under the `allowIndentationTabs` option but since a comment is non-indentation text it will still fire in this case.

After this change single-line comment markers "//" will not be considered non-indentation text. The reasoning for this is tools which "comment out" code (VS Code, vim plugins, lots of others) will generally prefix all code with "// ". After commenting out the code, previously-valid indentation characters will now be flagged as invalid.

If `allowIndentationTabs` is enabled we first make a `Map` of all line-style comments from the line to column. When a tab character is encountered we use this information to "splice out" the comment marker so that the existing indentation detection code works.

I also added an auto-fix which replaces each tab with a single space. This has a risk of changing program meaning because `const a = "\t"` (again, an actual tab character not an escape sequence) would become `const a = " "`. I suspect such cases are rare but I'm happy to leave out the autofix.

### DIFF
--- a/lib/rules/no-tabs.js
+++ b/lib/rules/no-tabs.js
@@ -48,12 +48,42 @@ module.exports = {
 
         return {
             Program(node) {
+                const lineComments = new Map();
+
+                /*
+                 * If indentation tabs are allowed then make a map of a line comments by their line
+                 * and column. We will use this to splice out the comment marker in order to allow
+                 * for tabs in "commented out" code.
+                 */
+                if (allowIndentationTabs) {
+                    for (const comment of sourceCode.getAllComments()) {
+                        if (comment.type === "Line") {
+                            lineComments.set(comment.loc.start.line, comment.loc.start.column);
+                        }
+                    }
+                }
+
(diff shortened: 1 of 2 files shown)


Rec0iL99:
Hi @laverdet, thanks for the pr. As per our rule policies, [stylistic rules are frozen](https://github.com/eslint/eslint#stylistic-rule-updates), so we can't accept this change to the `no-tabs` rule. You can copy the existing `no-tabs` rule and create a custom rule to fit your needs.

PR #17023, opened 2023-03-27
fix: correctly handle `null` default config in `RuleTester`

bradzacher:
#### Prerequisites checklist

[x] Bug fix

I was playing around with RuleTester and I broke the test by changing the `setDefaultConfig` function to use `defaultConfig = { rules: {}, ...config }`.

Previously the method threw because `defaultConfig.rules` will throw due to property access on `null`. The test was passing before because it wasn't asserting the error message.

This PR simply updates the method so that it handles `null` explicitly and updates the test as appropriate.

### DIFF
--- a/lib/rule-tester/flat-rule-tester.js
+++ b/lib/rule-tester/flat-rule-tester.js
@@ -345,7 +345,7 @@ class FlatRuleTester {
      * @returns {void}
      */
     static setDefaultConfig(config) {
-        if (typeof config !== "object") {
+        if (typeof config !== "object" || config === null) {
             throw new TypeError("FlatRuleTester.setDefaultConfig: config must be an object");
         }
         sharedDefaultConfig = config;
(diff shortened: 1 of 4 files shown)


fasttime (commented):
Thanks for the PR @bradzacher. We should apply the same fix to [`FlatRuleTester`](https://github.com/eslint/eslint/blob/main/lib/rule-tester/flat-rule-tester.js#L347-L355).

nzakas:
Nice catch. Agree with @fasttime, this change should also be applied to `FlatRuleTester`.

fasttime (approved):
LGTM. Leaving open in case @nzakas or someone else wants to take another look.

snitin315 (approved):
LGTM. I'll leave it open for some time in case anyone else wants to take a look.

mdjermanovic (approved):
LGTM, thanks!

PR #11027, opened 2018-10-28
Build: Add karma-chrome-launcher support

aladdin-add:
**What is the purpose of this pull request? (put an "X" next to item)**

[ ] Documentation update
[ ] Bug fix ([template](https://raw.githubusercontent.com/eslint/eslint/master/templates/bug-report.md))
[ ] New rule ([template](https://raw.githubusercontent.com/eslint/eslint/master/templates/rule-proposal.md))
[ ] Changes an existing rule ([template](https://raw.githubusercontent.com/eslint/eslint/master/templates/rule-change-proposal.md))
[ ] Add autofixing to a rule
[ ] Add a CLI option
[ ] Add something to the core
[x ] Other, please explain:

**What changes did you make? (Give an overview)**
the change was merged in https://github.com/eslint/eslint/pull/10898, but was reverted in https://github.com/eslint/eslint/pull/10973, due to a release build failing on Jenkins.

the change added it back again, and added `puppeteer ` in dev, to fix the build failing.

**Is there anything you'd like reviewers to focus on?**

what can we do to make sure it fixed the issue? (e.g. run on Jenkins)

PR #17126, opened 2023-04-26
test: replace Karma with Webdriver.IO

christian-bromann:
The current test framework for browser testing (Karma) is not maintained anymore and WebdriverIO provides a more modern stack that allows to test in different browser.

This patch replaces these test frameworks.

fixes: #17009

#### Prerequisites checklist

[ ] Documentation update
[ ] Add autofix to a rule
[ ] Add a CLI option
[ ] Add something to the core
[ ] Other, please explain:

This is still a work in progress PR but I would like the maintainer to take a quick look and verify that they are ok with the change.


lib/linter/rules.js
```diff
@@ -56,7 +56,7 @@ class Rules {
      */
     get(ruleId) {
         if (typeof this._rules[ruleId] === "string") {
-            this.define(ruleId, require(this._rules[ruleId]));
+            this.define(ruleId, require(`../rules/${this._rules[ruleId]}`));
```
aladdin-add, line 59: Can you explain why changing it? a rule may be loaded from a third-party plugin, so this won't always work.

christian-bromann: I see. Transpiling dynamic CJS imports into ESM is kinda of tricky so this was for now the only way to workaround it. I was able to verify that this code is not touched during the tests so it shouldn't be modified. I will work on reverting it.

package.json
```diff
@@ -154,10 +155,14 @@
     "puppeteer": "^13.7.0",
```
aladdin-add, line 155: question: do we need switch puppeteer to something like playwright too? 😄

mdjermanovic:
I converted this to draft to prevent accidental merging since it's still work in progress and some tests are `.skip`-ed,

package.json
christian-bromann: No, WebdriverIO does all the browser automation. I wonder if having Puppeteer here is still needed. I will check.

christian-bromann:
> Can we still use webpack

WebdriverIO uses Vite to power the component tests (including mocking etc.). Let me check if I can just bundle with webpack and import the compiled files. This would make the setup also simpler.

Rec0iL99:
Don't close. Work in progress.

christian-bromann:
Sorry, haven't forgotten about this. Will try to put some time in this or next week to finish it.

aladdin-add:
@christian-bromann no worries, I've removed it.

christian-bromann:
Made some progress on this:

- removed the changes in EsLint core
- use the webpack build artifact within the browser test
- removed unused puppeteer dependency
- used WebdriverIO based mocking for two tests as Sinon can't spy on ESM modules

Note: for the mocking to work I had to use import syntax. I will have to push an enhancement to WebdriverIO to switch back to require but I wonder if this test is also run in Node.js because if not I can just make everything an import statement.

christian-bromann:
The build is failing due to:

```
/home/runner/work/eslint/eslint/tests/lib/linter/linter.js
  12:1  error  Parsing error: 'import' and 'export' may appear only with 'sourceType: module'
```

Unfortunately WebdriverIOs mocking features only work when using `import`. Since the `linter.js` test is run within the browser I wonder if we can move the file into a directory for browser tests and change the file to use ESM. What do you think?

mdjermanovic:
> Note: for the mocking to work I had to use import syntax. I will have to push an enhancement to WebdriverIO to switch back to require but I wonder if this test is also run in Node.js because if not I can just make everything an import statement

This test file is run in Node.js too. Perhaps we could make another bundle from `tests/lib/linter/linter.js` entry point and then run that file in a browser?

nzakas:
@christian-bromann are you still working on this?

christian-bromann:
@nzakas yes, sorry for my delay, still many competing tasks but will try to get back to it in the following days

BYK:
@christian-bromann do you need any help with pushing this over the finish line? Looks like it is close so I'd like to help if you need any.

christian-bromann:
Thanks for pinging and sorry for the delay. I have been 2 weeks on vacation. Put it on my todo for this week.

christian-bromann:
@nzakas @BYK this is ready for review now

tests/lib/linter/linter.js
kecrily, line 1: Should be rename to `linter.mjs`

christian-bromann: Thanks for the review, made the changes

mdjermanovic:
We're running `mocha` on `.js` files only, so after switching the linter test file to `.mjs`, those tests won't be run in Node.

https://github.com/eslint/eslint/blob/c6f8cd0d62e4a3c314c6860ff367490bbd05325a/Makefile.js#L79

Why was it necessary to convert the file to ESM?

kecrily:
Because of linter and test errors

See https://github.com/eslint/eslint/actions/runs/5482202414/jobs/10005549763?pr=17126 and https://github.com/eslint/eslint/actions/runs/5482202414/jobs/10005549380?pr=17126

kecrily:
I also found missing compatibility code to use both import and require in the same file.

```
import { createRequire } from "module";
const require = createRequire(import.meta.url);

```

But when I added it locally and tested it, I got some test errors. Need to investigate further

image

mdjermanovic:
Why do we need to `import` modules? That seems like an extensive change in a test file whose primary purpose is to test how the core linter module works in Node.js.

christian-bromann:
> Why do we need to `import` modules? That seems like an extensive change in a test file whose primary purpose is to test how the core linter module works in Node.js.

These tests are run in the browser. It is currently a limitation of WebdriverIO. I didn't know these tests were also run in Mocha which makes everything a bit more complicated. I will see what I can do.

mdjermanovic:
Could we maybe simply webpack `tests/lib/linter/linter.js` and load the bundle in the browser as `specs`?

Here I made a [test repo](https://github.com/mdjermanovic/linter-browser-test) with the bundle (`linter.test.js`) where you can run it locally (`npm run mocha`), in browser with karma (`npm run karma`), or directly in browser (open https://mdjermanovic.github.io/linter-browser-test/).

nzakas:
Agree with others, we should not be changing `linter.js` to ESM -- we have a lot of changes coming and going in there, and this would complicate testing without providing much additional value.

And yes, we need all of the tests to stay in `linter.js` to make it easier as we implement the new config system.

christian-bromann:
Thank you all for your feedback!

I made it work 🎉  with the most minimal changes to the `linter.js` file as possible. I wrote a custom Vite plugin that would help me to dynamically update the import from the `linter.js` in `src` to `build`. I had to skip 2 tests because the way they spy on the module is not possible in the browser. I am not sure how Webpack was configured in Karma to make this possible but I think this is fine for now.

Let me know what you think!

wdio.conf.js
```diff
@@ -0,0 +1,395 @@
...
+     *
+     * ===================
+     * Test Configurations
+     * ===================
+     * Define all options that are relevant for the WebdriverIO instance here
+     *
+     * Level of logging verbosity: trace | debug | info | warn | error | silent
+     */
+    logLevel: "info",
+    outputDir: "./logs",
```
aladdin-add, line 113: nits: I prefer something like `"./wdio-logs"`, to avoid conflicting with other tools. :)

wdio.conf.js
```diff
@@ -0,0 +1,395 @@
...
+     */
+    runner: ["browser", {
+        viteConfig: {
+            resolve: {
+                alias: {
+                    util: "rollup-plugin-node-polyfills/polyfills/util",
+                    path: "rollup-plugin-node-polyfills/polyfills/path",
+                    assert: "rollup-plugin-node-polyfills/polyfills/assert"
+                }
+            },
```
aladdin-add, line 22: not sure this is needed.  IMHO, the webpack bundled file should have included node.js built-in modules, as configured: https://github.com/eslint/eslint/blob/8bcbf11b6050418262ffa8e0ca37f365ae92e7ce/webpack.config.js#L47

christian-bromann: WebdriverIOs component testing framework uses Vite under the hood, which allows Rollup plugins.

aladdin-add: well, I see!

the plugin was applied to `/lib/linter/linter.js`, but not `tests/lib/linter/linter.js`. 😄

aladdin-add (approved):
LGTM, thanks!

aladdin-add (changes requested):
the ci was failing as some dev deps were not compat with node.js v12.( it has been EOL, but eslint v8 still supports.)

I think we can merge it when we drop node.js v12 support.

kecrily:
But this is a devDependencies change, and I don't think it will break ESLint's support for node 12.

aladdin-add:
yes, but it causes ci to fail and we are not able to merge it.

christian-bromann:
> the ci was failing as some dev deps were not compat with node.js v12.( it has been EOL, but eslint v8 still supports.)

The [migration docs](https://eslint.org/docs/latest/use/migrate-to-8.0.0) state that v8 _"is officially dropping support for [Node.js 10, 13, 15] starting with ESLint v8.0.0"_. Am I missing something?

package.json
```diff
@@ -101,10 +102,16 @@
...
+    "@wdio/browser-runner": "^8.10.4",
+    "@wdio/cli": "^8.10.4",
+    "@wdio/concise-reporter": "^8.10.4",
+    "@wdio/globals": "^8.11.3",
+    "@wdio/mocha-framework": "^8.10.4",
     "babel-loader": "^8.0.5",
     "c8": "^7.12.0",
     "chai": "^4.0.1",
     "cheerio": "^0.22.0",
+    "chromedriver": "^114.0.0",
```
aladdin-add, line 114:
```suggestion
    "chromedriver": "^106.0.0",
```

it should be fixed if back to the version: https://github.com/aladdin-add/eslint/actions/runs/5571225831/jobs/10176049189?pr=21

christian-bromann:
@aladdin-add I reverted the version update of Chromedriver and added flag so that it detects the right version in CI. However this could confuse contributor trying to run the test and don't have the right Chromedriver version installed. I am working to get this all embedded into WebdriverIO so no one has to worry about the driver.

christian-bromann:
I am working on some WebdriverIO things to get the issue in the browser test elevated.

christian-bromann:
> I'm curious, why the problem with spying on ES modules appears here, since this test doesn't seem to use any ES modules?

These tests run in the browser so they are all ES modules. I am not sure how WebPack transpiled the code to not have this error thrown but basically all imported modules in the browser are sealed and can't be changed. However I pushed a commit that changes the tests so that it should work in both environments. Also I made some changes to WebdriverIO to cope some performance issues. I think it should all pass now 🤞

aladdin-add (approved):
✅all green, thanks!

just waiting for @mdjermanovic and other members to take another look before merging.

nzakas (approved):
LGTM. Thanks for all of your work on this.

Just waiting for @mdjermanovic to re-review.

.github/workflows/ci.yml
```diff
@@ -74,7 +74,15 @@ jobs:
...
-      run: node Makefile karma
+      run: node Makefile wdio
     - name: Fuzz Test
       run: node Makefile fuzz
+    - uses: actions/upload-artifact@v3
+      if: failure()
+      with:
+        name: logs
+        path: |
+          wdio-logs/*.log
```
snitin315, line 88: How will we download & see these logs?

christian-bromann: > How will we download & see these logs?

In the summary of a job under `Artifacts`, see picture below (bottom):
Screenshot 2023-07-26 at 21 27 09

https://github.com/eslint/eslint/actions/runs/5583652074

mdjermanovic:
I'm getting this locally:

```
C:\milos\tmp\wsdio-eslint>npm run wdio

> eslint@8.45.0 wdio
> wdio run ./wdio.conf.js


Execution of 1 workers started at 2023-07-27T12:26:28.122Z

[0-0] RUNNING in chrome - file:///C:/milos/tmp/wsdio-eslint/tests/lib/linter/linter.js

DevTools listening on ws://127.0.0.1:64612/devtools/browser/99c26fcd-2ca5-48b8-8dcd-a5fcdd775b26

DevTools listening on ws://127.0.0.1:64618/devtools/browser/f38249c5-2e5e-4959-812e-a00d2f3fc2d4

DevTools listening on ws://127.0.0.1:64624/devtools/browser/33a02012-1c10-45e4-a678-6ee2d6a551cd

DevTools listening on ws://127.0.0.1:64630/devtools/browser/9b25ff2d-d847-41b5-a43a-b195ac9ca8d6
[0-0] FAILED in chrome - file:///C:/milos/tmp/wsdio-eslint/tests/lib/linter/linter.js

 "concise" Reporter:
========= Your concise report ==========
undefined
❌ Failed to setup tests, no tests found


Spec Files:      0 passed, 1 failed, 1 total (100% completed) in 00:00:06
```

christian-bromann:
@kecrily @mdjermanovic I updated the branch, all wdio dependencies and made the path to the spec file more explicit. This should resolve the issue for you (I assume you've been running the command outside of the root dir). Let me know if there is anything else I can do.

nzakas:
@christian-bromann the test is now failing on the PR

christian-bromann:
@nzakas pushed a fix for it. I missed to remove the Chromedriver dependencies which is now all built-into WebdriverIO. Due to differences between CI and my local machine it passed when I tested it but failed in CI. Should be good to go now.

package.json
```diff
@@ -29,7 +29,8 @@
     "test": "node Makefile.js test",
     "test:cli": "mocha",
     "test:fuzz": "node Makefile.js fuzz",
-    "test:performance": "node Makefile.js perf"
+    "test:performance": "node Makefile.js perf",
+    "wdio": "wdio run ./wdio.conf.js"
```
kecrily, line 33: It doesn't match [our Package.json Conventions](https://github.com/eslint/eslint/blob/main/docs/src/contribute/package-json-conventions.md), can you modify it?

christian-bromann: ahh, good catch 👍  removed.

aladdin-add:
seems working fine on my macos! 👍

kecrily (approved):
LGTM

kecrily (approved):
LGTM

aladdin-add:
merged, thanks for contributing! 👍

christian-bromann:
Thanks folks for the support 👍 don't hesitate to ping me if you run into any issues with the tests.

mdjermanovic:
Question: Does this expect that Chrome is already installed and what happens if it isn't? Asking because we previously had issues with the browser test on our Jenkins server that we use for releases, and then we added `puppeteer` to fix the problem (https://github.com/eslint/eslint/pull/11027). But, `puppeteer` is removed now, so I'm worried if this test will work on our Jenkins server.

christian-bromann:
Currently WebdriverIO will prefer the browser installed on the host OS, if no Chrome is installed it will automatically install latest stable release of Chrome. If you set `browserVersion: 'stable'` in your capabilities it will always download the latest stable release.

PR #17212, opened 2023-05-23
feat: Support custom severity when reporting unused disable directives

bmish:
#### Prerequisites checklist

[ ] Documentation update
[ ] Add autofix to a rule
[X] Add a CLI option
[ ] Add something to the core
[ ] Other, please explain:

Implements phase 1 (non-breaking changes) of this RFC:
* https://github.com/eslint/rfcs/pull/100

That includes:

* Changes the `reportUnusedDisableDirectives` option in config files to support severity strings (new behavior) in addition booleans (existing behavior)
* Adds a new CLI option `--report-unused-disable-directives-severity <severity>`, in addition to the existing CLI option `--report-unused-disable` (untouched)

This allows customizing the severity with which unused disabled directives are reported.

TODO:
- [ ] Wait for RFC to be merged. This is a draft PR until the RFC is merged.


lib/eslint/eslint-helpers.js
```diff
@@ -675,7 +675,7 @@ function processOptions({
     overrideConfig = null,
     overrideConfigFile = null,
     plugins = {},
-    reportUnusedDisableDirectives = null, // ← should be null by default because if it's a string then it overrides the 'reportUnusedDisableDirectives' setting in config files. And we cannot use `overrideConfig.reportUnusedDisableDirectives` instead because we cannot configure the `error` severity with that.
+    reportUnusedDisableDirectives = null, // ← should be null by default because if it's a string then it overrides the 'reportUnusedDisableDirectives' setting in config files. And we cannot use `overrideConfig.reportUnusedDisableDirectives` instead because we cannot configure the `error` severity with that. TODO: should anything change based on this comment?
```
mdjermanovic, line 678: As for the first sentence, I think nothing should change.

The second sentence, I believe, explains why this option is needed. After this change, integrations and our CLI will be able to set this value in `overrideConfig`, so we might consider dropping this option.

lib/config/flat-config-schema.js
```diff
@@ -222,6 +222,16 @@ const booleanSchema = {
...
 };
 
+/** @type {ObjectPropertySchema} */
+const booleanOrStringSchema = {
+    merge: "replace",
+    validate(value) {
+        if (typeof value !== "string" && typeof value !== "boolean") {
+            throw new TypeError("Expected a string or a boolean.");
+        }
+    }
```
mdjermanovic, line 232: If it's a string, we should also check if the value is one of the allowed values.

lib/eslint/eslint-helpers.js
bmish: I'll look into dropping this option. It wouldn't be a breaking change, would it?

mdjermanovic: I think it wouldn't be a breaking change at this point since it's in the `FlatESLint` class, and I'm in favor of removing constructor options that turned out to be redundant with config options, but would like @nzakas's opinion too.

bmish: I agree with removing redundant options. I added a note to the RFC about this:

> Phase 2: Any breaking changes for flat config users (`warn` by default, remove redundant `reportUnusedDisableDirectives` option from `processOptions()`), suitable for a major release or potentially a minor release since the flat config is not considered stable yet.

nzakas:
@bmish what's the status on this?

bmish:
I need to address some feedback still, but this is ready for review. The RFC https://github.com/eslint/rfcs/pull/100 has been approved and will be merged soon, and this PR is not a breaking change, so we should be able to move forward with it once the review process is completed.

nzakas:
Preparing for v9, so leaving this open.

conf/default-cli-options.js
```diff
@@ -28,5 +28,6 @@ module.exports = {
     fix: false,
     allowInlineConfig: true,
     reportUnusedDisableDirectives: void 0,
+    reportUnusedDisableDirectivesSeverity: void 0,
```
mdjermanovic, line 31: These are actually default options for the `CLIEngine` class, not the ESLint CLI. Since that class does not have this option, there's no need to set this here.

lib/options.js
```diff
@@ -47,7 +47,7 @@ const optionator = require("optionator");
  * @property {Object} [parserOptions] Specify parser options
  * @property {string[]} [plugin] Specify plugins
  * @property {string} [printConfig] Print the configuration for the given file
- * @property {boolean | undefined} reportUnusedDisableDirectives Adds reported errors for unused eslint-disable and eslint-enable directives
+ * @property {boolean | string | undefined} reportUnusedDisableDirectives Adds reported errors for unused eslint-disable and eslint-enable directives
```
mdjermanovic, line 50: I'm guessing this is from an earlier version of the RFC. Now this option remains the same, but we should specify the new one.

bmish: Fixed.

lib/linter/linter.js
```diff
@@ -671,9 +672,18 @@ function normalizeVerifyOptions(providedOptions, config) {
...
     if (typeof reportUnusedDisableDirectives !== "string") {
-        reportUnusedDisableDirectives =
-            linterOptions.reportUnusedDisableDirectives
-                ? "warn" : "off";
+
+        // Handle either a boolean or severity string in linterOptions.reportUnusedDisableDirectives.
+        if (typeof linterOptions.reportUnusedDisableDirectives === "string") {
+            if (!["error", "warn", "off"].includes(linterOptions.reportUnusedDisableDirectives)) {
+                throw new Error("Invalid value for 'reportUnusedDisableDirectives' in config.");
+            }
```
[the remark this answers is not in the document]
bmish: Removed.

lib/config/flat-config-schema.js
bmish: Done.

bmish:
@nzakas @mdjermanovic finally got this working, feedback addressed, and the tests passing.

The goal is still to release this in a minor version of ESLint v8, pending a thorough review to ensure everything looks good. 

After that, I can open another PR to change the default for flat config users to `warn`, also suitable for a minor version of ESLint v8 since flat config is not yet considered stable.

lib/cli.js
```diff
@@ -125,6 +126,13 @@ async function translateOptions({
             rules: rule ? rule : {}
         }];
 
+        if (reportUnusedDisableDirectives || reportUnusedDisableDirectivesSeverity) {
+            if (!overrideConfig[0].linterOptions) {
+                overrideConfig[0].linterOptions = {};
+            }
+            overrideConfig[0].linterOptions.reportUnusedDisableDirectives = reportUnusedDisableDirectivesSeverity || (reportUnusedDisableDirectives ? "error" : void 0);
```
mdjermanovic, line 133:
```suggestion
            overrideConfig[0].linterOptions = {
                reportUnusedDisableDirectives: reportUnusedDisableDirectivesSeverity || "error"
            };
```

I think we can simplify this.

lib/options.js
```diff
@@ -306,6 +307,13 @@ module.exports = function(usingFlatConfig) {
                 default: void 0,
                 description: "Adds reported errors for unused eslint-disable and eslint-enable directives"
             },
+            {
+                option: "report-unused-disable-directives-severity",
+                type: "String",
+                default: void 0,
+                description: "Chooses severity level for reporting unused eslint-disable directives",
```
mdjermanovic, line 314:
```suggestion
                description: "Chooses severity level for reporting unused eslint-disable and eslint-enable directives",
```

Since the description for `report-unused-disable-directives` mentions `eslint-enable` directives, not mentioning them in the description for `report-unused-disable-directives-severity` might imply that this option does not apply to them.

docs/src/use/command-line-interface.md
```diff
@@ -98,6 +98,7 @@ Output:
 Inline configuration comments:
   --no-inline-config              Prevent comments from changing config or rules
   --report-unused-disable-directives  Adds reported errors for unused eslint-disable and eslint-enable directives
+  --report-unused-disable-directives-severity String  Choose what severity level that eslint-disable directives should be reported as
```
mdjermanovic, line 101:
```suggestion
  --report-unused-disable-directives-severity String  Chooses severity level for reporting unused eslint-disable and eslint-enable directives - either: off, warn, or error
```

This should be the same as in the output of `eslint --help`.

docs/src/use/command-line-interface.md
```diff
@@ -591,6 +592,16 @@ For example, suppose a rule has a bug that causes it to report a false positive,
 npx eslint --report-unused-disable-directives file.js
 ```
 
+#### `--report-unused-disable-directives-severity`
+
+Same as [`--report-unused-disable-directives`](#--report-unused-disable-directives), but allows you to specify the severity level (`error`, `warn`, `off`) of the reported errors. Only one of these two options can be used at a time.
```
mdjermanovic, line 597: We should specify Argument Type here, see for example [`--cache-strategy`](https://eslint.org/docs/latest/use/command-line-interface#--cache-strategy).

docs/src/use/configure/configuration-files-new.md
```diff
@@ -244,20 +244,22 @@ export default [
 
 #### Reporting unused disable directives
 
-Disable and enable directives such as `/*eslint-disable*/`, `/*eslint-enable*/` and `/*eslint-disable-next-line*/` are used to disable ESLint rules around certain portions of code. As code changes, it's possible for these directives to no longer be needed because the code has changed in such a way that the rule is no longer triggered. You can enable reporting of these unused disable directives by setting the `reportUnusedDisableDirectives` option to `true`, as in this example:
+Disable and enable directives such as `/*eslint-disable*/` and `/*eslint-disable-next-line*/` are used to disable ESLint rules around certain portions of code. As code changes, it's possible for these directives to no longer be needed because the code has changed in such a way that the rule is no longer triggered. You can enable reporting of these unused disable directives by setting the `reportUnusedDisableDirectives` option to a severity string, as in this example:
```
mdjermanovic, line 247: Why not keep ``/*eslint-enable*/`` here?

lib/eslint/flat-eslint.js
```diff
@@ -448,7 +447,7 @@ async function calculateConfigArray(eslint, {
  * @param {FlatConfigArray} config.configs The config.
  * @param {boolean} config.fix If `true` then it does fix.
  * @param {boolean} config.allowInlineConfig If `true` then it uses directive comments.
- * @param {boolean} config.reportUnusedDisableDirectives If `true` then it reports unused `eslint-disable` comments.
+ * @param {boolean|string} config.reportUnusedDisableDirectives If `true` or `error`, then it reports unused `eslint-disable` comments.
```
mdjermanovic, line 450: If we're removing the constructor option `reportUnusedDisableDirectives`, this parameter should be removed too since it was passed the value of the constructor option.

tests/lib/eslint/eslint.js
```diff
@@ -211,7 +211,6 @@ describe("ESLint", () => {
                     overrideConfig: "",
                     overrideConfigFile: "",
                     plugins: "",
-                    reportUnusedDisableDirectives: "",
```
mdjermanovic, line 214: This test can stay the same since the `ESLint` class doesn't have any changes.

lib/eslint/eslint-helpers.js
```diff
@@ -675,7 +675,6 @@ function processOptions({
     overrideConfig = null,
     overrideConfigFile = null,
     plugins = {},
-    reportUnusedDisableDirectives = null, // ← should be null by default because if it's a string then it overrides the 'reportUnusedDisableDirectives' setting in config files. And we cannot use `overrideConfig.reportUnusedDisableDirectives` instead because we cannot configure the `error` severity with that.
```
mdjermanovic, line 678: Since this is an API change when compared to previous APIs that had this option, it would be good to show a descriptive message if someone tries to pass this option. Like, for example, we're doing with the `globals` options:

https://github.com/eslint/eslint/blob/e0cb960e45961a36845b075a2986a1fa6cd6b8aa/lib/eslint/eslint-helpers.js#L702-L704

```js
if (unknownOptionKeys.includes("reportUnusedDisableDirectives")) {
    errors.push("'reportUnusedDisableDirectives' has been removed. Please use the 'overrideConfig.linterOptions.reportUnusedDisableDirectives' option instead.");
}
```

docs/src/use/command-line-interface.md
bmish: Added.

lib/eslint/flat-eslint.js
bmish: Fixed.

docs/src/use/configure/configuration-files-new.md
bmish: Added it back. This was a merge conflict since `eslint-enable` was added by another PR later.

mdjermanovic:
> * For simplicity, I've limited this to severity strings for now. We can add support for severity numbers now or later as needed.

If we're going to support severity numbers, I think it would be best to support them from the start (this PR).

I'm not sure if we'd like to make any further changes in the `ESLint` class, so its constructor option could keep accepting only strings. We could just translate numbers into strings in `cli.js`, in the code path that uses the `ESLint` class ([around here](https://github.com/bmish/eslint/blob/7c88af089bebb164202f7dad2f3b3200456c73e9/lib/cli.js#L199)).

As for the flat config array, it would make sense to normalize the value of `linterOptions.reportUnusedDisableDirectives` in calculated configs, either while merging configs or in [`finalizeConfig`](https://github.com/eslint/eslint/blob/e0cb960e45961a36845b075a2986a1fa6cd6b8aa/lib/config/flat-config-array.js#L184). Since rule severities are normalized to numbers, perhaps `linterOptions.reportUnusedDisableDirectives` should be normalized to a number as well?

packages/eslint-config-eslint/base.js
```diff
@@ -378,7 +378,7 @@ const eslintCommentsConfigs = [eslintComments.configs.recommended, {
 }];
 
 module.exports = [
-    { linterOptions: { reportUnusedDisableDirectives: true } },
+    { linterOptions: { reportUnusedDisableDirectives: "warn" } },
```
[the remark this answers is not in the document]
bmish: Good idea, fixed.

lib/config/flat-config-schema.js
```diff
@@ -262,6 +262,20 @@ const booleanSchema = {
...
 
+/** @type {ObjectPropertySchema} */
+const severitySchema = {
+    merge(first, second) {
+        const value = second === void 0 ? first : second;
+
+        return normalizeSeverity(value);
+    },
+    validate(value) {
+        if (!(["true", "false", "error", "warn", "off", 2, 1, 0].includes(value) || typeof value === "boolean")) {
```
mdjermanovic, line 273: String values `"true"` and `"false"` shouldn't be allowed.

bmish: Removed.

tests/lib/eslint/flat-eslint.js
```diff
@@ -3862,6 +3953,66 @@ describe("FlatESLint", () => {
                 assert.strictEqual(results[0].suppressedMessages.length, 0);
             });
 
+            it("should not warn unused 'eslint-disable' comments if 'reportUnusedDisableDirectives = false'.", async () => {
```
bmish, line 3956: Not sure what's going on but this test is failing in CI but passing locally for me.

lib/config/flat-config-schema.js
```diff
@@ -262,6 +263,24 @@ const booleanSchema = {
...
+/** @type {ObjectPropertySchema} */
+const severitySchema = {
+    merge(first, second) {
+        const value = second === void 0 ? first : second;
+
+        if (typeof value === "boolean") {
+            return value ? "warn" : "off";
+        }
+
+        return normalizeSeverityToString(value);
```
mdjermanovic, line 275: Thoughts about normalizing to number values here, for consistency with rule severities?

For example:

```js
// eslint.config.js
module.exports = [{
    linterOptions: {
        reportUnusedDisableDirectives: "error"
    },
    rules: {
        "no-undef": "error"
    }
}];
```

```
eslint --print-config foo.js
```

Currently:

```json
{
  "linterOptions": {
    "reportUnusedDisableDirectives": "error"
  },
  "languageOptions": {
    "ecmaVersion": "latest",
    "sourceType": "module",
    "parser": "espree@9.6.1",
    "parserOptions": {},
    "globals": {}
  },
  "plugins": [
    "@"
  ],
  "rules": {
    "no-undef": [
      2
    ]
  }
}
```

If we normalize to numbers:

```json
{
  "linterOptions": {
    "reportUnusedDisableDirectives": 2
  },
  "languageOptions": {
    "ecmaVersion": "latest",
    "sourceType": "module",
    "parser": "espree@9.6.1",
    "parserOptions": {},
    "globals": {}
  },
  "plugins": [
    "@"
  ],
  "rules": {
    "no-undef": [
      2
    ]
  }
}
```

bmish: Seems reasonable. I updated to normalize to a number here as it was trivial to do so.

As I commented before, I'm still curious to learn about the long-term plans for strings vs. numbers, if there are any plans around this.

tests/lib/cli.js
```diff
@@ -1380,6 +1380,84 @@ describe("cli", () => {
         });
     });
 
+    describe("--report-unused-disable-directives", () => {
```
mdjermanovic, line 1383: Since these options are available in both modes (eslintrc and flat config), can we run these tests in both modes? (for example, move this to line 1370 and use `useFlatConfig` variable instead of `true`).

lib/config/flat-config-schema.js
nzakas: There are no long-term plans. Strings have always just been aliases for numbers, and normalizing to numbers allows us to avoid string comparisons.

lib/eslint/flat-eslint.js
```diff
@@ -460,7 +458,6 @@ function verifyText({
     configs,
     fix,
     allowInlineConfig,
-    reportUnusedDisableDirectives,
```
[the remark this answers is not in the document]
bmish: Nice, fixed! I'm used to catching stuff like this with TypeScript...

tests/lib/eslint/flat-eslint.js
```diff
@@ -3888,7 +4043,7 @@ describe("FlatESLint", () => {
 
                 it("should warn unused 'eslint-disable' comments as error if 'reportUnusedDisableDirectives=error' was given in runtime.", async () => {
                     const teardown = createCustomTeardown({
-                        cwd: root,
+                        cwd: `${root}${i}`,
                         files: {
                             "test.js": "/* eslint-disable eqeqeq */",
                             ".eslintrc.yml": "reportUnusedDisableDirectives: true"
```
mdjermanovic, line 4049:
```suggestion
                            "eslint.config.js": "module.exports = [{ linterOptions: { reportUnusedDisableDirectives: true } }]"
```

Just to fix this test to work as intended.

tests/lib/eslint/flat-eslint.js
```diff
@@ -3862,10 +3955,70 @@ describe("FlatESLint", () => {
...
+            });
+
             describe("the runtime option overrides config files.", () => {
                 it("should not warn unused 'eslint-disable' comments if 'reportUnusedDisableDirectives=off' was given in runtime.", async () => {
                     const teardown = createCustomTeardown({
-                        cwd: root,
+                        cwd: `${root}${i}`,
                         files: {
                             "test.js": "/* eslint-disable eqeqeq */",
                             ".eslintrc.yml": "reportUnusedDisableDirectives: true"
```
mdjermanovic, line 4024:
```suggestion
                            "eslint.config.js": "module.exports = [{ linterOptions: { reportUnusedDisableDirectives: true } }]"
```

mdjermanovic (commented):
Some small suggestions, then LGTM.

docs/src/use/configure/configuration-files-new.md
```diff
@@ -244,20 +244,22 @@ export default [
...
+            reportUnusedDisableDirectives: "error"
         }
     }
 ];
 ```
 
-By default, unused disable and enable directives are reported as warnings. You can change this setting using the `--report-unused-disable-directives` command line option.
+You can override this setting using the [`--report-unused-disable-directives`](../command-line-interface#--report-unused-disable-directives) or the [`--report-unused-disable-directives-severity`](../command-line-interface#--report-unused-disable-directives-severity) command line options.
+
+For legacy compatibility, `true` is equivalent to `"warn"` and `false` is equivalent to `"off"`.
```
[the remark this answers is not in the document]
bmish: @mdjermanovic just making sure you saw this. I'm fine either way.

mdjermanovic (approved):
LGTM, thanks! Would like @nzakas to verify before merging.

docs/src/use/configure/configuration-files-new.md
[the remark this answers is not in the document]
mdjermanovic: Sorry for not replying earlier, I remember we discussed this in the RFC and decided to keep booleans.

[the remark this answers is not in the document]
mdjermanovic: Here: https://github.com/eslint/rfcs/pull/100#discussion_r1044770688

lib/shared/severity.js
```diff
@@ -0,0 +1,44 @@
+"use strict";
```
nzakas, line 1: Please add a file header comment describing the file.

lib/config/flat-config-schema.js
```diff
@@ -262,6 +263,24 @@ const booleanSchema = {
     validate: "boolean"
 };
 
+/** @type {ObjectPropertySchema} */
+const severitySchema = {
```
nzakas, line 267: This isn't strictly a severity schema as it supports boolean values, too. So maybe we rename it to `disableDirectiveSeveritySchema` for clarity?

lib/config/flat-config-schema.js
```diff
@@ -262,6 +263,24 @@ const booleanSchema = {
...
+
+        if (typeof value === "boolean") {
+            return value ? "warn" : "off";
+        }
+
+        return normalizeSeverityToNumber(value);
+    },
+    validate(value) {
+        if (!(["error", "warn", "off", 2, 1, 0].includes(value) || typeof value === "boolean")) {
+            throw new TypeError("Expected a severity string/number or boolean.");
```
nzakas, line 279: For clarity:

```suggestion
            throw new TypeError("Expected one of: \"error\", \"warn\", \"off\", 0, 1, 2, or a boolean.");
```

lib/config/flat-config-schema.js
```diff
@@ -262,6 +263,24 @@ const booleanSchema = {
...
+        const value = second === void 0 ? first : second;
+
+        if (typeof value === "boolean") {
+            return value ? "warn" : "off";
+        }
+
+        return normalizeSeverityToNumber(value);
+    },
+    validate(value) {
+        if (!(["error", "warn", "off", 2, 1, 0].includes(value) || typeof value === "boolean")) {
```
nzakas, line 278: Because this array's values never change, let's move it out of this function and change it to a set. Sets have much faster lookup than arrays.

nzakas (approved):
Overall LGTM. I left a few suggestions for clarity and clean up. Once those are addressed this is good to merge.

lib/shared/severity.js
bmish: Added.

lib/config/flat-config-schema.js
bmish: Updated.

lib/config/flat-config-schema.js
bmish: Good idea, updated.

mdjermanovic (approved):

nzakas (approved):
LGTM. Thanks!

PR #17315, opened 2023-06-27
test: More test cases for no-case-declarations

ElianCordoba:
#### Prerequisites checklist

[ ] Documentation update
[ ] Add autofix to a rule
[ ] Add a CLI option
[ ] Add something to the core
[x] Other, please explain:
	A (minor) perf improvement

Improved heuristic for the rule, if there are no `consequents` or if the `case` block contains a block statement, then we can determine that the rule does not need to be executed.

I benchmarked this with the `TIMING` env variable but sadly I didn't see any improvements, leaving this up to the maintainers if you want to merge it or not. I didn't know how to bench this in a larger codebase, maybe with `npm link`?

### DIFF
--- a/tests/lib/rules/no-case-declarations.js
+++ b/tests/lib/rules/no-case-declarations.js
@@ -35,9 +35,49 @@ ruleTester.run("no-case-declarations", rule, {
         {
             code: "switch (a) { case 1: { class C {} break; } default: { class C {} break; } }",
             parserOptions: { ecmaVersion: 6 }
+        },
+        {
+            code: `
+                switch (a) { 
+                    case 1: 
+                    case 2: {} 
+                }
+            `,
+            parserOptions: { ecmaVersion: 6 }
+        },
+        {
+            code: `
+                switch (a) {
+                    case 1: var x; 
+                }
+            `,
+            parserOptions: { ecmaVersion: 6 }
         }
     ],
     invalid: [
+        {
+            code: `
+                switch (a) { 
+                    case 1: 
+                        {} 
+                        function f() {} 
+                        break; 
+                }
+            `,
+            parserOptions: { ecmaVersion: 6 },
+            errors: [{ messageId: "unexpected", type: "FunctionDeclaration" }]
+        },
+        {
+            code: `
+                switch (a) { 
+                    case 1: 
+                    case 2: 
+                        let x; 
+                }
+            `,
+            parserOptions: { ecmaVersion: 6 },
+            errors: [{ messageId: "unexpected", type: "VariableDeclaration" }]
+        },
         {
             code: "switch (a) { case 1: let x = 1; break; }",
             parserOptions: { ecmaVersion: 6 },
--- a/tests/lib/rules/no-case-declarations.js
+++ b/tests/lib/rules/no-case-declarations.js
@@ -36,23 +36,17 @@ ruleTester.run("no-case-declarations", rule, {
             code: "switch (a) { case 1: { class C {} break; } default: { class C {} break; } }",
             parserOptions: { ecmaVersion: 6 }
         },
-        {
-            code: `
+        `
                 switch (a) { 
                     case 1: 
                     case 2: {} 
                 }
             `,
-            parserOptions: { ecmaVersion: 6 }
-        },
-        {
-            code: `
+        `
                 switch (a) {
                     case 1: var x; 
                 }
-            `,
-            parserOptions: { ecmaVersion: 6 }
-        }
+            `
     ],
     invalid: [
         {
@@ -64,7 +58,6 @@ ruleTester.run("no-case-declarations", rule, {
                         break; 
                 }
             `,
-            parserOptions: { ecmaVersion: 6 },
             errors: [{ messageId: "unexpected", type: "FunctionDeclaration" }]
         },
         {


fasttime:
Thanks for your interest in contributing to ESLint @ElianCordoba.

I think this change does not preserve the current behavior. For example, a code like this should report an error:

```js
/* eslint no-case-declarations: 2 */

switch (a) { case 1: {} function f() {} break; }
```

[**DEMO**](https://eslint.org/play/#eyJ0ZXh0IjoiLyogZXNsaW50IG5vLWNhc2UtZGVjbGFyYXRpb25zOiAyICovXG5cbnN3aXRjaCAoYSkgeyBjYXNlIDE6IHt9IGZ1bmN0aW9uIGYoKSB7fSBicmVhazsgfSIsIm9wdGlvbnMiOnsicGFyc2VyT3B0aW9ucyI6eyJlY21hVmVyc2lvbiI6ImxhdGVzdCIsInNvdXJjZVR5cGUiOiJzY3JpcHQiLCJlY21hRmVhdHVyZXMiOnt9fSwicnVsZXMiOnt9LCJlbnYiOnsiZXM2Ijp0cnVlfX19)

But it no longer does after this change. What kind of code are you expecting to be impacted by this optimization?

ElianCordoba:
Ah you are right, the current logic won't catch that example. So, this is the original case I was trying improve:

Screenshot 2023-06-28 at 17 39 39

If the `case` clause is empty or, if it has the brackes (having only one `consequents` being a Block statement), then we know that we can skip the check.

I updated my branch with a change that covers the example that @fasttime shared, I will also add it as a test case.

lib/rules/no-case-declarations.js
```diff
@@ -47,6 +47,13 @@ module.exports = {
 
         return {
             SwitchCase(node) {
+                if (
+                    node.consequent.length === 0 || // Empty clause
+                    node.consequent[0].type === "BlockStatement" && node.consequent.length === 1 // Clause has {} around child expressions
```
BYK, line 52: Better to check "the only element" part earlier
```suggestion
                    node.consequent.length === 1 && node.consequent[0].type === "BlockStatement"  // Clause has {} around child expressions
```

BYK (commented):
I appreciate the new test cases that said the scenario this is optimizing for does not seem worth optimizing to me:

1. If `node.consequent.length` is `0`, then the for loop will not be executed at all, so there should be no visible improvement for that case
2. If `node.consequent.length` is `1`, then the for loop will only be run once which does almost exactly the same thing you are doing in the if statement -- take the first element, compare one of its attributes with a string.

So my inclination is to reject this change as it does add more code and logic but the return on investment on that seems 0 or very close to 0.

What do you think?

mdjermanovic:
> 1. If `node.consequent.length` is `0`, then the for loop will not be executed at all, so there should be no visible improvement for that case
> 2. If `node.consequent.length` is `1`, then the for loop will only be run once which does almost exactly the same thing you are doing in the if statement -- take the first element, compare one of its attributes with a string.
> 
> So my inclination is to reject this change as it does add more code and logic but the return on investment on that seems 0 or very close to 0.

I agree with @BYK, it doesn't seem that the added code would really improve the performance.

Test cases look useful. @ElianCordoba if you want, you can undo the code changes and we'll merge the tests.

ElianCordoba:
Hi @BYK and @mdjermanovic!

I agree, it's not worth the extra complexity, I took the suggestion and undid the logic changes and left the new tests.

Thanks for the guidance!

BYK (approved):

tests/lib/rules/no-case-declarations.js
```diff
@@ -35,9 +35,49 @@ ruleTester.run("no-case-declarations", rule, {
...
             parserOptions: { ecmaVersion: 6 }
+        },
+        {
+            code: `
+                switch (a) { 
+                    case 1: 
+                    case 2: {} 
+                }
+            `,
+            parserOptions: { ecmaVersion: 6 }
```
fasttime, line 46: Just noting, in case it wasn't clear, that `parserOptions` is not required here, since no ES6 features are being used.

tests/lib/rules/no-case-declarations.js
```diff
@@ -35,9 +35,49 @@ ruleTester.run("no-case-declarations", rule, {
...
+        },
+        {
+            code: `
+                switch (a) {
+                    case 1: var x; 
+                }
+            `,
+            parserOptions: { ecmaVersion: 6 }
```
fasttime, line 54: Same as above, `parserOptions` is not required.

tests/lib/rules/no-case-declarations.js
```diff
@@ -35,9 +35,49 @@ ruleTester.run("no-case-declarations", rule, {
...
+        {
+            code: `
+                switch (a) { 
+                    case 1: 
+                        {} 
+                        function f() {} 
+                        break; 
+                }
+            `,
+            parserOptions: { ecmaVersion: 6 },
```
fasttime, line 67: Same as above, `parserOptions` is not required.

fasttime (approved):
LGTM. I just left some notes.

ElianCordoba:
Fixed @fasttime!

mdjermanovic (approved):
LGTM, thanks!

PR #17386, opened 2023-07-18
docs: Add private class features info to no-underscore-dangle

matwilko:
#### Prerequisites checklist

[X] Documentation update
[ ] Add autofix to a rule
[ ] Add a CLI option
[ ] Add something to the core
[ ] Other, please explain:

`no-underscore-dangle` docs currently state that 'JavaScript doesn’t have truly private members' and go into detail around the history of underscored identifiers.

Since ES2022/ES13 brought in truly private members, I've slimmed down the historical context and added info about private members and a recommendation to use them instead of underscored identifiers.

I removed the historical context because it seems less relevant now that actual private fields can be recommended, but I can restore it if you think it still provides useful context.

As a follow-up, would a proposal for a new rule that highlights leading-underscore members and suggests they be refactored as private members be reasonable? Or should that perhaps be rolled into this rule to prevent any overlap?

### DIFF
--- a/docs/src/rules/no-underscore-dangle.md
+++ b/docs/src/rules/no-underscore-dangle.md
@@ -3,16 +3,17 @@ title: no-underscore-dangle
 rule_type: suggestion
 ---
 
-
 As far as naming conventions for identifiers go, dangling underscores may be the most polarizing in JavaScript. Dangling underscores are underscores at either the beginning or end of an identifier, such as:
 
 ```js
 var _foo;
 ```
 
-There is actually a long history of using dangling underscores to indicate "private" members of objects in JavaScript (though JavaScript doesn't have truly private members, this convention served as a warning). This began with SpiderMonkey adding nonstandard methods such as `__defineGetter__()`. The intent with the underscores was to make it obvious that this method was special in some way. Since that time, using a single underscore prefix has become popular as a way to indicate "private" members of objects.
+There is a long history of marking "private" members with dangling underscores in JavaScript, beginning with SpiderMonkey adding nonstandard methods such as `__defineGetter__()`. Since that time, using a single underscore prefix has become the most popular convention for indicating a member is not part of the public interface of an object.
 
-Whether or not you choose to allow dangling underscores in identifiers is purely a convention and has no effect on performance, readability, or complexity. It's purely a preference.
+It is recommended to use the formal [private class features](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_class_fields) introduced in ECMAScript 2022 for encapsulating private data and methods rather than relying on naming conventions.
+
+Allowing dangling underscores in identifiers is purely a convention and has no effect on performance, readability, or complexity. They do not have the same encapsulation benefits as private class features, even with this rule enabled.
 
 ## Rule Details
 


docs/src/rules/no-underscore-dangle.md
```diff
@@ -10,9 +10,11 @@ As far as naming conventions for identifiers go, dangling underscores may be the
 var _foo;
 ```
 
-There is actually a long history of using dangling underscores to indicate "private" members of objects in JavaScript (though JavaScript doesn't have truly private members, this convention served as a warning). This began with SpiderMonkey adding nonstandard methods such as `__defineGetter__()`. The intent with the underscores was to make it obvious that this method was special in some way. Since that time, using a single underscore prefix has become popular as a way to indicate "private" members of objects.
```
nzakas, line 13: I'd like to keep this history -- there's no harm in providing it and I think it provides helpful context both for the convention and for the rule. We can probably just change the last sentence to read like this:

> Since that time, using a single underscore prefix became a popular way to indicate that an object member was to be considered "private" and not part of its public interface.

docs/src/rules/no-underscore-dangle.md
```diff
@@ -10,9 +10,11 @@ As far as naming conventions for identifiers go, dangling underscores may be the
...
+There is a long history of using dangling underscores to indicate "private" members of objects in JavaScript, with the intent to make it obvious that the property or method is special in some way.
 
-Whether or not you choose to allow dangling underscores in identifiers is purely a convention and has no effect on performance, readability, or complexity. It's purely a preference.
+ES2022/ES13 has since introduced [Private class features](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_class_fields), with broad ecosystem support. It is recommended you use those features where possible as they provide truly private members.
+
+Allowing dangling underscores in identifiers is purely a convention and has no effect on performance, readability, or complexity. However, they do not have the same encapsulation benefits as private class features, even with this rule enabled.
```
nzakas, line 17:
```suggestion
Official [private class features](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_class_fields) were introduced ECMAScript 2022. It is recommended you use those features where private fields are necessary.

Allowing dangling underscores in identifiers is purely a convention and has no effect on performance, readability, or complexity. They do not have the same encapsulation benefits as private class features, even with this rule enabled.
```

nzakas (changes requested):
Thanks for looking at this. I've left some notes to clean this up a bit but I like the direction.

matwilko:
Put the history back in, but tried to tighten it up a bit.

Put the Private class features notice in an `Important` box to really highlight it.

docs/src/rules/no-underscore-dangle.md
```diff
@@ -3,16 +3,19 @@ title: no-underscore-dangle
...
 ```js
 var _foo;
 ```
 
-There is actually a long history of using dangling underscores to indicate "private" members of objects in JavaScript (though JavaScript doesn't have truly private members, this convention served as a warning). This began with SpiderMonkey adding nonstandard methods such as `__defineGetter__()`. The intent with the underscores was to make it obvious that this method was special in some way. Since that time, using a single underscore prefix has become popular as a way to indicate "private" members of objects.
+There is a long history of marking "private" members with dangling underscores in JavaScript, beginning with SpiderMonkey adding nonstandard methods such as `__defineGetter__()`. Since that time, using a single underscore prefix has become the most popular convention for indicating a member is not part of the public interface of an object.
+
+::: important
+Official [Private class features](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_class_fields) were introduced in ECMAScript 2022. It is recommended you use those features when you need private members, rather than rely on conventions.
+:::
```
nzakas, line 16: We don't use asides like these in the intro material for rules, and a bit of cleanup.

```suggestion
Formal [private class fields](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_class_fields) were introduced in ECMAScript 2022. It is recommended you use private class fields when you need private members rather than relying on naming conventions.
```

nzakas: > Sure, noted for the future. I can't find any particular style guide for the docs anywhere, is this sort of thing just institutional memory?

Yeah, we don't have a style guide for docs. We tend to save asides for when they are really pointing out something that is important not to miss. In this case, it's just part of the normal flow of the text so there's really not a need for any extra callout.

> I was trying to keep consistent with the title on MDN

Ah, fair! I thought because the URL had "private class fields" that the title would, too.

> This feels a bit redundant 😄. I think They are recommended when you need private members instead of relying on naming conventions. covers being brief and hits the same points 😄

Yeah, it is a bit redundant, but "they" is unclear. For technical documentation, we always lean towards being redundant in order to improve clarity. (Words like "they", "it", etc., are best replaced with the actual things they are describing.)

I've updated the suggestion above.

matwilko: *tries not to bikeshed this too much*

If we rearrange slightly and combine into one sentence, I think that eliminates any ambiguity and gives us the brevity we're aiming at:

> We recommend using the formal [private class features](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_class_fields) introduced in ECMAScript 2022 for encapsulating private data and methods rather than relying on naming conventions.

nzakas: Works for me. Just make sure "recommend" -> "recommended"

docs/src/rules/no-underscore-dangle.md
```diff
@@ -3,16 +3,17 @@ title: no-underscore-dangle
...
 
 ```js
 var _foo;
 ```
 
-There is actually a long history of using dangling underscores to indicate "private" members of objects in JavaScript (though JavaScript doesn't have truly private members, this convention served as a warning). This began with SpiderMonkey adding nonstandard methods such as `__defineGetter__()`. The intent with the underscores was to make it obvious that this method was special in some way. Since that time, using a single underscore prefix has become popular as a way to indicate "private" members of objects.
+There is a long history of marking "private" members with dangling underscores in JavaScript, beginning with SpiderMonkey adding nonstandard methods such as `__defineGetter__()`. Since that time, using a single underscore prefix has become the most popular convention for indicating a member is not part of the public interface of an object.
+
+It is recommend to use the formal [private class features](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_class_fields) introduced in ECMAScript 2022 for encapsulating private data and methods rather than relying on naming conventions.
```
nzakas, line 14:
```suggestion
It is recommended to use the formal [private class features](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_class_fields) introduced in ECMAScript 2022 for encapsulating private data and methods rather than relying on naming conventions.
```

matwilko: 🤦 This is what happens when I make commits before I've woken up properly 😆

nzakas: We've all been there. :)

PR #17471, opened 2023-08-16
fix: no-lone-blocks wont flag blocks followed by break in switch-case

srijan-paul:
…that are followed by a break

#### Prerequisites checklist

Bug fix ([template](https://raw.githubusercontent.com/eslint/eslint/HEAD/templates/bug-report.md))

* **Node version:** v18.14.0
* **npm version:** v9.3.1
* **Local ESLint version:** v8.47.0
* **Global ESLint version:** Not found
* **Operating System:**  darwin 22.5.0

`Default (Espree)`

<details>

```js
{
  rules: { "no-lone-blocks": "error" }
}
```

</details>

Wrote a switch case like so:

```js
switch (foo) {
  case "bar":
    {
      const one = 1;
      console.log(one);
      // ... 
    }
    break;
    default: break;
}
```

The rule `no-lone-blocks` should **not** have been reported on the `BlockStatement` inside the switch case.

Yes, it's possible to move the `break` inside the block and leave the behavior untouched, but I think that the linter should identify and ignore this particular case since it's a common pattern.

The `no-lone-blocks` lint was raised here:

Screenshot 2023-08-16 at 2 28 30 PM

Add a check in the rule that confirms if the block-statement inside a switch-case is followed by a single break statement, and nothing else

Nothing in particular, since it's a small change.
Though, I'd like to be wary of any edge cases that I might have left out.

### DIFF
--- a/lib/rules/no-lone-blocks.js
+++ b/lib/rules/no-lone-blocks.js
@@ -62,7 +62,10 @@ module.exports = {
                 node.parent.type === "Program" ||
 
                 // Don't report blocks in switch cases if the block is the only statement of the case.
-                node.parent.type === "SwitchCase" && !(node.parent.consequent[0] === node && node.parent.consequent.length === 1);
+                node.parent.type === "SwitchCase" &&
+                !(node.parent.consequent[0] === node && node.parent.consequent.length === 1) &&
+                !(node.parent.consequent.length === 2 && node.parent.consequent[0] === node &&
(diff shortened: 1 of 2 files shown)


Rec0iL99:
@srijan-paul there's a review comment that you need to look into.

snitin315:
No response, hence closing. Feel free to open an issue with more info.

PR #17500, opened 2023-08-25
docs: add typescript template

Zamiell:
#### Prerequisites checklist

[X] Documentation update

This is a follow-up on my last PR, #17423.

@nzakas mentioned in the ESLint Discord server that we should move forward with reducing the duplication introduced with my last PR by using a frontmatter template, which this PR includes.

Additionally, I also documented the remaining known ESLint core rules that conflict with the TypeScript compiler.

#### Additional Discussion

When doing this PR, I noticed that the rules pages have inconsistent formatting.
Rules randomly have either 1, 2, 3, or 5 newlines after the frontmatter.
This kind of thing makes it difficult for people to contribute, as it makes finding the proper style non-trivial.
I propose that docs should be formatted using Prettier (or dprint), in the same way that the @typescript-eslint project does.

### DIFF
--- a/docs/src/_includes/layouts/doc.html
+++ b/docs/src/_includes/layouts/doc.html
@@ -27,7 +27,7 @@ layout: base.html
             </p>
             {% if extra_typescript_info %}
                 <p>
-                    {{ extra_typescript_info | markdown }}
+                    {{ extra_typescript_info | markdown | safe }}
                 </p>
             {% endif %}
         {% endset %}
--- a/docs/src/_includes/layouts/doc.html
+++ b/docs/src/_includes/layouts/doc.html
@@ -27,7 +27,7 @@ layout: base.html
             </p>
             {% if extra_typescript_info %}
                 <p>
-                    {{ extra_typescript_info }}
+                    {{ extra_typescript_info | markdown }}
                 </p>
             {% endif %}
         {% endset %}
--- a/docs/src/_includes/layouts/doc.html
+++ b/docs/src/_includes/layouts/doc.html
@@ -22,11 +22,14 @@ layout: base.html
     {% if handled_by_typescript %}
         {% set handled_by_typescript_content %}
             <h2 id="handled_by_typescript">Handled by TypeScript</h2>
-            It is safe to disable this rule when using TypeScript because TypeScript's compiler enforces this check (
-            {%- for code in handled_by_typescript -%}
-                <code>ts({{ code }})</code>{% if not loop.last %} & {% endif %}
-            {%- endfor -%}
-            ).
+            <p>
+                It is safe to disable this rule when using TypeScript because TypeScript's compiler enforces this check.
+            </p>
+            {% if extra_typescript_info %}
+                <p>
+                    {{ extra_typescript_info }}
+                </p>
+            {% endif %}
         {% endset %}
 
         {% set all_content = [all_content, handled_by_typescript_content] | join %}
--- a/docs/src/_includes/layouts/doc.html
+++ b/docs/src/_includes/layouts/doc.html
@@ -19,6 +19,19 @@ layout: base.html
     {% set added_version = rule_versions.added[title] %}
     {% set removed_version = rule_versions.removed[title] %}
 
+    {% if handled_by_typescript %}
+        {% set handled_by_typescript_content %}
+            <h2 id="handled_by_typescript">Handled by TypeScript</h2>
+            It is safe to disable this rule when using TypeScript because TypeScript's compiler enforces this check (
+            {%- for code in handled_by_typescript -%}
+                <code>ts({{ code }})</code>{% if not loop.last %} & {% endif %}
+            {%- endfor -%}
+            ).
+        {% endset %}
+
+        {% set all_content = [all_content, handled_by_typescript_content] | join %}
+    {% endif %}
+
     {% if related_rules %}
         {% set related_rules_content %}
             <h2 id="related-rules">Related Rules</h2>
@@ -48,7 +61,7 @@ layout: base.html
 
         {% set all_content = [all_content, further_reading_content] | join %}
     {% endif %}
-        
+
     {% if rule_meta %}
         {% set resources_content %}
             <h2 id="resources">Resources</h2>
@@ -76,7 +89,7 @@ layout: base.html
                 {% endif %}
 
                 {% include 'components/docs-toc.html' %}
-                
+
                 {{ all_content | safe }}
             </div>
 
@@ -102,6 +115,7 @@ layout: base.html
             {% include "partials/docs-footer.html" %}
         </div>
     </div>
+
     <a id="scroll-up-btn" href="#site_top">
         <svg fill="none" height="24" stroke="currentColor" stroke-linecap="round" stroke-linejoin="round" stroke-width="2" viewBox="0 0 24 24" width="24"><line x1="12" x2="12" y1="19" y2="5"/><polyline points="5 12 12 5 19 12"/></svg>
     </a>
--- a/docs/src/rules/constructor-super.md
+++ b/docs/src/rules/constructor-super.md
@@ -1,9 +1,7 @@
 ---
 title: constructor-super
 rule_type: problem
-handled_by_typescript:
-- 2335
-- 2377
+handled_by_typescript: true
 ---
 
 Constructors of derived classes must call `super()`.
--- a/docs/src/rules/constructor-super.md
+++ b/docs/src/rules/constructor-super.md
@@ -1,7 +1,12 @@
 ---
 title: constructor-super
 rule_type: problem
+handled_by_typescript:
+- 2335
+- 2377
 ---
+
+
 
 Constructors of derived classes must call `super()`.
 Constructors of non derived classes must not call `super()`.
@@ -69,5 +74,3 @@ class A extends B {
 ## When Not To Use It
 
 If you don't want to be notified about invalid/missing `super()` callings in constructors, you can safely disable this rule.
-
-It is safe to disable this rule when using TypeScript because TypeScript's compiler enforces this check (`ts(2335) & ts(2377)`).
--- a/docs/src/rules/constructor-super.md
+++ b/docs/src/rules/constructor-super.md
@@ -6,8 +6,6 @@ handled_by_typescript:
 - 2377
 ---
 
-
-
 Constructors of derived classes must call `super()`.
 Constructors of non derived classes must not call `super()`.
 If this is not observed, the JavaScript engine will raise a runtime error.
--- a/docs/src/rules/no-func-assign.md
+++ b/docs/src/rules/no-func-assign.md
@@ -1,8 +1,7 @@
 ---
 title: no-func-assign
 rule_type: problem
-handled_by_typescript:
-- 2630
+handled_by_typescript: true
 ---
 
 
--- a/docs/src/rules/no-func-assign.md
+++ b/docs/src/rules/no-func-assign.md
@@ -2,7 +2,7 @@
 title: no-func-assign
 rule_type: problem
 handled_by_typescript:
-- 2539
+- 2630
 ---
 
 
--- a/docs/src/rules/no-func-assign.md
+++ b/docs/src/rules/no-func-assign.md
@@ -1,6 +1,8 @@
 ---
 title: no-func-assign
 rule_type: problem
+handled_by_typescript:
+- 2539
 ---
 
 
--- a/docs/src/rules/no-import-assign.md
+++ b/docs/src/rules/no-import-assign.md
@@ -1,6 +1,8 @@
 ---
 title: no-import-assign
 rule_type: problem
+handled_by_typescript: true
+extra_typescript_info: However, note that the compiler will not catch the "Object.assign" case. Thus, if you use "Object.assign" in your codebase, this rule will still provide some miniscule value.
 ---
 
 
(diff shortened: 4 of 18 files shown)


AndreaPontrandolfo:
@Zamiell 

Are you sure about `no-redeclare`?
From the docs: 
> In JavaScript, it’s possible to redeclare the same variable name using var.

Typescript seem to warn about re-declared `let` and `const` variables, but not `var` variables, which is precisely the usecase of this rule.

Also for `no-import-assign` Typescript doesn't seem to warn about this case:

`Object.assign(mod_ns, { foo: "foo" })`

while the ESLint rule does.

Zamiell:
@AndreaPontrandolfo Thank you. I have verified your findings, so I will update the PR.

To clarify, I was using this as a reference: https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/src/configs/eslint-recommended.ts

I will also bring this up with the `typeScript-eslint` team.

Zamiell:
The PR is updated now. (I also [opened an issue at the `typescript-eslint` repository](https://github.com/typescript-eslint/typescript-eslint/issues/7547).)

nzakas (approved):
Nice! This LGTM. Would like someone to verify the compiler warnings for TypeScript.

(@JoshuaKGoldberg would you be able to do that?)

AndreaPontrandolfo:
> @AndreaPontrandolfo See [Brad's reply here](https://github.com/typescript-eslint/typescript-eslint/issues/7547#issuecomment-1694774593). Given this, I think it does make sense to update the language in this PR as well.

I saw it. I'm not 100% onboard with his reasoning on that, but i understand it. 
Your re-wording sounds good to me, it informs the user of the difference between the TS check and ESLint rule, makes sense to me 👍

AndreaPontrandolfo:
> @AndreaPontrandolfo See [Brad's reply here](https://github.com/typescript-eslint/typescript-eslint/issues/7547#issuecomment-1694774593). Given this, I think it does make sense to update the language in this PR as well.

I saw it. I'm not 100% onboard with his reasoning on that, but i understand it. 
Your re-wording sounds good to me, it informs the user of the difference between the TS check and ESLint rule, makes sense to me 👍

docs/src/rules/constructor-super.md
```diff
@@ -1,6 +1,9 @@
 ---
 title: constructor-super
 rule_type: problem
+handled_by_typescript:
+- 2335
+- 2377
```
JoshuaKGoldberg, line 6: Just noting that these version numbers aren't guaranteed to exist version-to-version: https://github.com/microsoft/TypeScript/issues/19139[^1]

[^1]: I think there's an issue floating around somewhere in the TypeScript issue tracker that more explicitly states the errors can change version-to-version... but I couldn't find it after 5 minutes. 🤷

docs/src/rules/no-func-assign.md
```diff
@@ -1,6 +1,8 @@
 ---
 title: no-func-assign
 rule_type: problem
+handled_by_typescript:
+- 2539
```
JoshuaKGoldberg, line 5: I get `2630` in TS 5.2.2:

```
Cannot assign to 'foo' because it is a function.(2630)
```

https://www.typescriptlang.org/play?#code/GYVwdgxgLglg9mABAIwIYCcAUBKRBvAXwCgjg45EBeFDAbhNElgUTLh32NPKpvXqA

`2539` is the more general _"not a variable"_ error:
https://github.com/microsoft/TypeScript/blob/cd391b066b49cead0d1b1ec089ab4f4de59a16d1/src/compiler/diagnosticMessages.json#L2563-L2566

docs/src/rules/no-import-assign.md
```diff
@@ -57,3 +57,5 @@ test(mod_ns) // Not errored because it doesn't know that 'test' updates the memb
 ## When Not To Use It
 
 If you don't want to be notified about modifying imported bindings, you can disable this rule.
+
+If you use TypeScript, it is not recommended to use this rule, since the compiler will mostly catch this error (`ts(2539)` & `ts(2540)`). However, note that the compiler does not catch the `Object.assign` case, so this rule still provides some miniscule value.
```
JoshuaKGoldberg, line 61: I get `2632` here: https://www.typescriptlang.org/play?#code/JYWwDg9gTgLgBFApgQwMYwCKIGbIK4A2MANHAN5wDCE4EAdonfAL5zZQ1wBESaMXAKFCRYcAFRxkAZwQp0AOWQhEUsGkRsOIbr3SCBuzDnxE4AXjgBGAdVoMm5uACYDcmIuWr1AOluR78BYAzK58HipqqBoWZMwCAgDyAEYAVojo3tJSwADmdAAUhuFeUaQU2BAQAFzcFRBccMwAlAJAA

docs/src/rules/no-this-before-super.md
```diff
@@ -1,6 +1,8 @@
 ---
 title: no-this-before-super
 rule_type: problem
+handled_by_typescript:
+- 2376
```
JoshuaKGoldberg, line 5: `2376` is _`A 'super' call must be the first statement in the constructor to refer to 'super' or 'this' when a derived class contains initialized properties, parameter properties, or private identifiers.`_. But there's also `17009`: _`'super' must be called before accessing 'this' in the constructor of a derived class.`_

docs/src/rules/no-undef.md
```diff
@@ -1,6 +1,8 @@
 ---
 title: no-undef
 rule_type: problem
+handled_by_typescript:
+- 2304
```
JoshuaKGoldberg, line 5: `2552`: _`Cannot find name 'someFunction'. Did you mean 'Function'?`_

https://www.typescriptlang.org/play?#code/G4QwTgBAZg9jEF4IGcYFsCmAxArgOwGMAXASxjwAoBKAbgChRIAjcRCECAaggEYag

docs/src/rules/no-unsafe-negation.md
```diff
@@ -1,6 +1,10 @@
 ---
 title: no-unsafe-negation
 rule_type: problem
+handled_by_typescript:
+- 2358
+- 2360
+- 2365
```
JoshuaKGoldberg, line 7: https://www.typescriptlang.org/play?#code/CYUwxgNghgTiAEEQBd4GsQE8Bc8DOyMAlgHYDmA3AFCiSwJKoD2ARgFa4DeAvlVXXjzwAwsiYx4PPkQBm8ABQBCDJnil4rNuGQBKSVXiH4AemMaADiBhQxE83DAhQJR-AC2UDEKKoQARwBXIgA3KCQSVDEFZSw9dU1tAyNTeCgSYHhkTEt4MCYSYKs8Inz3TxBvX0CQsJAIzKYFFXgAfngAIhkwvBB2+Fx2wgDeuJINdkTeKllozTUSAjTHJjlRcT1OJMMUpktrW3h7cCc61w8vNSqg0PDIxqVNUcWXEBWRWy2TMzSMn3h-a61epRMIAdygmCEXQgPXwpFcLCYTCQaXgN2GQno8BITGYEzAyDwADoqLwgA

* `2358` ✔️ 
* `2322`: _`Type 'boolean' is not assignable to type 'string | number | symbol'.`_

I don't see a `2360` in TypeScript's `diagnosticMessages.json`?

JoshuaKGoldberg (changes requested):
Layout changes look good to me! I do wonder if it'd be possible to automate it _just_ a bit more so that no `.md` file includes the explicit _"it's mostly useless, but for this one edge case"_ text? Maybe not worth it.

A few of the error messages don't match what I get in the TS playground for 5.2.2. Requesting changes on that?

docs/src/_includes/layouts/doc.html
JoshuaKGoldberg, line 1: I wonder if there's a build step that could be added as a followup which, for each file with `handled_by_typescript_content`:

1. Runs the code samples in the latest stable version of TypeScript
2. Checks that the output complaints include all the mentioned error codes

docs/src/rules/no-func-assign.md
Zamiell: ok, i'll update these, but can you update the upstream source too? e.g. the comments here: https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/src/configs/eslint-recommended.ts

JoshuaKGoldberg: https://github.com/typescript-eslint/typescript-eslint/issues/7582

docs/src/_includes/layouts/doc.html
Zamiell: documenting the actual error codes does not provide enough value to be worth it, so i will just remove all of the code numbers from this PR

docs/src/rules/no-invalid-this.md
```diff
@@ -1,6 +1,8 @@
 ---
 title: no-invalid-this
 rule_type: suggestion
+handled_by_typescript: true
+extra_typescript_info: Note that technically, TypeScript will only catch this if you have the "strict" or "noImplicitThis" flags enabled. These are enabled in most TypeScript projects, since they are considered to be best practice.
```
nzakas, line 5:
```suggestion
extra_typescript_info: Note that, technically, TypeScript will only catch this if you have the `strict` or `noImplicitThis` flags enabled. These are enabled in most TypeScript projects, since they are considered to be best practice.
```

docs/src/rules/no-import-assign.md
```diff
@@ -1,6 +1,8 @@
 ---
 title: no-import-assign
 rule_type: problem
+handled_by_typescript: true
+extra_typescript_info: However, note that the compiler will not catch the "Object.assign" case. Thus, if you use "Object.assign" in your codebase, this rule will still provide some miniscule value.
```
nzakas, line 5:
```suggestion
extra_typescript_info: Note that the compiler will not catch the `Object.assign()` case. Thus, if you use `Object.assign()` in your codebase, this rule will still provide some value.
```

docs/src/rules/no-redeclare.md
```diff
@@ -1,6 +1,8 @@
 ---
 title: no-redeclare
 rule_type: suggestion
+handled_by_typescript: true
+extra_typescript_info: However, note that while TypeScript will catch `let` redeclares and `const` redeclares, it will not catch `var` redeclares. Thus, if you use the legacy `var` keyword in your TypeScript codebase, this rule will still provide some miniscule value.
```
nzakas, line 5:
```suggestion
extra_typescript_info: Note that while TypeScript will catch `let` redeclares and `const` redeclares, it will not catch `var` redeclares. Thus, if you use the legacy `var` keyword in your TypeScript codebase, this rule will still provide some value.
```

docs/src/_includes/layouts/doc.html
```diff
@@ -19,6 +19,22 @@
...
 
+    {% if handled_by_typescript %}
+        {% set handled_by_typescript_content %}
+            <h2 id="handled_by_typescript">Handled by TypeScript</h2>
+            <p>
+                It is safe to disable this rule when using TypeScript because TypeScript's compiler enforces this check.
+            </p>
+            {% if extra_typescript_info %}
+                <p>
+                    {{ extra_typescript_info }}
```
nzakas, line 30: This will not interpret the content as Markdown, which means we will see the backticks. I believe this is the correct filter to use.
```suggestion
                    {{ extra_typescript_info | markdown }}
```

nzakas (changes requested):
Overall this looks good. Cleaned up some of the language and left a note about Markdown processing.

docs/src/rules/no-invalid-this.md
Zamiell: I deliberately used quotes instead of backticks here because they don't propagate properly through to the markdown template - it results in the literal backtick instead of code. I also tried using `<code>`, which results in the same thing, a literal `<code>`. Thus, I recommend using quotes for now, unless you know a specific way around that.

docs/src/rules/no-import-assign.md
Zamiell: can't use backticks here for the same reason that i put in my previous comment

docs/src/rules/no-redeclare.md
Zamiell: can't use backticks here for the same reason that i put in my previous comment

Zamiell:
I commited all of your suggestions, but now it looks like this:

image

Not sure what the correct solution is here, I'm new to Eleventy.

nzakas:
Okay, we're getting closer. I think if you add `| safe` after `| markdown` that should work. So `| markdown | safe` altogether.

Zamiell:
That worked:

image

I committed the fix now.

docs/src/_includes/layouts/doc.html
JoshuaKGoldberg: Even better!

nzakas (approved):
LGTM. Would like @JoshuaKGoldberg to verify his changes before merging.

JoshuaKGoldberg (approved):
💯 fantastic!

docs/src/rules/no-invalid-this.md
```diff
@@ -1,6 +1,8 @@
 ---
 title: no-invalid-this
 rule_type: suggestion
+handled_by_typescript: true
+extra_typescript_info: Note that, technically, TypeScript will only catch this if you have the `strict` or `noImplicitThis` flags enabled. These are enabled in most TypeScript projects, since they are considered to be best practice.
```
snitin315, line 5:
```suggestion
extra_typescript_info: Note that, technically, TypeScript will only catch this if you have the `strict` or `noImplicitThis` flags enabled. These are enabled in most TypeScript projects since they are considered to be best practices.
```

snitin315 (commented):
One minor suggestion & then LGTM.

docs/src/rules/no-invalid-this.md
Zamiell: its written as non-plural because strict is a superset containing noimplicitthis, so strict is best practice and noimplicitthis is not

PR #17548, opened 2023-09-08
feat: Stdout outputting for stdin

dev1437:
#### Prerequisites checklist

[ ] Documentation update
[ ] Add autofix to a rule
[X] Add a CLI option
[ ] Add something to the core
[ ] Other, please explain:

Add --stdout option to cli and allow fix to be set if outputting

Does the change make sense, does it need tests?

### DIFF
--- a/lib/cli.js
+++ b/lib/cli.js
@@ -368,12 +368,17 @@ const cli = {
             log.error("The --fix option and the --fix-dry-run option cannot be used together.");
             return 2;
         }
-        if (useStdin && options.fix) {
+        if (useStdin && options.fix && !options.stdout) {
             log.error("The --fix option is not available for piped-in code; use --fix-dry-run instead.");
             return 2;
         }
         if (options.fixType && !options.fix && !options.fixDryRun) {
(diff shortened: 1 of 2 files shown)


nzakas:
@dev1437 thanks for the suggestion. It's not clear from the diff what exactly it is that you're trying to accomplish with this change. Can you explain what problem you're trying to solve?

And yes, all code changes require tests, and any changes to the CLI also require documentation updates.

dev1437:
@nzakas I'm trying to allow code fixes when using stdin as an input. At the moment it just forces dry-run only.

My main problem is using dprint with Vue. I would like to use Vue's eslint plugin to format the template while using dprint to format the script tag. To do this I can set dprint to pass the template into eslint via stdin (which then expects the formatted code to be outputted to stdout); however, eslint currently does not allow fixed code to be put into stdout. 

After further research, I've decided to use a custom formatter that works with dry-run:

```js
module.exports = function(results, context) {
  return results.map((r) => r.output && r.output.trim()).join("\0");
};
```

Maybe this will help others that come across my issue, or maybe it was just me being silly 🙂

PR #17567, opened 2023-09-15
chore: add .github/renovate.json5

JoshuaKGoldberg:
#### Prerequisites checklist

[ ] Documentation update
[ ] Add autofix to a rule
[ ] Add a CLI option
[ ] Add something to the core
[x] Other, please explain: GitHub automation

Creates a `.github/renovate.json5` file based on:

* Discussion in https://github.com/eslint/eslint/issues/15726
* Docs in https://docs.renovatebot.com/configuration-options/#packagerules
* The existing typescript-eslint file at https://github.com/typescript-eslint/typescript-eslint/blob/7826910d8c6ba4d2d40b8d55fcf7fd39fe290d88/.github/renovate.json5

* If & when this config creates update PRs, we'll be able to see which packages are blocked from upgrading. Those can be added to an `ignoreDeps` field. Are there any we already know should be?
* I removed any configuration around semantic commit titles in the hope that it'll auto-detect based on the Git history. If it doesn't, I suspect we'd want to report that as a bug to Renovate.
* I set it to create majors, create and auto-merge minors, and ignore patches to reduce. An alternative could be to create and auto-merge patches to be more up-to-date. Thoughts? 
* Are there any other package groups that I missed?

### DIFF
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -52,7 +52,7 @@
         }
     ],
 
-    // By default renovate will auto-rebase whenever the dep branch falls behind main.
+    // By default Renovate will auto-rebase whenever the dep branch falls behind main.
     // This is annoying as it spams notifications and creates unnecessary action runs.
     // Instead only auto-rebase when conflicted, and we can trigger a manual rebase if required.
     "rebaseWhen": "conflicted"
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -50,10 +50,5 @@
             "groupName": "metascraper",
             "matchPackagePrefixes": ["metascraper"]
         }
-    ],
-
-    // By default Renovate will auto-rebase whenever the dep branch falls behind main.
-    // This is annoying as it spams notifications and creates unnecessary action runs.
-    // Instead only auto-rebase when conflicted, and we can trigger a manual rebase if required.
-    "rebaseWhen": "conflicted"
+    ]
 }
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -15,7 +15,7 @@
         // Enable major version updates, but don't auto-merge them
         {
             "automerge": false,
-            "matchUpdateTypes": ["minor"]
+            "matchUpdateTypes": ["major"]
         },
 
         // Enable minor version updates with auto-merging
@@ -32,11 +32,11 @@
 
         // Label PRs appropriately per their manager
         {
-            "addLabels": ["javascript"],
+            "addLabels": ["deps:npm"],
             "matchManagers": ["npm"]
         },
         {
-            "addLabels": ["github-actions"],
+            "addLabels": ["deps:actions"],
             "matchManagers": ["github-actions"]
         },
 
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -1,0 +1,66 @@
+{
+    "$schema": "https://docs.renovatebot.com/renovate-schema.json",
+    "extends": [
+        "config:recommended",
+        ":dependencyDashboardApproval",
+        ":semanticCommitTypeAll(ci)"
+    ],
+    "dependencyDashboard": true,
+    "dependencyDashboardLabels": ["dependencies", "repo maintenance"],
+    "enabledManagers": ["github-actions", "npm"],
+    "internalChecksFilter": "strict",
+    "labels": ["dependencies"],
+
+    // Wait a few days for any new package, as a precaution against malicious publishes
+    "minimumReleaseAge": "3 days",
+
+    "packageRules": [
+        // Enable major version updates, but don't auto-merge them
+        {
+            "automerge": false,
+            "matchUpdateTypes": ["minor"]
+        },
+
+        // Enable minor version updates with auto-merging
+        {
+            "automerge": true,
+            "automergeStrategy": "squash",
+            "matchUpdateTypes": ["minor"]
+        },
+
+        // Ignore patch versions to reduce noise
+        {
+            "enabled": false,
+            "matchUpdateTypes": ["patch"]
+        },
+
+        // Label PRs appropriately per their manager
+        {
+            "addLabels": ["javascript"],
+            "matchManagers": ["npm"]
+        },
+        {
+            "addLabels": ["github-actions"],
+            "matchManagers": ["github-actions"]
+        },
+
+        // Batch package sets together
+        {
+            "groupName": "babel",
+            "matchPackagePrefixes": ["@babel", "babel-"]
+        },
+        {
+            "groupName": "wdio",
+            "matchPackagePrefixes": ["@wdio"]
+        },
+        {
+            "groupName": "metascraper",
+            "matchPackagePrefixes": ["metascraper"]
+        }
+    ],
+
+    // By default renovate will auto-rebase whenever the dep branch falls behind main.
+    // This is annoying as it spams notifications and creates unnecessary action runs.
+    // Instead only auto-rebase when conflicted, and we can trigger a manual rebase if required.
+    "rebaseWhen": "conflicted"
+}
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -1,5 +1,9 @@
 {
     "$schema": "https://docs.renovatebot.com/renovate-schema.json",
+    "extends": [
+        "config:recommended",
+        ":dependencyDashboardApproval"
+    ],
     "internalChecksFilter": "strict",
     "labels": ["dependencies"],
 
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -2,7 +2,7 @@
     "$schema": "https://docs.renovatebot.com/renovate-schema.json",
     "extends": [
         "config:recommended",
-        ":dependencyDashboardApproval",
+        ":approveMajorUpdates",
         ":semanticCommitScopeDisabled"
     ],
     "internalChecksFilter": "strict",
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -12,12 +12,8 @@
     "minimumReleaseAge": "7 days",
 
     "packageRules": [
-        // Enable major version updates, but don't auto-merge them
-        {
-            "matchUpdateTypes": ["major"]
-        },
-
         // Enable minor version updates with auto-merging
+        // (Major versions will need to be triggered from the dashboard issue)
         {
             "automerge": true,
             "matchUpdateTypes": ["minor"]
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -1,11 +1,5 @@
 {
     "$schema": "https://docs.renovatebot.com/renovate-schema.json",
-    "extends": [
-        "config:recommended",
-        ":dependencyDashboardApproval"
-    ],
-    "dependencyDashboard": true,
-    "dependencyDashboardLabels": ["dependencies", "repo maintenance"],
     "enabledManagers": ["github-actions", "npm"],
     "internalChecksFilter": "strict",
     "labels": ["dependencies"],
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -2,8 +2,7 @@
     "$schema": "https://docs.renovatebot.com/renovate-schema.json",
     "extends": [
         "config:recommended",
-        ":dependencyDashboardApproval",
-        ":semanticCommitTypeAll(ci)"
+        ":dependencyDashboardApproval"
     ],
     "dependencyDashboard": true,
     "dependencyDashboardLabels": ["dependencies", "repo maintenance"],
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -16,7 +16,6 @@
         // Enable minor version updates with auto-merging
         {
             "automerge": true,
-            "automergeStrategy": "squash",
             "matchUpdateTypes": ["minor"]
         },
 
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -2,7 +2,8 @@
     "$schema": "https://docs.renovatebot.com/renovate-schema.json",
     "extends": [
         "config:recommended",
-        ":dependencyDashboardApproval"
+        ":dependencyDashboardApproval",
+        ":semanticCommitScopeDisabled"
     ],
     "internalChecksFilter": "strict",
     "labels": ["dependencies"],
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -9,7 +9,7 @@
     "labels": ["dependencies"],
 
     // Wait a few days for any new package, as a precaution against malicious publishes
-    "minimumReleaseAge": "3 days",
+    "minimumReleaseAge": "7 days",
 
     "packageRules": [
         // Enable major version updates, but don't auto-merge them
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -1,6 +1,5 @@
 {
     "$schema": "https://docs.renovatebot.com/renovate-schema.json",
-    "enabledManagers": ["github-actions", "npm"],
     "internalChecksFilter": "strict",
     "labels": ["dependencies"],
 
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -14,7 +14,6 @@
     "packageRules": [
         // Enable major version updates, but don't auto-merge them
         {
-            "automerge": false,
             "matchUpdateTypes": ["major"]
         },
 
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -5,32 +5,23 @@
         ":approveMajorUpdates",
         ":semanticCommitScopeDisabled"
     ],
+    "ignorePresets": [":semanticPrefixFixDepsChoreOthers"],
     "internalChecksFilter": "strict",
     "labels": ["dependencies"],
 
-    // Wait a few days for any new package, as a precaution against malicious publishes
+    // Wait well over three days for any new package as a precaution against malicious publishes
     "minimumReleaseAge": "7 days",
 
     "packageRules": [
         {
-            "description": "Automerge minor version updates. Major versions must be approved from the Dependency Dashboard issue, because we use the :approveMajorUpdates preset.",
-            "automerge": true,
-            "matchUpdateTypes": ["minor"]
-        },
-        {
-            "description": "Ignore patch versions to reduce noise.",
-            "enabled": false,
-            "matchUpdateTypes": ["patch"]
+            "description": "Use the deps:actions label for github-action manager updates (this means Renovate's github-action manager).",
+            "addLabels": ["deps:actions"],
+            "matchManagers": ["github-actions"]
         },
         {
             "description": "Use the deps:npm label for npm manager packages (this means Renovate's npm manager).",
             "addLabels": ["deps:npm"],
             "matchManagers": ["npm"]
-        },
-        {
-            "description": "Use the deps:actions label for github-action manager updates (this means Renovate's github-action manager).",
-            "addLabels": ["deps:actions"],
-            "matchManagers": ["github-actions"]
         },
         {
             "description": "Group Babel packages into a single PR.",
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -6,7 +6,6 @@
         ":semanticCommitScopeDisabled"
     ],
     "ignorePresets": [":semanticPrefixFixDepsChoreOthers"],
-    "internalChecksFilter": "strict",
     "labels": ["dependencies"],
 
     // Wait well over three days for any new package as a precaution against malicious publishes
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -12,39 +12,38 @@
     "minimumReleaseAge": "7 days",
 
     "packageRules": [
-        // Enable minor version updates with auto-merging
-        // (Major versions will need to be triggered from the dashboard issue)
         {
+            "description": "Automerge minor version updates. Major versions must be approved from the Dependency Dashboard issue, because we use the :approveMajorUpdates preset.",
             "automerge": true,
             "matchUpdateTypes": ["minor"]
         },
-
-        // Ignore patch versions to reduce noise
         {
+            "description": "Ignore patch versions to reduce noise.",
             "enabled": false,
             "matchUpdateTypes": ["patch"]
         },
-
-        // Label PRs appropriately per their manager
         {
+            "description": "Use the deps:npm label for npm manager packages (this means Renovate's npm manager).",
             "addLabels": ["deps:npm"],
             "matchManagers": ["npm"]
         },
         {
+            "description": "Use the deps:actions label for github-action manager updates (this means Renovate's github-action manager).",
             "addLabels": ["deps:actions"],
             "matchManagers": ["github-actions"]
         },
-
-        // Batch package sets together
         {
+            "description": "Group Babel packages into a single PR.",
             "groupName": "babel",
             "matchPackagePrefixes": ["@babel", "babel-"]
         },
         {
+            "description": "Group wdio packages into a single PR.",
             "groupName": "wdio",
             "matchPackagePrefixes": ["@wdio"]
         },
         {
+            "description": "Group metascraper packages into a single PR.",
             "groupName": "metascraper",
             "matchPackagePrefixes": ["metascraper"]
         }
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -8,7 +8,8 @@
     "ignorePresets": [":semanticPrefixFixDepsChoreOthers"],
     "labels": ["dependencies"],
 
-    // Wait well over three days for any new package as a precaution against malicious publishes
+    // Wait well over npm's three day window for any new package as a precaution against malicious publishes
+    // https://docs.npmjs.com/policies/unpublish/#packages-published-less-than-72-hours-ago
     "minimumReleaseAge": "7 days",
 
     "packageRules": [


.github/renovate.json5
```diff
@@ -0,0 +1,66 @@
+{
+    "$schema": "https://docs.renovatebot.com/renovate-schema.json",
+    "extends": [
+        "config:recommended",
+        ":dependencyDashboardApproval",
+        ":semanticCommitTypeAll(ci)"
+    ],
```
JoshuaKGoldberg, line 7: @HonkingGoose I don't recall ever using `extends` before. The docs at https://docs.renovatebot.com/presets-config/#configrecommended don't say what the explicit values are. Is that intentional? If not, would a good next step be for me to file a docs issue?

.github/renovate.json5
```diff
@@ -0,0 +1,65 @@
+{
+    "$schema": "https://docs.renovatebot.com/renovate-schema.json",
+    "extends": [
+        "config:recommended",
+        ":dependencyDashboardApproval"
+    ],
```
JoshuaKGoldberg, line 6: @HonkingGoose I don't recall ever using `extends` before. The docs at https://docs.renovatebot.com/presets-config/#configrecommended don't say what the explicit values are. Is that intentional? If not, would a good next step be for me to file a docs issue?

.github/renovate.json5
```diff
@@ -0,0 +1,65 @@
...
+    "dependencyDashboard": true,
+    "dependencyDashboardLabels": ["dependencies", "repo maintenance"],
+    "enabledManagers": ["github-actions", "npm"],
+    "internalChecksFilter": "strict",
+    "labels": ["dependencies"],
+
+    // Wait a few days for any new package, as a precaution against malicious publishes
+    "minimumReleaseAge": "3 days",
```
JoshuaKGoldberg, line 14: @HonkingGoose curious your thoughts on this. It's an opinion I've been spouting online without explicit validation from anyone in Dependabot/Renovate/etc. 😄

.github/renovate.json5
rarkins: Hi @JoshuaKGoldberg. Renovate has supported `extends` for 5+ years (inspired by `eslint` terminology). The page you linked to lists which presets it includes:

image

All the presets listed there are also documented at docs.renovatebot.com and once you drill down you can see the explicit values.

Can you explain where your confusion lies? Maybe I'm not understanding what you mean by "explicit values".

.github/renovate.json5
rarkins: In the Renovate org itself we use 7 days as default: https://github.com/renovatebot/.github/blob/a9b025a4302b34da7b219887630cb3bf50393ab8/default.json#L89-L95

Combined with dependency dashboard - which lets you push forward any update on demand - I think it makes sense.

.github/renovate.json5
HonkingGoose: > I don't recall ever using `extends` before.

Really? It almost sounds like you usually build your Renovate config from the ground up... 🙃 I sure hope I'm reading this wrong, or you'll have spent a lot of time on basic things...

By the way, we do have docs that try to explain the `extends` config option. [^1]

> All the presets listed there are also documented at docs.renovatebot.com and once you drill down you can see the explicit values.

I think they're trying to say that it's hard to drill down. To drill down the `config:recommended` preset, you'll have to have 7 tabs open for each `extends` item, and then try to keep it all in your head.

> Can you explain where your confusion lies? Maybe I'm not understanding what you mean by "explicit values".

It sounds like they want to see all the fully expanded configuration in a single JSON codeblock. That's probably why they want to file a docs issue.

[^1]: [Renovate docs, `extends` config option](https://docs.renovatebot.com/configuration-options/#extends)

.github/renovate.json5
```diff
@@ -0,0 +1,65 @@
+{
+    "$schema": "https://docs.renovatebot.com/renovate-schema.json",
+    "extends": [
+        "config:recommended",
+        ":dependencyDashboardApproval"
+    ],
+    "dependencyDashboard": true,
```
HonkingGoose, line 7:
```suggestion
```

You can drop this, if you want, because the `config:recommended` preset already enables the Dependency Dashboard.

.github/renovate.json5
HonkingGoose: When you wait a few days, the community may discover that the package is bad. By the time you would get the Renovate PR - when `minimumReleaseAge` is reached - the bad package _may_ already have been pulled by the registry maintainers. In that perfect scenario you wouldn't even get the PR for the bad update, and instead stay on the known-good version, or get a non-malicious update after the real maintainers get back control. I don't recommend relying on this (alone) to keep you safe.

The Renovate docs mention another use for setting the `minimumReleaseAge`. [^1]

> There are a couple of uses for `minimumReleaseAge`:
>
> ...snip...
>
> #### Prevent holding broken npm packages
> 
> npm packages less than 72 hours (3 days) old can be unpublished, which could result in a service impact if you have already updated to it. Set `minimumReleaseAge` to `3 days` for npm packages to prevent relying on a package that can be removed from the registry:
> 
> ```json
> {
>   "packageRules": [
>     {
>       "matchDatasources": ["npm"],
>       "minimumReleaseAge": "3 days"
>     }
>   ]
> }
> ```

So you're right that setting `minimumReleaseAge` to `3` is "kinda-sorta" recommended by Renovate. 😄 

[^1]: [Renovate docs for `minimumReleaseAge`](https://docs.renovatebot.com/configuration-options/#minimumreleaseage)

.github/renovate.json5
```diff
@@ -0,0 +1,65 @@
...
+            "groupName": "wdio",
+            "matchPackagePrefixes": ["@wdio"]
+        },
+        {
+            "groupName": "metascraper",
+            "matchPackagePrefixes": ["metascraper"]
+        }
+    ],
+
+    // By default renovate will auto-rebase whenever the dep branch falls behind main.
```
HonkingGoose, line 61:
```suggestion
    // By default Renovate will auto-rebase whenever the dep branch falls behind main.
```

You should capitalize proper nouns like Renovate. 🙃

.github/renovate.json5
```diff
@@ -0,0 +1,65 @@
...
+    // This is annoying as it spams notifications and creates unnecessary action runs.
+    // Instead only auto-rebase when conflicted, and we can trigger a manual rebase if required.
+    "rebaseWhen": "conflicted"
```
HonkingGoose, line 64: The current config will automerge `minor` updates:

> ```
>         // Enable minor version updates with auto-merging
>         {
>             "automerge": true,
>             "automergeStrategy": "squash",
>             "matchUpdateTypes": ["minor"]
>         },
> ```

Setting `rebaseWhen=conflicted` may cause problems when automerging! I'll quote the full `rebaseWhen` docs: [^1]

> ## rebaseWhen
> 
> Possible values and meanings:
> 
> - `auto`: Renovate will autodetect the best setting. It will use `behind-base-branch` if configured to automerge or repository has been set to require PRs to be up to date. Otherwise, `conflicted` will be used instead
> - `never`: Renovate will never rebase the branch or update it unless manually requested
> - `conflicted`: Renovate will rebase only if the branch is conflicted
> - `behind-base-branch`: Renovate will rebase whenever the branch falls 1 or more commit behind its base branch
> 
> `rebaseWhen=conflicted` is not recommended if you have enabled Renovate automerge, because:
> 
> - It could result in a broken base branch if two updates are merged one after another without testing the new versions together
> - If you have enforced that PRs must be up-to-date before merging (e.g. using branch protection on GitHub), then automerge won't be possible as soon as a PR gets out-of-date but remains non-conflicted
> 
> It is also recommended to avoid `rebaseWhen=never` as it can result in conflicted branches with outdated PR descriptions and/or status checks.
> 
> Avoid setting `rebaseWhen=never` and then also setting `prCreation=not-pending` as this can prevent creation of PRs.

Did you read the [Noise reduction docs](https://docs.renovatebot.com/noise-reduction/)? That has ways to reduce noise, I'll summarize them here:

- Package grouping
- Scheduling when Renovate runs
- Automerging
- Lock file considerations

[^1]: [Renovate docs for `rebaseWhen`](https://docs.renovatebot.com/configuration-options/#rebasewhen)

.github/renovate.json5
```diff
@@ -0,0 +1,65 @@
...
+            "automerge": false,
+            "matchUpdateTypes": ["minor"]
+        },
+
+        // Enable minor version updates with auto-merging
+        {
+            "automerge": true,
+            "automergeStrategy": "squash",
+            "matchUpdateTypes": ["minor"]
+        },
```
HonkingGoose, line 28: Automerging and `rebaseWhen=conflicted` can cause issues. Read my comment below.

.github/renovate.json5
```diff
@@ -0,0 +1,65 @@
+{
+    "$schema": "https://docs.renovatebot.com/renovate-schema.json",
+    "extends": [
+        "config:recommended",
+        ":dependencyDashboardApproval"
+    ],
+    "dependencyDashboard": true,
+    "dependencyDashboardLabels": ["dependencies", "repo maintenance"],
+    "enabledManagers": ["github-actions", "npm"],
```
HonkingGoose, line 9: Why are you only allowing the `github-actions` and `npm` managers? Usually I would recommend that you allow Renovate to update all the dependencies that it can find.

.github/renovate.json5
```diff
@@ -0,0 +1,65 @@
+{
+    "$schema": "https://docs.renovatebot.com/renovate-schema.json",
+    "extends": [
+        "config:recommended",
```
HonkingGoose, line 4: > I removed any configuration around semantic commit titles in the hope that it'll auto-detect based on the Git history. If it doesn't, I suspect we'd want to report that as a bug to Renovate.

Read the [Semantic Commit messages](https://docs.renovatebot.com/semantic-commits/) docs to learn more. But I'll quote the relevant part here:

> # Semantic Commit messages
> 
> Renovate looks at the last 10 commit messages in the base branch to decide if the repository uses semantic commits.
> If there are semantic commits, Renovate uses the [conventional-commits-detector](https://github.com/conventional-changelog/conventional-commits-detector) to decide what convention the commit messages follow.
> 
> Renovate only finds Angular-style conventional commits, it ignores other commit conventions.
> 
> When Renovate finds Angular-style commits, Renovate creates commit messages and PR titles like this:
> 
> - chore(deps): update eslint to v7.30.0
> 
> By default, Renovate uses the `chore` prefix.
> 
> If you extend from `config:recommended` then Renovate:
> 
> - still defaults to the `chore` prefix
> - uses the `fix` prefix for npm production dependencies
> - uses the `chore` prefix for npm development dependencies (`devDependencies`)

You're extending from `config:recommended` so, you'll get the behavior in the last part of the quote. 😉

.github/renovate.json5
```diff
@@ -0,0 +1,65 @@
+{
```
HonkingGoose, line 1: You may want to look into using GitHub's Merge Queue. I see you're not using it...

The main Renovate repository uses the Merge Queue, and it's made merging so much easier and faster. I don't need to update my branches when `main` changes under me after approval. The GitHub Merge Queue will auto-update branches, without making any more noise.

Read the [GitHub Docs, How merge queues work](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/configuring-pull-request-merges/managing-a-merge-queue#how-merge-queues-work) to learn the specifics.

Read about [Renovate support for GitHub's Merge Queue](https://docs.renovatebot.com/key-concepts/automerge/#github-merge-queue).

You'll need to ask somebody else for help to get GitHub Actions configured correctly for the Merge Queue. 😄

.github/renovate.json5
JoshuaKGoldberg: Cool, thanks, I wasn't sure that this was connected up to how we extended the config. I'll remove it as ESLint actually uses `chore: ...`, not `chore(deps): ...`.

Note that I didn't set the systems up and don't have ownership over it - this is just a proposal.

.github/renovate.json5
JoshuaKGoldberg: > All the presets listed there are also documented at docs.renovatebot.com and once you drill down you can see the explicit values.

> It sounds like they want to see all the fully expanded configuration in a single JSON codeblock. That's probably why they want to file a docs issue.

I'm not confused about anything 😄 @rarkins is right. I'm just being unclear about what I'm trying to say. Sorry. Filed https://github.com/renovatebot/renovate/discussions/24448.

The part that's relevant to this PR is what benefit this specific repo gets from using `config:recommended`. `eslint/eslint` establishes a more fine-grained control of its automations compared to most end-user repositories. Other than `group:babelMonorepo` / `monorepo:babel` I don't see anything relevant to this repo?

.github/renovate.json5
JoshuaKGoldberg: 👍 thanks, removing.

.github/renovate.json5
JoshuaKGoldberg: My context coming in was to enable only the bits desired. Since the repo mainly just those, I figured it made sense to just enable for them. But I can just remove it - makes sense. Thanks!

JoshuaKGoldberg:
Thanks for hopping on this HonkingGoose & rarkins! Really appreciate the advice 🙂

.github/renovate.json5
JoshuaKGoldberg: Actually, wait, re-reading: are you saying that it will always be `chore(deps)`, not `chore`? ESLint uses `chore`, not `chore(deps)`. So if it's always prefix(scope) and not just prefix then we'd want to manually override here.

.github/renovate.json5
```diff
@@ -0,0 +1,53 @@
...
+    "$schema": "https://docs.renovatebot.com/renovate-schema.json",
```
JoshuaKGoldberg, line 2: Note that I removed references to a dependency dashboard. I don't remember personally experiencing benefit from it, and have always just gone with issues. Removed from this initial proposal to reduce noise & impact of PR. But it's not a hard opinion of mine. 🙂

.github/renovate.json5
rarkins: I think you actually benefit from most of them, for example migrations and workarounds are presets which get updated regularly by the community and there's no reason to think this repo would benefit any less than others.

You certain are welcome/encouraged to break that preset into separate presets though, and exclude any which aren't applicable

HonkingGoose: Yeah, you would probably benefit from `config:recommended`. 😉 

You may use [`ignorePresets`](https://docs.renovatebot.com/configuration-options/#ignorepresets) to ignore parts of `config:recommended` you don't want. For example:

```json
{
  "extends": ["config:recommended"],
  "ignorePresets": ["config:thanks-i-hate-it"]
}
```

.github/renovate.json5
```diff
@@ -0,0 +1,53 @@
...
+        // Enable major version updates, but don't auto-merge them
+        {
+            "automerge": false,
+            "matchUpdateTypes": ["minor"]
+        },
+
+        // Enable minor version updates with auto-merging
+        {
+            "automerge": true,
+            "automergeStrategy": "squash",
```
viceice, line 19: not yet supported on GitHub .

@rarkins we really should add support for it to GitHub platform 🙃

.github/renovate.json5
HonkingGoose: ## How I understand Renovate's docs for Commit Message

The way I'm reading our docs: if you don't use any presets, then Renovate uses `chore`. But if you use `config:recommended` then you still get `chore` for most things, with one exception:

- `fix` for npm production deps
- keep `chore` for npm development deps

## Why `config:recommended` appears to do the above behavior

`config:recommended` extends the `":semanticPrefixFixDepsChoreOthers",` preset which expands to this:

```json
{
  "packageRules": [
    {
      "matchPackagePatterns": [
        "*"
      ],
      "semanticCommitType": "chore"
    },
    {
      "matchDepTypes": [
        "dependencies",
        "require"
      ],
      "semanticCommitType": "fix"
    },
    {
      "matchDatasources": [
        "maven"
      ],
      "matchDepTypes": [
        "compile",
        "provided",
        "runtime",
        "system",
        "import",
        "parent"
      ],
      "semanticCommitType": "fix"
    }
  ]
}
```

## Error in the Renovate docs?

I'm getting confused here... [Renovate's Semantic Commit docs](https://docs.renovatebot.com/semantic-commits/#semantic-commit-messages) say:

> By default, Renovate uses the `chore` prefix.
> 
> If you extend from `config:recommended` then Renovate:
> 
> - still defaults to the `chore` prefix
> - uses the `fix` prefix for npm production dependencies
> - uses the `chore` prefix for npm development dependencies (`devDependencies`)

But the ``":semanticPrefixFixDepsChoreOthers"` preset seems to do something different. IT's a bit tricky for me to read.

### Question for Renovate maintainer

@rarkins should I change the Semantic Commit docs page, and if yes, how?

## Disabling the Semantic Commit scope (the `(chore)` part)

Quote from the [Renovate docs, Changing the Semantic Commit scope](https://docs.renovatebot.com/semantic-commits/#changing-the-semantic-commit-scope):

> ## Changing the Semantic Commit scope
> 
> ... snip ...
> 
> To _remove_ the semantic commit scope, so Renovate uses `chore:` instead of `chore(deps):`, add the `":semanticCommitScopeDisabled"` preset to your `extends` array:
> 
> ```json
> {
>   "extends": [":semanticCommitScopeDisabled"]
> }
> ```
>

.github/renovate.json5
HonkingGoose: I found this old and closed issue:

- https://github.com/renovatebot/renovate/issues/16814

I think the original posters problem was solved by blocking rebase merges in the GitHub repository settings?

Read the [GitHub Docs, Configuring commit squashing for pull requests](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/configuring-pull-request-merges/configuring-commit-squashing-for-pull-requests) to learn where to find this setting. 😉

.github/renovate.json5
HonkingGoose: I love the Dependency Dashboard! ❤️ Here's why:

- You may use the [Dependency Dashboard approval workflow](https://docs.renovatebot.com/key-concepts/dashboard/#dependency-dashboard-approval-workflow) to approve certain updates before you get them. So you _first_ select the checkbox in the Issue, then you get the Renovate PR. Now you only get "noise" when you want to, for those updates.
- You can see any updates that you've ignored, for whatever reason, so you don't forget about them.

I usually pin the Dependency Dashboard issue to the top of the Issues list. That way I always know where to find it, no matter what repository I'm in.

.github/renovate.json5
viceice: this option is only supported on bitbucket cloud and gitea

https://docs.renovatebot.com/configuration-options/#automergestrategy

.github/renovate.json5
JoshuaKGoldberg: Thanks - I'll add it back in.

.github/renovate.json5
JoshuaKGoldberg: Added `":semanticCommitScopeDisabled"` to `extends`, thanks!

.github/renovate.json5
HonkingGoose: I don't want to hold up this PR by waiting for GitHub's Merge Queue to be enabled on this repository. So I'm opening a discussion instead:

- #17582

Can you please resolve this comment? 😉

.github/renovate.json5
```diff
@@ -0,0 +1,57 @@
...
+    "labels": ["dependencies"],
+
+    // Wait a few days for any new package, as a precaution against malicious publishes
+    "minimumReleaseAge": "3 days",
+
+    "packageRules": [
+        // Enable major version updates, but don't auto-merge them
+        {
+            "automerge": false,
+            "matchUpdateTypes": ["minor"]
```
nzakas, line 18: Should this be "major"?
```suggestion
            "matchUpdateTypes": ["major"]
```

.github/renovate.json5
```diff
@@ -0,0 +1,57 @@
...
+
+        // Ignore patch versions to reduce noise
+        {
+            "enabled": false,
+            "matchUpdateTypes": ["patch"]
+        },
+
+        // Label PRs appropriately per their manager
+        {
+            "addLabels": ["javascript"],
```
nzakas, line 35: "javascript" only makse sense in the context of the PRs, so maybe something more specific is better?
```suggestion
            "addLabels": ["deps:npm"],
```

.github/renovate.json5
```diff
@@ -0,0 +1,57 @@
...
+    "packageRules": [
+        // Enable major version updates, but don't auto-merge them
+        {
+            "automerge": false,
+            "matchUpdateTypes": ["minor"]
+        },
+
+        // Enable minor version updates with auto-merging
+        {
+            "automerge": true,
```
nzakas, line 23: How does automerge work with protected branches? Is it exempt?

.github/renovate.json5
```diff
@@ -0,0 +1,57 @@
...
+            "matchUpdateTypes": ["patch"]
+        },
+
+        // Label PRs appropriately per their manager
+        {
+            "addLabels": ["javascript"],
+            "matchManagers": ["npm"]
+        },
+        {
+            "addLabels": ["github-actions"],
```
nzakas, line 39: Similarly.

```suggestion
            "addLabels": ["deps:actions"],
```

.github/renovate.json5
nzakas: I'd also prefer 7 days. 3 days is tight if something is released on a Friday. :)

nzakas (commented):
Thanks for taking a look at this! I left a few suggestions and comments.

If you could please resolve any conversations that are complete, it will help the rest of us more easily review the changes. :)

.github/renovate.json5
rarkins: No, bots don't get any exemptions and need to follow the branch protections like any other user. So e.g. if a PR requires reviews, there needs to be a PR review. If you restrict who can merge to main branch and the bot isn't on the list, it won't be able to, etc.

nzakas: Okay, so for automerge to work we'd need to explicitly allow the bot permission to merge, otherwise the PR would remain open?

viceice: yes

.github/renovate.json5
```diff
@@ -0,0 +1,57 @@
...
+
+    // Wait a few days for any new package, as a precaution against malicious publishes
+    "minimumReleaseAge": "7 days",
+
+    "packageRules": [
+        // Enable major version updates, but don't auto-merge them
+        {
+            "automerge": false,
+            "matchUpdateTypes": ["major"]
+        },
```
viceice, line 19: can be removed, automerge is disabled by default

.github/renovate.json5
```diff
@@ -0,0 +1,56 @@
...
+    "labels": ["dependencies"],
+
+    // Wait a few days for any new package, as a precaution against malicious publishes
+    "minimumReleaseAge": "7 days",
+
+    "packageRules": [
+        // Enable major version updates, but don't auto-merge them
+        {
+            "matchUpdateTypes": ["major"]
+        },
```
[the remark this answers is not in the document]
JoshuaKGoldberg: Sorry, not sure how to parse the last bit - are you saying that if we don't include this line, ...

* Option A: major version PRs will still be sent? And that if Renovate is taking too long, we can manually trigger them?
* Option B: because `:dependencyDashboardApproval` is enabled, we'll have to manually trigger PRs from major versions from the dashboard?

https://docs.renovatebot.com/configuration-options/#dependencydashboardapproval makes me think it's the latter (B). I personally have prefer the former (A) but have gotten the impression in this PR that you all have recommended B.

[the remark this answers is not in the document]
rarkins: He's saying that a package rule with only `match` config does nothing. Think of a package rule as like "when you match X, do Y". You don't have any "do Y" in your rule, so it does nothing.

Your config sets dependencyDashboardApproval=true which means _no_ PR is created by default, and instead PRs are only created when someone with repo permissions checks a box in the Dependency Dashboard issue. In other words it's like "on demand" PRs instead of fully automated. It's not just major versions either, unless you write further rules to exclude non-major.

My recommendation in most cases is to use dependency dashboard approval for major updates, but allow non-major updates to be created automatically.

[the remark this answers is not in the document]
JoshuaKGoldberg: Swell, thanks - I'll remove this line.

.github/renovate.json5
```diff
@@ -0,0 +1,52 @@
+{
+    "$schema": "https://docs.renovatebot.com/renovate-schema.json",
+    "extends": [
+        "config:recommended",
+        ":dependencyDashboardApproval",
```
HonkingGoose, line 5: `:dependencyDashboardApproval` [^1] means you'll have to approve _all_ PRs from the Dependency Dashboard. Unless you set excemptions in `packageRules` below... but that is extra work. 😄 

I think you actually want `:approveMajorUpdates` [^2], where you only approve `major` updates beforehand, and get all other updates automatically?

```suggestion
        ":approveMajorUpdates",
```

[^1]: [Renovate docs, `:approveMajorUpdates` preset](https://docs.renovatebot.com/presets-default/#approvemajorupdates)
[^2]: [Renovate docs, `:dependencyDashboardApproval`](https://docs.renovatebot.com/presets-default/#dependencydashboardapproval)

.github/renovate.json5
HonkingGoose: Read the [Renovate docs, key concepts page for automerge](https://docs.renovatebot.com/key-concepts/automerge/) to learn more.

.github/renovate.json5
```diff
@@ -0,0 +1,52 @@
...
+    ],
+    "internalChecksFilter": "strict",
+    "labels": ["dependencies"],
+
+    // Wait a few days for any new package, as a precaution against malicious publishes
+    "minimumReleaseAge": "7 days",
+
+    "packageRules": [
+        // Enable minor version updates with auto-merging
+        // (Major versions will need to be triggered from the dashboard issue)
```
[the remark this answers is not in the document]
JoshuaKGoldberg: Out of curiosity, what's the benefit of having this string over a comment?

viceice (approved):

.github/renovate.json5
```diff
@@ -0,0 +1,52 @@
...
+        {
+            "groupName": "wdio",
+            "matchPackagePrefixes": ["@wdio"]
+        },
+        {
+            "groupName": "metascraper",
+            "matchPackagePrefixes": ["metascraper"]
+        }
+    ]
+}
```
HonkingGoose, line 52: This suggestion adds a `description` field to all the `packageRules` objects:

```suggestion
        {
            "description": "Automerge minor version updates. Major versions must be approved from the Dependency Dashboard issue, because we use the :approveMajorUpdates preset.",
            "automerge": true,
            "matchUpdateTypes": ["minor"]
        },
        {
            "description": "Ignore patch versions to reduce noise.",
            "enabled": false,
            "matchUpdateTypes": ["patch"]
        },
        {
            "description": "Use the deps:npm label for npm manager packages (this means Renovate's npm manager).",
            "addLabels": ["deps:npm"],
            "matchManagers": ["npm"]
        },
        {
            "description": "Use the deps:actions label for github-action manager updates (this means Renovate's github-action manager).",
            "addLabels": ["deps:actions"],
            "matchManagers": ["github-actions"]
        },
        {
            "description": "Group Babel packages into a single PR.",
            "groupName": "babel",
            "matchPackagePrefixes": ["@babel", "babel-"]
        },
        {
            "description": "Group wdio packages into a single PR.",
            "groupName": "wdio",
            "matchPackagePrefixes": ["@wdio"]
        },
        {
            "description": "Group metascraper packages into a single PR.",
            "groupName": "metascraper",
            "matchPackagePrefixes": ["metascraper"]
        }
    ]
}
```

.github/renovate.json5
JoshuaKGoldberg: Just confirming - is there any code change that needs to be done in this PR for this thread?

.github/renovate.json5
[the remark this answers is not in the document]
JoshuaKGoldberg: Thanks! Makes sense 🙂. Resolved by accepting https://github.com/eslint/eslint/pull/17567#discussion_r1342616420.

JoshuaKGoldberg:
Sorry for the delay! I believe there's nothing left waiting on me to address.

aladdin-add (changes requested):
we need to ignore all esm-only deps - they are non-upgradable as eslint is still cjs. otherwise LGTM! 👍

aladdin-add:
@HonkingGoose Thanks for the explanation, it sounds great! 👍

aladdin-add (approved):

HonkingGoose (approved):
I'm happy with this Renovate configuration.

nzakas:
@mdjermanovic would like you to take a look before merging.

.github/renovate.json5
```diff
@@ -0,0 +1,51 @@
...
+        {
+            "description": "Automerge minor version updates. Major versions must be approved from the Dependency Dashboard issue, because we use the :approveMajorUpdates preset.",
+            "automerge": true,
+            "matchUpdateTypes": ["minor"]
+        },
+        {
+            "description": "Ignore patch versions to reduce noise.",
+            "enabled": false,
+            "matchUpdateTypes": ["patch"]
+        },
```
mdjermanovic, line 24: Is ignoring patch versions a common practice? We mostly use major ranges (`^`) and I guess we want automatic in-range dependency updates mostly to make sure that bug fixes are included. From that perspective, making minor updates but ignoring patch updates doesn't make sense.

.github/renovate.json5
```diff
@@ -0,0 +1,51 @@
...
+    "internalChecksFilter": "strict",
+    "labels": ["dependencies"],
+
+    // Wait a few days for any new package, as a precaution against malicious publishes
+    "minimumReleaseAge": "7 days",
+
+    "packageRules": [
+        {
+            "description": "Automerge minor version updates. Major versions must be approved from the Dependency Dashboard issue, because we use the :approveMajorUpdates preset.",
+            "automerge": true,
```
mdjermanovic, line 17: Does automerge work only for in-range updates? In particular, we sometimes need to pin dependencies (for example, https://github.com/eslint/eslint/pull/16897) and in that case further updates shouldn't be automatically merged.

.github/renovate.json5
rarkins: It's not common to ignore them, and I'd recommend against that - they're usually lower risk bug fixes. What you _can_ do is group and schedule patch updates so that you get e.g. 1 PR per week

.github/renovate.json5
rarkins: This setting will work for all passing updates - in-range or otherwise. You can perhaps skip automerging initially until more familiar with Renovate's approach?

.github/renovate.json5
mdjermanovic, line 1: In [typescript-eslint's config](https://github.com/typescript-eslint/typescript-eslint/blob/7826910d8c6ba4d2d40b8d55fcf7fd39fe290d88/.github/renovate.json5), there's this line:

```js
  ignorePaths: [
    // integration test package.json's should never be updated as they're purposely fixed tests
    'tests/integration/fixtures',
  ],
```

Does this mean that Renovate looks for package.json files in subdirectories too? If that's the case, we'd probably want to ignore some paths too, for instance [`docs/_examples`](https://github.com/eslint/eslint/tree/main/docs/_examples).

.github/renovate.json5
rarkins: Not currently - Renovate doesn't expose metadata about in-range vs not

mdjermanovic: Okay, then in such exceptional cases that rarely happen, we'll probably need to temporarily configure Renovate to ignore the problematic package, which seems fine to me.

.github/renovate.json5
mdjermanovic, line 1: Per https://github.com/eslint/eslint/pull/17567#discussion_r1327825312, updates for npm production dependencies will have `fix` prefix? Can we configure Renovate to always use `chore` prefix?

rarkins: The preset `config:recommended` adds this setting: https://docs.renovatebot.com/presets-default/#semanticprefixfixdepschoreothers

Otherwise, Renovate defaults to chore as the default: https://docs.renovatebot.com/configuration-options/#semanticcommittype

You can add `ignorePresets` with a value of `[":semanticPrefixFixDepsChoreOthers"]` to undo it, or else consider removing `config:recommended` and instead including some of its sub-presets individually: https://docs.renovatebot.com/presets-config/#configrecommended

.github/renovate.json5
HonkingGoose: > Okay, then in such exceptional cases that rarely happen, we'll probably need to temporarily configure Renovate to ignore the problematic package, which seems fine to me.

If you want to "ignore" a dependency, I recommend setting it to `dependencyDashboardApproval=true`. This means you can always get the update from Renovate's Dependency Dashboard. For example:

```json
{
  "packageRules": [
    {
      "matchPackagePatterns": ["^@somescope"],
      "dependencyDashboardApproval": true
    }
  ]
}
```

Some users ignore a dependency by disabling the package manager in Renovate. That means you don't see anything about the update. With my method above, you'll always be reminded about the update on the Dependency Dashboard.

.github/renovate.json5
HonkingGoose: @rarkins did you see the `":semanticCommitScopeDisabled"` in their `extends` array? According to the Renovate docs, this is what the preset does:

> Disable semantic commit scope for all commits and PR titles.
>
> ```json
> {
>   "semanticCommitScope": null
> }
> ```

rarkins: Yes, that will disable the part in brackets. e.g. instead of `chore(deps)`` they'll get `chore:`

mdjermanovic: Looks like `docs/_examples` isn't ignored by default, so we'll need to define our own `ignorePaths`.

JoshuaKGoldberg: Do we want to ignore those though? Both only have dependencies on the `"eslint"` package that I can see. Is it not a good idea to keep them up-to-date to the latest version?

.github/renovate.json5
JoshuaKGoldberg: 👍 removing this. By principle of least resistance I haven't added anything around grouping, but certainly can if asked to.

.github/renovate.json5
HonkingGoose: Up to date doc examples are good! You could automate the minor and patch level updates, and request the major update from the Dependency Dashboard, for example.

For the Renovate docs we even setup a special regex rule that updates the version of the Renovate program in the Markdown files. 🙂

mdjermanovic: Okay, I don't have a strong opinion on this, so we can keep the default ignore patterns and decide which updates we'd like to exclude when the PRs start coming in.

mdjermanovic (approved):
LGTM, thanks!

@rarkins, @HonkingGoose would you like to verify the latest changes before we merge this?

.github/renovate.json5
```diff
@@ -0,0 +1,42 @@
...
+    "extends": [
+        "config:recommended",
+        ":approveMajorUpdates",
+        ":semanticCommitScopeDisabled"
+    ],
+    "ignorePresets": [":semanticPrefixFixDepsChoreOthers"],
+    "internalChecksFilter": "strict",
+    "labels": ["dependencies"],
+
+    // Wait well over three days for any new package as a precaution against malicious publishes
```
HonkingGoose, line 12: You _can not_ use comments with plain JSON. 😉

Pick one of these solutions:

- Drop the comment and use plain JSON, or
- Convert to the `.json5` format to use comments. Renovate supports `.json5`

HonkingGoose (changes requested):
The Renovate config looks good, except that you _can't_ use comments in plain JSON. 😉

.github/renovate.json5
viceice: it's already a json5 file 😉

HonkingGoose: Whoops, I was trying the config on my fork of ESLint and accidentally used the plain JSON format...

I missed that the upstream uses JSON5 already. 🙃

.github/renovate.json5
```diff
@@ -0,0 +1,42 @@
+{
+    "$schema": "https://docs.renovatebot.com/renovate-schema.json",
+    "extends": [
+        "config:recommended",
+        ":approveMajorUpdates",
+        ":semanticCommitScopeDisabled"
+    ],
+    "ignorePresets": [":semanticPrefixFixDepsChoreOthers"],
+    "internalChecksFilter": "strict",
```
HonkingGoose, line 9:
```suggestion
```

By default, Renovate uses `internalChecksFilter=strict` [^1]. So you can drop the explicit configuration here, if you want.

[^1]: [Renovate docs, `internalChecksFilter` config option ](https://docs.renovatebot.com/configuration-options/#internalchecksfilter)

.github/renovate.json5
HonkingGoose:
```suggestion
    // Wait seven days for any new package as a precaution against malicious publishes.
```

The comment says: well over three days, but the Renovate config will wait seven days. The comment should match the config. 😉

HonkingGoose (approved):
Good to go, _after_ fixing two niggles. 😉

.github/renovate.json5
JoshuaKGoldberg: > well over three

["Well over"](https://www.merriam-webster.com/dictionary/well%20over) implies that it's intentionally more a number much larger than three. Is this not a common idiom? (I'm constantly learning about phrases I thought were but actually aren't, heh)

HonkingGoose: The minimum release age is set to three in one of Renovate's presets  to prevent holding a nom package that might get unpublished in the three day window. After that its user preference how long they want to wait to upgrade.

The idiom is fine, but to me it feels like there's a missing part. 🙂 But feel free to ignore me on this.

JoshuaKGoldberg: Backlinking for reference: https://docs.npmjs.com/policies/unpublish#packages-published-less-than-72-hours-ago

HonkingGoose (approved):

.github/renovate.json5
nzakas: @JoshuaKGoldberg let's just add that link into the comment? It did read as a little strange to me, too, but makes sense in the context of this link.

nzakas:
All right, here we go!!!!

HonkingGoose:
By the way, you still have a Dependabot config file in the repository.

HonkingGoose:
I expected Renovate to create a Dependency Dashboard, and new PRs. But I'm not seeing any...

It looks like Renovate is not enabled for this repository, or that the Mend hosted Renovate app is not yet installed into the ESLint organization?

See [Renovate docs, installing Renovate's hosted GitHub app](https://docs.renovatebot.com/getting-started/installing-onboarding/#hosted-githubcom-app) for more information. 😉

nzakas:
I just verified that Renovate is both installed into the ESLint org and enabled for this repo. I'm not sure what else to do.

rarkins:
I'll ask someone to check it.

HonkingGoose:
The `renovate-config-validator` run on my fork of `eslint/eslint` passes:

```
@HonkingGoose ➜ /workspaces/eslint (main) $ renovate-config-validator
 INFO: Validating .github/renovate.json5
 INFO: Config validated successfully`
```

I'm out of ideas what's going wrong here. I'll let the Renovate maintainers/admins look into it further.

nzakas:
Looks like PRs started getting generated: https://github.com/eslint/eslint/pull/17740

PR #17569, opened 2023-09-15
feat: Add `--no-warn-ignored` CLI option for flat config

domnantas:
#### Prerequisites checklist

[ ] Documentation update
[ ] Add autofix to a rule
[x] Add a CLI option
[ ] Add something to the core
[ ] Other, please explain:

This is related to https://github.com/eslint/rfcs/pull/90 and https://github.com/eslint/eslint/issues/15010

- Added --no-warn-ignored CLI option
- Removed inconsistent warnIgnored defaults **for FlatESLint**, now it always defaults to `true` regardless if you're using CLI, or `lintFiles()`, or `lintText()`.
- Updated warning messages to inform users about the ability to suppress the warning

- Currently the help text is under `Handle Warnings` heading. RFC had a [short discussion](https://github.com/eslint/rfcs/pull/90#discussion_r889596069) on the help message placement where maintainers agreed on putting the `--no-warn-ignored` under `Miscellaneous`. Should I move it under `Miscellaneous`, or keep it under `Handle Warnings`?
- I have updated the warning message with `Use \"--no-warn-ignored\" to suppress this warning.` This was not discussed in the RFC

### DIFF
--- a/docs/src/use/command-line-interface.md
+++ b/docs/src/use/command-line-interface.md
@@ -110,6 +110,7 @@ Miscellaneous:
   --env-info                      Output execution environment information - default: false
   --no-error-on-unmatched-pattern  Prevent errors when pattern is unmatched
   --exit-on-fatal-error           Exit with exit code 2 in case of fatal error - default: false
+  --no-warn-ignored               Suppress warning when the file list includes ignored files. *Flat Config Mode Only*
   --debug                         Output debugging information
   -h, --help                      Show help
   -v, --version                   Output the version number
@@ -701,6 +702,18 @@ This option causes ESLint to exit with exit code 2 if one or more fatal parsing 
 
 ```shell
 npx eslint --exit-on-fatal-error file.js
+```
+
+#### `--no-warn-ignored`
+
+**Flat Config Mode Only.** This option suppresses `File ignored by default / File ignored because of a matching ignore pattern` warnings when an ignored filename is passed explicitly. It is useful when paired with `--max-warnings 0` as it will prevent exit code 1 due to the aforementioned warning.
+
+* **Argument Type**: No argument.
+
+##### `--no-warn-ignored` example
+
+```shell
+npx eslint --no-warn-ignored --max-warnings 0 ignored-file.js
 ```
 
 #### `--debug`
--- a/docs/src/use/command-line-interface.md
+++ b/docs/src/use/command-line-interface.md
@@ -110,7 +110,7 @@ Miscellaneous:
   --env-info                      Output execution environment information - default: false
   --no-error-on-unmatched-pattern  Prevent errors when pattern is unmatched
   --exit-on-fatal-error           Exit with exit code 2 in case of fatal error - default: false
-  --no-warn-ignored               Suppress warning when the file list includes ignored files. *Flat Config Mode Only*
+  --no-warn-ignored               Suppress warnings when the file list includes ignored files. *Flat Config Mode Only*
   --debug                         Output debugging information
   -h, --help                      Show help
   -v, --version                   Output the version number
@@ -706,7 +706,7 @@ npx eslint --exit-on-fatal-error file.js
 
 #### `--no-warn-ignored`
 
-**Flat Config Mode Only.** This option suppresses `File ignored by default / File ignored because of a matching ignore pattern` warnings when an ignored filename is passed explicitly. It is useful when paired with `--max-warnings 0` as it will prevent exit code 1 due to the aforementioned warning.
+**Flat Config Mode Only.** This option suppresses both `File ignored by default` and `File ignored because of a matching ignore pattern` warnings when an ignored filename is passed explicitly. It is useful when paired with `--max-warnings 0` as it will prevent exit code 1 due to the aforementioned warning.
 
 * **Argument Type**: No argument.
 
--- a/docs/src/use/configure/ignore.md
+++ b/docs/src/use/configure/ignore.md
@@ -149,7 +149,7 @@ You'll see this warning:
 
 ```text
 foo.js
-  0:0  warning  File ignored because of a matching ignore pattern. Use "--no-ignore" to override. Use \"--no-warn-ignored\" to suppress this warning.
+  0:0  warning  File ignored because of a matching ignore pattern. Use "--no-ignore" to disable file ignore settings or use "--no-warn-ignored" to suppress this warning.
 
 ✖ 1 problem (0 errors, 1 warning)
 ```
--- a/docs/src/use/configure/ignore.md
+++ b/docs/src/use/configure/ignore.md
@@ -149,7 +149,7 @@ You'll see this warning:
 
 ```text
 foo.js
-  0:0  warning  File ignored because of a matching ignore pattern. Use "--no-ignore" to override.
+  0:0  warning  File ignored because of a matching ignore pattern. Use "--no-ignore" to override. Use \"--no-warn-ignored\" to suppress this warning.
 
 ✖ 1 problem (0 errors, 1 warning)
 ```
--- a/lib/cli.js
+++ b/lib/cli.js
@@ -91,7 +91,8 @@ async function translateOptions({
     reportUnusedDisableDirectives,
     resolvePluginsRelativeTo,
     rule,
-    rulesdir
+    rulesdir,
+    warnIgnored
 }, configType) {
 
     let overrideConfig, overrideConfigFile;
@@ -182,6 +183,7 @@ async function translateOptions({
 
     if (configType === "flat") {
         options.ignorePatterns = ignorePattern;
+        options.warnIgnored = warnIgnored;
     } else {
         options.resolvePluginsRelativeTo = resolvePluginsRelativeTo;
         options.rulePaths = rulesdir;
--- a/lib/cli.js
+++ b/lib/cli.js
@@ -387,8 +387,9 @@ const cli = {
         if (useStdin) {
             results = await engine.lintText(text, {
                 filePath: options.stdinFilename,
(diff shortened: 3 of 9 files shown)


mdjermanovic:
Thanks for the PR!

> * Currently the help text is under `Handle Warnings` heading. RFC had a [short discussion](https://github.com/eslint/rfcs/pull/90#discussion_r889596069) on the help message placement where maintainers agreed on putting the `--no-warn-ignored` under `Miscellaneous`. Should I move it under `Miscellaneous`, or keep it under `Handle Warnings`?

I still think it should be under `Miscellaneous`, for the reasons described in https://github.com/eslint/rfcs/pull/90#discussion_r889596069

> * I have updated the warning message with `Use \"--no-warn-ignored\" to suppress this warning.` This was not discussed in the RFC

Looks good to me. It was planned in one of the previous versions of the RFC (see the text above https://github.com/eslint/rfcs/pull/90#discussion_r889595519) and we agreed on it.

mdjermanovic (changes requested):
When used with `--stdin`, `--no-warn-ignored` does not seem to apply:

```
$ echo foo | npx eslint --stdin --stdin-filename test.js --no-warn-ignored

C:\milos\eslint\test.js
  0:0  warning  File ignored because of a matching ignore pattern. Use "--no-ignore" to override. Use "--no-warn-ignored" to suppress this warning
✖ 1 problem (0 errors, 1 warning)
```

I think it should apply. Removing `warnIgnored: true` (but only in the flat config mode), as you suggested in https://github.com/eslint/rfcs/pull/90#discussion_r892144145, will probably do the work:

https://github.com/eslint/eslint/blob/20893d48b9012f2b61bbbfeac8bee70d68d90e5e/lib/cli.js#L385-L389

lib/cli.js
```diff
@@ -385,7 +387,8 @@ const cli = {
         if (useStdin) {
             results = await engine.lintText(text, {
                 filePath: options.stdinFilename,
-                warnIgnored: true
+                // eslint-disable-next-line no-undefined -- flatConfig respects CLI flag and constructor warnIgnored, eslintrc forces true for backwards compatibility
+                warnIgnored: usingFlatConfig ? undefined : true
```
mdjermanovic, line 391:
```suggestion

                // flatConfig respects CLI flag and constructor warnIgnored, eslintrc forces true for backwards compatibility
                warnIgnored: usingFlatConfig ? void 0 : true
```

We're using `void 0` to get the `undefined` value.

tests/lib/eslint/flat-eslint.js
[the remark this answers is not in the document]
domnantas: Good point, added a couple of test cases

lib/eslint/eslint-helpers.js
```diff
@@ -781,6 +782,9 @@ function processOptions({
     ) {
         errors.push("'reportUnusedDisableDirectives' must be any of \"error\", \"warn\", \"off\", and null.");
     }
+    if (typeof warnIgnored !== "boolean") {
+        errors.push("'warnIgnored' must be a boolean.");
+    }
```
mdjermanovic, line 787: Can you update the following test to make sure that we're validating the new option:

https://github.com/eslint/eslint/blob/1ea4cfb585dcb52ac3cb1522a32f25cfe507121b/tests/lib/eslint/flat-eslint.js#L197-L235

mdjermanovic (changes requested):
One small request, and then LGTM.

Also, I'm not sure if and where we could document the new CLI option. [Command Line Interface Reference](https://eslint.org/docs/latest/use/command-line-interface) is currently for eslintrc CLI options, and [Configuration Files (New)](https://eslint.org/docs/latest/use/configure/configuration-files-new) is only about config files. @nzakas any ideas?

docs/src/use/configure/ignore.md
```diff
@@ -149,7 +149,7 @@ You'll see this warning:
 
 ```text
 foo.js
-  0:0  warning  File ignored because of a matching ignore pattern. Use "--no-ignore" to override.
+  0:0  warning  File ignored because of a matching ignore pattern. Use "--no-ignore" to override. Use \"--no-warn-ignored\" to suppress this warning.
```
nzakas, line 152: Just cleaning up the language a bit.

```suggestion
  0:0  warning  File ignored because of a matching ignore pattern. Use "--no-ignore" to disable file ignore settings or use \"--no-warn-ignored\" to suppress this warning.
```

lib/eslint/eslint-helpers.js
```diff
@@ -594,9 +594,9 @@ function createIgnoreResult(filePath, baseDir) {
     const isInNodeModules = baseDir && path.dirname(path.relative(baseDir, filePath)).split(path.sep).includes("node_modules");
 
     if (isInNodeModules) {
-        message = "File ignored by default because it is located under the node_modules directory. Use ignore pattern \"!**/node_modules/\" to override.";
+        message = "File ignored by default because it is located under the node_modules directory. Use ignore pattern \"!**/node_modules/\" to override. Use \"--no-warn-ignored\" to suppress this warning.";
     } else {
-        message = "File ignored because of a matching ignore pattern. Use \"--no-ignore\" to override.";
+        message = "File ignored because of a matching ignore pattern. Use \"--no-ignore\" to override. Use \"--no-warn-ignored\" to suppress this warning.";
```
nzakas, line 599: Just a note to update these if my rewording of the message is accepted.

lib/eslint/flat-eslint.js
```diff
@@ -922,7 +928,7 @@ class FlatESLint {
             throw new Error("'options.filePath' must be a non-empty string or undefined");
         }
 
-        if (typeof warnIgnored !== "boolean") {
+        if (!(typeof warnIgnored === "boolean" || typeof warnIgnored === "undefined")) {
```
nzakas, line 931: This might be a bit easier to understand.

```suggestion
        if (typeof warnIgnored !== "boolean" && typeof warnIgnored !== "undefined")) {
```

lib/eslint/eslint-helpers.js
domnantas: I have expanded the assertion, but I suspect the test might be flaky – I wasn't able to get it to fail. I tried commenting out options and the strings for regex construction, but the test was still passing.

nzakas (changes requested):
Looks good to me, I'd just like to clean up the messaging a bit.

For documentation, I think it makes sense to add to the CLI documentation page and just put in bold "Flat Config Mode Only".

lib/eslint/eslint-helpers.js
domnantas: It passes when `warnIgnored: ""` is in the test, but the error string is commented out. Is that expected?

image

mdjermanovic: > It passes when `warnIgnored: ""` is in the test, but the error string is commented out. Is that expected?

It's because the regex looks for a substring only, so when we append more text to the actual error the expected is still passing. It's okay, you updated the expected error so if we mistakenly remove the validation this test will fail.

tests/lib/eslint/flat-eslint.js
```diff
@@ -230,6 +231,8 @@ describe("FlatESLint", () => {
                     "- 'overrideConfigFile' must be a non-empty string, null, or true.",
                     "- 'plugins' must be an object or null.",
                     "- 'reportUnusedDisableDirectives' must be any of \"error\", \"warn\", \"off\", and null."
+
+                    // "- 'warnIgnored' must be a boolean."
```
mdjermanovic, line 235:
```suggestion
                    "- 'reportUnusedDisableDirectives' must be any of \"error\", \"warn\", \"off\", and null.",
                    "- 'warnIgnored' must be a boolean."
```

mdjermanovic:
> For documentation, I think it makes sense to add to the CLI documentation page and just put in bold "Flat Config Mode Only".

Agreed. @domnantas can you update [`command-line-interface.md`](https://github.com/eslint/eslint/blob/main/docs/src/use/command-line-interface.md)?

tests/lib/eslint/flat-eslint.js
domnantas: oops 😬

docs/src/use/command-line-interface.md
```diff
@@ -703,6 +704,18 @@ This option causes ESLint to exit with exit code 2 if one or more fatal parsing
 npx eslint --exit-on-fatal-error file.js
 ```
 
+#### `--no-warn-ignored`
+
+**Flat Config Mode Only.** This option suppresses `File ignored by default / File ignored because of a matching ignore pattern` warnings when an ignored filename is passed explicitly. It is useful when paired with `--max-warnings 0` as it will prevent exit code 1 due to the aforementioned warning.
```
nzakas, line 709: Just to clean up a bit.
```suggestion
**Flat Config Mode Only.** This option suppresses both `File ignored by default` and `File ignored because of a matching ignore pattern` warnings when an ignored filename is passed explicitly. It is useful when paired with `--max-warnings 0` as it will prevent exit code 1 due to the aforementioned warning.
```

lib/options.js
```diff
@@ -139,6 +140,17 @@ module.exports = function(usingFlatConfig) {
...
     }
 
+    let warnIgnoredFlag;
+
+    if (usingFlatConfig) {
+        warnIgnoredFlag = {
+            option: "warn-ignored",
+            type: "Boolean",
+            default: "true",
+            description: "Suppress warning when the file list includes ignored files"
```
nzakas, line 150:
```suggestion
            description: "Suppress warnings when the file list includes ignored files"
```

nzakas (approved):
LGTM. Just left a couple suggestions for clarifying wording.

mdjermanovic (approved):
LGTM, thanks!

PR #17640, opened 2023-10-10
docs: Plugin flat config migration guide

### LINKED ISSUE #17242: Change Request: document flat configs in plugins
### ESLint version

8.42.0

Current [documentation page on writing plugins](https://eslint.org/docs/latest/extend/plugins) uses legacy configs

It should provide a useful example of how a flat config can be included in a plugin without circular imports.

### Participation

### Additional comments

_No response_

nzakas:
#### Prerequisites checklist

[x] Documentation update
[ ] Add autofix to a rule
[ ] Add a CLI option
[ ] Add something to the core
[ ] Other, please explain:

Added a migration guide for plugins to implement flat config. This focuses on the `configs` key but also touches on everything else that plugins can export.

Fixes #17242

Did I miss any details? Do the examples make sense?

ping @bradzacher @JoshuaKGoldberg

### DIFF
--- a/docs/src/extend/plugin-migration-flat-config.md
+++ b/docs/src/extend/plugin-migration-flat-config.md
@@ -21,7 +21,7 @@ const plugin = {
     processors: {}
 };
 
-export { plugin };
+export default plugin;
 ```
 
 This structure allows the most flexibility when making other changes discussed on this page.
@@ -48,7 +48,7 @@ const plugin = {
     }
 };
 
-export { plugin };
+export default plugin;
 ```
 
 The name `".md"` is no longer valid for a processor, so it must be replaced with a valid identifier such as `markdown`:
@@ -67,7 +67,7 @@ const plugin = {
     }
 };
 
-export { plugin };
+export default plugin;
 ```
 
 In order to use this renamed processor, you'll also need to manually specify it inside of a config, such as:
@@ -134,7 +134,7 @@ Object.assign(plugin.configs, {
     }
 })
 
-export { plugin };
+export default plugin;
 ```
 
 Your users can then use this exported config like this:
@@ -205,7 +205,7 @@ Object.assign(plugin.configs, {
     }
 })
 
-export { plugin };
+export default plugin;
 ```
 
 Your users can then use this exported config like this:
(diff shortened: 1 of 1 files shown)


bradzacher (commented):
The thing that's missing from my POV is how a plugin/config package should support both `.eslintrc` and flat configs.

Whilst, yes, v9 will switch the default - I have no doubt that there will be a long-tail of users that take a long time to (a) upgrade to v9 and (b) migrate to flat configs.

If we switch our configs over to only support flat configs then the majority of our users will be "left in the lurch" and thus will be stuck.

For example I know that at Canva or at Meta - migrating the eslint config would likely be a multi-week project given how many overrides there are - it will take a lot of testing to validate the migration is 1:1. Such a project requires prioritisation and scheduling which can delay it for quite some time.

So to summarise - I think these docs need to answer the question "how can I support flat configs and legacy configs simultaneously?"

docs/src/extend/plugin-migration-flat-config.md
```diff
@@ -0,0 +1,293 @@
...
+
+1. **Export a CommonJS entrypoint.** The old configuration system cannot load plugins that are published only in ESM format. If your source code is in ESM, then you'll need to use a bundler that can generate a CommonJS version and use the [`exports`](https://nodejs.org/api/packages.html#package-entry-points) key in your `package.json` file to ensure the CommonJS version can be found by Node.js.
+1. **Keep the `environments` key.** If your plugin exports custom environments, you should keep those as they are and also export the equivalent flat configs as described above. The `environments` key is ignored when ESLint is running in flat config mode.
+1. **Export both eslintrc and flat configs.** The `configs` key is only validated when a config is used, so you can provide both formats of configs in the `configs` key. We recommend that you append older format configs with `-legacy` to make it clear that these configs will not be supported in the future. For example, if your primary config is called `recommended` and is in flat config format, then you can also have a config named `recommended-legacy` that is the eslintrc config format.
```
bradzacher, line 287: I guess delineating the config name with `-legacy` is probably the absolute best way to ensure that you can keep the backwards compat and cleanly define both configs without muddling the new vs old keys and such.

bradzacher (commented):
that's looking good to me - thanks for the clarification!

mdjermanovic (approved):
LGTM, just one typo.

jaydenseric:
Many legacy ESLint plugins published to npm don't contain types because the authors didn't expect people to be importing the plugins and using them in code. TypeScript users will now greatly appreciate the published plugin packages having types, so they can be imported and used in project code with type safety. It would be good to add to the migration guide that flat config plugins should be typed (if you want to support TypeScript projects), and how to type them.

PR #17656, opened 2023-10-16
feat: add meta.defaultOptions

JoshuaKGoldberg:
#### Prerequisites checklist

[ ] Documentation update
[ ] Add autofix to a rule
[ ] Add a CLI option
[x] Add something to the core
[ ] Other, please explain:

Augments linting to factor in a new optional `meta.defaultOptions`. If a rule describes default options this way, any user-provided options are merged on top of the default options.

These defaults are applied by a new `deepMergeArrays` utility in:

* `ConfigValidator`
* `Linter` in two places:
    * For configs: by augmenting the existing `getRuleOptions` helper
    * For inline comments in flat config: with a new `getRuleOptionsInline` helper
        * Inline comments in legacy configs go through `RuleValidator` instead
* `RuleValidator`

* Issue: https://github.com/eslint/eslint/issues/17448
* RFC: https://github.com/eslint/rfcs/pull/113

Co-authored-by: Milos Djermanovic <milos.djermanovic@gmail.com>


tests/lib/linter/deep-merge.js
```diff
@@ -0,0 +1,61 @@
...
+        [123, undefined, 123],
+        [123, null, null],
+        [123, { a: 0 }, { a: 0 }],
+        ["abc", undefined, "abc"],
+        ["abc", null, null],
+        ["abc", { a: 0 }, { a: 0 }],
+        [["abc"], undefined, ["abc"]],
+        [["abc"], null, null],
+        [[], ["def"], []],
+        [["abc"], ["def"], ["abc"]]
```
mdjermanovic, line 53: Shouldn't the result of merging these be the second array?

```suggestion
        [[], ["def"], ["def"]],
        [["abc"], ["def"], ["def"]]
```

Rec0iL99:
Not stale. @JoshuaKGoldberg there's a review comment for you to look into. Thanks.

nzakas:
@JoshuaKGoldberg can you please rebase and address the conflicts?

JoshuaKGoldberg:
The browser tests aren't failing for me locally. Test flake?

nzakas:
Yes, the browser tests can be a little bit flaky, but they look okay right now.

nzakas (changes requested):
This is looking really good! I left some questions and suggestions throughout, mostly for clarity.

As part of this work, can you also update the docs related to create custom rules?

lib/linter/linter.js
```diff
@@ -1005,7 +1023,7 @@ function runRules(sourceCode, configuredRules, ruleMapper, parserName, languageO
                 Object.create(sharedTraversalContext),
                 {
                     id: ruleId,
-                    options: getRuleOptions(configuredRules[ruleId]),
+                    options: getRuleOptions(rule.meta && rule.meta.defaultOptions, configuredRules[ruleId]),
```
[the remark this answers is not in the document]
JoshuaKGoldberg: Heh, I'd forgotten ESLint core can use that now!

tests/lib/config/flat-config-array.js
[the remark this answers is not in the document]
JoshuaKGoldberg: `camelcase.js` in `main` uses ajv schema defaults for `ignoreDestructuring`, `ignoreImports`, and `ignoreGlobals` - but not for `properties` or `allow`. This PR switches the rule to providing defaults for all five option properties. So the resultant options in the flat config now have all five properties.

We'd have to make the same change to this test file if we added defaults for `allow` and `properties` in `camelcase.js` on the `main` branch.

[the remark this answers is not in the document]
JoshuaKGoldberg: `camelcase.js` in `main` uses ajv schema defaults for `ignoreDestructuring`, `ignoreImports`, and `ignoreGlobals` - but not for `properties` or `allow`. This PR switches the rule to providing defaults for all five option properties. So the resultant options in the flat config now have all five properties.

We'd have to make the same change to this test file if we added defaults for `allow` and `properties` in `camelcase.js` on the `main` branch.

Other than changing the printed options config for the `camelcase` rule, I don't think this is a behavioral change. It shouldn't impact anything at runtime. Arguably I'd say it fixes a bug in that those values options practically did have those defaults, but didn't reflect that in the options schema. 😄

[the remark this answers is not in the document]
nzakas: Gotcha. :+1:

docs/src/extend/custom-rules.md
```diff
@@ -787,6 +789,47 @@ module.exports = {
...
 
+#### Option Defaults
+
+Rules may specify a `meta.defaultOptions` array of default values for any options.
+When the rule is enabled in a user configuration, ESLint will recursively merge any user-provided option elements on top of the default elements.
+
+For example, given the following defaults:
+
+```js
+module.exports = {
```
nzakas, line 800: We've been moving our docs to ESM by default.
```suggestion
export default {
```

docs/src/extend/custom-rules.md
```diff
@@ -787,6 +789,47 @@ module.exports = {
...
+        return { /* ... */ };
+    }
+}
+```
+
+The rule would have a runtime `alias` value of `"basic"` unless the user configuration specifies a different value, such as with `["error", { alias: "complex" }]`.
+
+Each element of the options array is merged according to the following rules:
+
+* User-provided `undefined` uses any default option
```
nzakas, line 827: The wording of this bullet is a bit confusing to me. Do you mean a hardcoded `undefined` or an undefined as in it wasn't specified? And by "any default option", I think you mean that the specified default in `defaultOptions` is used?

nzakas (changes requested):
This is looking really good. Just left a few notes on the docs to see if we can clean it up a bit.

docs/src/extend/custom-rules.md
JoshuaKGoldberg: Heh, yeah, this confused me re-reading it.

nzakas (approved):
LGTM, thanks! Would like @mdjermanovic to review before merging.

docs/src/extend/custom-rules.md
```diff
@@ -787,6 +789,47 @@ module.exports = {
 
 To learn more about JSON Schema, we recommend looking at some examples on the [JSON Schema website](https://json-schema.org/learn/), or reading the free [Understanding JSON Schema](https://json-schema.org/understanding-json-schema/) ebook.
 
+#### Option Defaults
```
bmish, line 792: Should we mention how this interacts with `default` values from the schema and which one is recommended?

bmish (dismissed):
Nice!

I also filed:
* https://github.com/eslint-community/eslint-plugin-eslint-plugin/issues/430
* https://github.com/bmish/eslint-doc-generator/issues/513

Rec0iL99:
> Hi everyone, it looks like we lost track of this pull request. Please review and see what the next steps are. This pull request will auto-close in 7 days without an update.

Waiting for @mdjermanovic's review before merging

mdjermanovic:
Maybe we could simplify `function deepMergeObjects` like this:

```js
function deepMergeObjects(first, second) {
    if (second === void 0) {
        return first;
    }

    if (!isObjectNotArray(first) || !isObjectNotArray(second)) {
        return second;
    }

    const result = {
        ...first,
        ...second
    };

    for (const key of Object.keys(second)) {
        if (Object.prototype.propertyIsEnumerable.call(first, key)) {
            result[key] = deepMergeObjects(first[key], second[key]);
        }
    }

    return result;
}
```

lib/shared/deep-merge-arrays.js
```diff
@@ -0,0 +1,78 @@
+/**
+ * @fileoverview Applies default rule options
+ * @author JoshuaKGoldberg
+ */
+/* eslint-disable eqeqeq, no-undefined -- `null` and `undefined` are different in options */
```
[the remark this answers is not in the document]
JoshuaKGoldberg: Thanks - applied to this file.

Do you also want the rules re-enabled in the test file too? I have a mild preference for using `undefined` to be more close to how users will enter them. But I also am accustomed to avoiding `void 0` and don't know how valid that instinct is. 🙂

JoshuaKGoldberg:
Wow! Really nice refactor, thanks! 👏

tests/lib/linter/linter.js
```diff
@@ -6737,6 +6760,49 @@ var a = "test2";
...
+                    "my-rule": "error"
+                }
+            };
+
+            const code = "";
+            const messages = linter.verify(code, config);
+
+            assert.deepStrictEqual(
+                JSON.parse(messages[0].message),
+                { inBoth: "from-default-options", inDefaultOptions: "from-default-options" }
```
JoshuaKGoldberg, line 6801: Spinning out from https://github.com/eslint/eslint/pull/17656#discussion_r1475266210:

> Should we mention how `defaultOptions` interacts with `default` values from the schema and which one is recommended?

I'd like to add a unit test that shows the end-to-end behavior. But I can't figure out where that unit test should go. Does anybody have any advice?

This test here shows the same behavior as on `main`: for custom rules, the `default`s in `schema` properties aren't applied.

lib/shared/deep-merge-arrays.js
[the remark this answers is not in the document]
mdjermanovic: Yes, for consistency as we have the `no-undefined` rule enabled. (we could discuss separately if we want to keep this rule enabled.

lib/linter/linter.js
```diff
@@ -778,14 +779,31 @@ function stripUnicodeBOM(text) {
...
  * Get the options for a rule (not including severity), if any
  * @param {Array|number} ruleConfig rule configuration
+ * @param {Object|undefined} defaultOptions rule.meta.defaultOptions
  * @returns {Array} of rule options, empty Array if none
  */
-function getRuleOptions(ruleConfig) {
+function getRuleOptions(ruleConfig, defaultOptions) {
     if (Array.isArray(ruleConfig)) {
-        return ruleConfig.slice(1);
+        return deepMergeArrays(defaultOptions, ruleConfig.slice(1));
```
mdjermanovic, line 787: In flat config case, `defaultOptions` have already been merged during the validation so this would merge them again?

tests/lib/linter/linter.js
nzakas: Are you asking if it belongs in `linter.js`? Or are you asking where it belongs within `linter.js`?

It would definitely go in `linter.js`, and I'd just put a new `describe` for this.

Note that `linter.js` is split into two sections: eslintrc config and flat config, so you need to add a test in both sections.

JoshuaKGoldberg:
Ah, timing - I'm going through the code now. Will respond soon.

lib/linter/linter.js
JoshuaKGoldberg: Aha! Yes, you're right. And there's an additional bug, I think, that default values from schemas aren't being applied in the branch right now. 0dc481013 has a passing unit test for flat config and a failing unit test for legacy config.

```plaintext
1) Linter
       options
         rules should apply meta.defaultOptions on top of schema defaults:

      AssertionError: expected { Object (inBoth, inDefaultOptions) } to deeply equal { Object (inBoth, inDefaultOptions, ...) }
      + expected - actual

       {
         "inBoth": "from-default-options"
         "inDefaultOptions": "from-default-options"
      +  "inSchema": "from-schema"
       }
```

I _think_ what needs to happen is moving the validation so it happens once for each config, and with both meta-based and schema-based default options applied. Working on that now.

---

Adding my notes as I found the code flow hard to keep track of without them:
* `Linter` takes in a `configType` of type `"eslintrc" | "flat"`
    * For `"eslintrc"`, it calls `this._verifyWithConfigArray` -> ... -> `_verifyWithoutProcessors` -> `runRules`
    * For `"flat"`, it calls `this._verifyWithFlatConfigArray` -> ... -> `_verifyWithFlatConfigArrayAndWithoutProcessors` -> `runRules`
        * In `_verifyWithFlatConfigArrayAndWithoutProcessors`, `const ruleValidator = new RuleValidator()` calls to a `validate()` method. That uses a `this.validators` to build validators from `ajv.compile(schema)`. This applies **schema** default options.
        * Later in `validate()`, `deepMergeArrays` is called with `deepMergeArrays`. This applies **meta** default options.
* `runRules` calls `getRuleOptions(configuredRules[ruleId], rule.meta?.defaultOptions)`. This applies **meta** default options.

With both flat and legacy configs, rule validation happens inside `runRules`.

With flat configs, rule validation additionally happens _before_ `runRules`. This earlier validation is applied to the result of merging both **meta** and **schema** default options.

With legacy configs, only the `runRules` validation is applied. That's why it doesn't apply to **schema** default options.

JoshuaKGoldberg: OK, sorry, I'm thoroughly confused and think I need help here. On `main`, `Linter`'s unit tests never apply **schema** defaults - neither for flat nor legacy config. https://github.com/JoshuaKGoldberg/eslint/commit/9ccaa85d81f6b108cf92bcea458dc80437698b6a shows a passing unit test per config type showing schema defaults not being applied.

In the flat config code path, `Linter` calls to `ruleValidator.validate(...)`. That means Ajv's options validation and **schema** defaulting applies there as well as in `flat-config-array.js`.

So now I'm unsure what the intended boundaries are supposed to be. Do we want **schema** defaulting to be moved to happen inside `Linter` now? Is that a breaking change?

tests/lib/linter/linter.js
JoshuaKGoldberg: The question here on boundaries of when to apply schema defaults is relevant, I think: https://github.com/eslint/eslint/pull/17656/files#r1491809729

tests/lib/linter/linter.js
```diff
@@ -6739,6 +6762,52 @@ var a = "test2";
         });
     });
 
+    describe("options", () => {
+        it("rules should apply meta.defaultOptions on top of schema defaults", () => {
```
JoshuaKGoldberg: Note that this is different from `main`. I asked about the code intent: https://github.com/eslint/eslint/pull/17656#discussion_r1491809729

lib/linter/linter.js
mdjermanovic: Just to clarify what happens with schema defaults.

Schema defaults in the top-level array never apply. That is, I believe, an unintentional omission (i.e., a bug) in the code that handles provided options. We could have considered fixing that for v9 if we hadn't decided to switch to `defaultOptions` and eventually drop schema defaults. So, in https://github.com/JoshuaKGoldberg/eslint/commit/9ccaa85d81f6b108cf92bcea458dc80437698b6a, `default: { inSchema: "from-schema-root" }` is a no-op. On the other hand, `inSchema: { default: "from-schema", type: "string" }` can apply, but you'd need to pass `"my-rule": ["error", {}]`. This should work in config _files_, both eslintrc and flat config. This should also work when `Linter` is used directly in flat config mode. But it won't work when `Linter` is used directly in eslintrc mode, because in eslintrc mode Linter doesn't validate rule options, which is probably another unintentional omission. I'm not sure if it's worth fixing that at this point, because eslintrc mode is deprecated and will be dropped in v10.

JoshuaKGoldberg: Got it, thanks - that's very helpful.

> should work in config _files_, both eslintrc and flat config
> won't work when `Linter` is used directly in eslintrc mode

Just confirming then - is the behavior in the latest commit, adb7972d9, an acceptable one for v9? That its `Linter` unit test for flat config applies on top of schema defaults, while its unit test for legacy config ignores schema defaults?

nzakas: I think if we are maintaining the current eslintrc behavior, then we are okay. @mdjermanovic?

Rec0iL99:
Waiting for @mdjermanovic to review here.

@JoshuaKGoldberg there's a conflict to resolve.

lib/linter/linter.js
Rec0iL99: ping @mdjermanovic

JoshuaKGoldberg:
I believe it's waiting on re-review from @mdjermanovic?

There were a few merge conflicts introduced since the last merge from `main`. I resolved them just now.

mdjermanovic:
Sorry for the delay. A problem is that eslintrc will still be supported in ESLint v9, and rules should work the same regardless of which config system is used.

In flat config, default options are merged before validation. I think this generally makes more sense than validating provided options first and then merging defaults. Ajv also applies schema defaults before validating options. In eslintrc mode, however, as currently implemented, default options are merged after validation (right before running rules).

This means that the same configuration for a rule could be valid in one config system but invalid in the other.

For example, if this is a rule:

```js
{
    meta: {
        defaultOptions: [{
            foo: 42
        }],
        schema: [{
            type: "object",
            maxProperties: 2 // allows one or two in addition to "foo"?
        }]
    },
    create() {
        return {};
    }
}
```

configuration `["error", { bar: 6, baz: 7 }]` would be valid in `.eslintrc.js` file, but invalid in `eslint.config.js` file (`Value {"foo":42,"bar":6,"baz":7} should NOT have more than 2 properties`).

I think we should address this, but there are two possible ways:

1. Change eslintrc to apply defaults before validating options.
2. Change flat config to apply defaults after validating options.

I'd be in favor of 1), although it probably requires changes in `@eslint/eslintrc` package. 2) seems easier to implement at the moment, but it's unlikely that we'd ever change this behavior later, so if we think that merging defaults should be before the validation, it's probably better to do 1). Thoughts?

lib/config/rule-validator.js
```diff
@@ -164,7 +165,10 @@ class RuleValidator {
 
             if (validateRule) {
 
-                validateRule(ruleOptions.slice(1));
+                const slicedOptions = ruleOptions.slice(1);
+                const mergedOptions = deepMergeArrays(rule.meta?.defaultOptions, slicedOptions);
```
mdjermanovic, line 169: Defaults options wouldn't be applied if the rule has `schema: false`?

Tanujkanti4441:
Hi @JoshuaKGoldberg, some reviews are there you can checkout.

JoshuaKGoldberg:
> I'd be in favor of 1), although it probably requires changes in @eslint/eslintrc package.

I'm in favor of that as well. Is there some process I should follow?

nzakas:
@JoshuaKGoldberg nothing special, just go ahead and open a PR on `eslint/eslintrc` and mention this PR in it.

JoshuaKGoldberg:
I still plan on continuing this work, but likely won't be able to for at least the month of May. ESLint v9 & typescript-eslint v8 are much higher priority for me at the moment.

If anybody else wants to tackle the `eslint/eslintrc` changes in the meantime, please do go ahead! That would be lovely 🙂. But if not, I should be able to later this summer.

lib/linter/linter.js
mdjermanovic: I think it's okay to keep the current behavior regarding validation and schema defaults when `Linter` is used directly in eslintrc mode. That is, rule options passed in the config are not validated and JSON schema defaults are not applied to them. We'd probably want to change this if the eslintrc config system wasn't deprecated.

But the new feature `meta.defaultOptions` should apply when `Linter` is used directly in eslintrc mode.

lib/linter/linter.js
```diff
@@ -1723,9 +1756,7 @@ class Linter {
 
                         try {
 
-                            let ruleOptions = Array.isArray(ruleValue) ? ruleValue : [ruleValue];
-
-                            assertIsRuleSeverity(ruleId, ruleOptions[0]);
+                            let ruleOptions = getRuleOptionsInline(ruleId, ruleValue, rule.meta?.defaultOptions);
```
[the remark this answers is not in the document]
JoshuaKGoldberg: Nice, there's now a failing unit test to show this too!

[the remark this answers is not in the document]
JoshuaKGoldberg: I made it more explicit in 3109906f4.

nzakas:
@mdjermanovic this is ready for another review.

@JoshuaKGoldberg it looks like there's a conflict to resolve.

mdjermanovic:
Sorry for the delay. There's a lot to check here: changes in the core + each of the 90 rules. I'll try to allocate time for this soon.

lib/config/rule-validator.js
mdjermanovic, line 169: I think the default options should apply regardless of whether the options are validated (i.e., even if the rule has `schema: false`).

lib/linter/linter.js
```diff
@@ -1796,7 +1828,7 @@ class Linter {
 
                             try {
 
-                                let ruleOptions = Array.isArray(ruleValue) ? ruleValue : [ruleValue];
+                                let ruleOptions = getRuleOptionsInline(ruleId, ruleValue, rule.meta?.defaultOptions);
```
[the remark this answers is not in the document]
JoshuaKGoldberg: I tried that out but still ran into inline comments not having default options applied. 9c984b3b1 goes a bit of a different direction and applies them directly in `getDirectiveComments` - what do you think?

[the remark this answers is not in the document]
mdjermanovic: Hm, when I tried https://github.com/eslint/eslint/pull/17656#discussion_r1788653623 locally, all tests were passing. I'll try again with the latest version of this branch.

lib/config/rule-validator.js
JoshuaKGoldberg: I believe they are - and added 299d44fd8 with more tests. Is there a case where they aren't?

lib/linter/linter.js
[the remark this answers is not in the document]
mdjermanovic: It works for me locally when I replace this:

```diff
- let ruleOptions = getRuleOptionsInline(ruleId, ruleValue, rule.meta?.defaultOptions);
+ let ruleOptions = Array.isArray(ruleValue) ? ruleValue : [ruleValue];
```

```diff
- if (shouldValidateOptions) {
-    ruleValidator.validate({
-        plugins: config.plugins,
-        rules: {
-            [ruleId]: ruleOptions
-        }
-    });
- }
-
- mergedInlineConfig.rules[ruleId] = ruleOptions;
+ if (shouldValidateOptions) {
+     const configToValidate = {
+         plugins: config.plugins,
+         rules: {
+             [ruleId]: ruleOptions
+         }
+     };
+     ruleValidator.validate(configToValidate);
+     mergedInlineConfig.rules[ruleId] = configToValidate.rules[ruleId]; // this config has default options applied
+ } else {
+     mergedInlineConfig.rules[ruleId] = ruleOptions;
+ }
```

lib/config/rule-validator.js
mdjermanovic: They apply with inline config comments (with the current implementation), but not if the rule is enabled in the config. For example, when I set `schema: false` in the `no-undef` rule, all tests for the rule fail with `Cannot read properties of undefined (reading 'typeof')`.

vladshcherbin:
what a great feature, checking default options on a website or in rules comments is a nightmare. hopefully this will be reviewed and merged soon 👀

lib/linter/linter.js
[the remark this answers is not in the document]
JoshuaKGoldberg: 220b0eb00 -> https://github.com/eslint/eslint/actions/runs/11421164360/job/31777667027?pr=17656:

```plaintext
1) Linter
       when evaluating code with comments to enable rules
         when the rule has default options and a schema
           preserves default options when the comment only has severity:

      AssertionError: expected 'Configuration for rule "with-default-…' to equal 'default-rule-option'
      + expected - actual

      -Configuration for rule "with-default-option" is invalid:
      -	Value [] should NOT have fewer than 1 items.
      +default-rule-option
      
      at Context.<anonymous> (tests/lib/linter/linter.js:1[53](https://github.com/eslint/eslint/actions/runs/11421164360/job/31777667027?pr=17656#step:5:54)3:24)
      at process.processImmediate (node:internal/timers:483:21)
```

I'm pretty sure I'm still somehow missing the point of what you're saying, and I apologize if so. But I've tried staring at this a few times now and just don't get it 😬. What am I missing here?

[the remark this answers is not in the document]
mdjermanovic: The failing test is for the legacy eslintrc mode. It will probably pass with https://github.com/eslint/eslintrc/pull/166.

This code handles flat config mode. The point is, if `ruleOptions` has the default config applied before the following branch, the execution will not fall into it because `ruleOptions.length` would be `> 1`:

https://github.com/eslint/eslint/blob/1a7cdbfad8d3355132aab3201c7dd3f2e879b19d/lib/linter/linter.js#L1826-L1841

and that would cause a bug described in https://github.com/eslint/eslint/pull/17656#discussion_r1788653623.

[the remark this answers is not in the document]
mdjermanovic: To clarify, here are some tests that pass with https://github.com/eslint/eslint/pull/17656/commits/220b0eb00e05429de6e8573360cf73e42e77a616 but fail with https://github.com/eslint/eslint/pull/17656/commits/9c984b3b18bf297fe3212b7fb4e7c14cb1f21c24: https://github.com/mdjermanovic/eslint/commit/ce21aec59875e614fd3edd06fbc171c9c3897ed0

vladshcherbin (commented):
I've used this PR in my project and found some inconsistencies with the defaults from the types. 

Additionally, there are many rules where default options are set to `false` and in `defaultOptions` they are omitted, just a blank `{}`. Is it intentional or what's the case?

Just no name a few:

arrow-body-style - `requireReturnForObjectLiteral`
func-style - `allowArrowFunctions`
no-empty - `allowEmptyCatch`
no-labels - `allowLoop`, `allowSwitch`
no-shadow - `builtinGlobals`, `ignoreOnInitialization`

lib/linter/linter.js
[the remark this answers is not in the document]
JoshuaKGoldberg: Aha! Ok, I think I understand. By moving the `ruleValidator.validate` to _after_ the `ruleOptions.length === 1` check, we're no longer polluting the length check with default values. Thanks! 

I pushed up a commit with these tests.

lib/config/rule-validator.js
JoshuaKGoldberg: Thanks!

lib/rules/dot-notation.js
```diff
@@ -25,6 +25,8 @@ module.exports = {
     meta: {
         type: "suggestion",
 
+        defaultOptions: [{}],
```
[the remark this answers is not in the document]
JoshuaKGoldberg: Good catch 🙂

lib/rules/id-match.js
```diff
@@ -14,6 +14,8 @@ module.exports = {
     meta: {
         type: "suggestion",
 
+        defaultOptions: ["^.+$", {}],
```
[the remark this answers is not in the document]
JoshuaKGoldberg: `^.+$` is a _"match anything"_ pattern. It effectively makes the `pattern` always return true for a match.

lib/rules/no-fallthrough.js
```diff
@@ -90,6 +90,8 @@ module.exports = {
     meta: {
         type: "problem",
 
+        defaultOptions: [{}],
```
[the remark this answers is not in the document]
JoshuaKGoldberg: https://github.com/eslint/eslint/blob/b8fb57256103b908712302ccd508f464eff1c9dc/lib/rules/no-fallthrough.js#L17

Similar to `default-case`. Note the `i` there, if a config passes in a string then the `i` is omitted. This would be a separate (technically breaking, I think) change for the rule to normalize.

JoshuaKGoldberg:
> Additionally, there are many rules where default options are set to `false` and in `defaultOptions` they are omitted, just a blank `{}`. Is it intentional or what's the case?

For the most part, if the default is a falsy value, then I didn't bother adding in a `defaultOptions` entry. Were there any behavioral bugs that you found from this change @vladshcherbin?

I can see an argument of having explicit `defaultOptions` being beneficial for docs generators, even when the type is `boolean`.

JoshuaKGoldberg:
Ok! Per https://github.com/eslint/eslint/pull/17656#discussion_r1807498156, I believe https://github.com/eslint/eslintrc/pull/166 will fix the one last failing unit test.

vladshcherbin:
> For the most part, if the default is a falsy value, then I didn't bother adding in a `defaultOptions` entry. Were there any behavioral bugs that you found from this change @vladshcherbin?
> 
> I can see an argument of having explicit `defaultOptions` being beneficial for docs generators, even when the type is `boolean`.

I'm trying to use `defaultOptions` to diff config rule settings. Imagine one config has rule with just an `'error'` and the other one has `['error', { something: true }]`. With explicit `defaultOptions` I can get full rule settings for both of them and see the which options match and which are different.

lib/rules/default-case.js
```diff
@@ -15,6 +15,8 @@ module.exports = {
     meta: {
         type: "suggestion",
 
+        defaultOptions: [{}],
```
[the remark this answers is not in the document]
vladshcherbin: Well, I'm so confused about `commentPattern`:

- [rule](https://github.com/eslint/eslint/blob/v9.13.0/lib/rules/default-case.js#L7) say it is `/^no default$/iu`
- [website](https://eslint.org/docs/latest/rules/default-case#options) say it is `/^no default$/i`.
- [types](https://github.com/eslint/eslint/blob/v9.13.0/lib/types/rules/best-practices.d.ts#L166) say it is `^no default$`

It must be the same in all 3 of them, no? 😵‍💫

lib/rules/no-fallthrough.js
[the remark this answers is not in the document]
vladshcherbin: I'm confused here too. If I pass another config value (be it a string or other regex) then I expect default value no longer apply.

Isn't it safe to set `defaultOptions.commentPattern` to the value of `DEFAULT_FALLTHROUGH_COMMENT`?

lib/rules/id-match.js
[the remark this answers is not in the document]
vladshcherbin: Oh, I get it. In [current 9.13.0](https://github.com/eslint/eslint/blob/v9.13.0/lib/rules/id-match.js#L61) it's explicitly set inside `create` and in this awesome PR version it's taken directly from `defaultOptions`. Man it made me think what kind of magic sets it there 😅

lib/rules/no-fallthrough.js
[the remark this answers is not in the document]
JoshuaKGoldberg: I tried it out and a unit test failed. So I think that's something that'll have to be treated as a separate issue.

IMO this PR doesn't have to switch _every_ rule to using `meta.defaultOptions`. Some would be a breaking change like this. Others literally can't per the current ESLint test suite, such as `id-length`'s `max` having a non-JSON-serializable `Infinity`.

mdjermanovic:
> Ok! Per [#17656 (comment)](https://github.com/eslint/eslint/pull/17656#discussion_r1807498156), I believe [eslint/eslintrc#166](https://github.com/eslint/eslintrc/pull/166) will fix the one last failing unit test.

You can update package.json to use the branch from eslint/eslintrc#166 and push the change here to run tests in CI. Once we release `@eslint/eslintrc`, we'll update package.json in this PR again before merging it.

```diff
- "@eslint/eslintrc": "^3.1.0",
+ "@eslint/eslintrc": "github:JoshuaKGoldberg/eslintrc#factor-in-default-options",
```

mdjermanovic:
Can you undo changes in rules that have been deprecated in the meantime? We'd rather not change those rules anymore:

array-bracket-spacing
array-element-newline
arrow-parens
arrow-spacing
block-spacing
brace-style
comma-dangle
comma-spacing
comma-style
computed-property-spacing
generator-star-spacing
indent
lines-around-comment
no-multi-spaces

(please doublecheck if I missed some)

mdjermanovic:
> For the most part, if the default is a falsy value, then I didn't bother adding in a `defaultOptions` entry. Were there any behavioral bugs that you found from this change @vladshcherbin?
> 
> I can see an argument of having explicit `defaultOptions` being beneficial for docs generators, even when the type is `boolean`.

I think it makes sense to add `false` defaults. That ensures rules work with boolean values, and seems beneficial for docs generators (i.e., for users reading the docs).

docs/src/extend/custom-rules.md
```diff
@@ -800,6 +802,51 @@ module.exports = {
 
 To learn more about JSON Schema, we recommend looking at some examples on the [JSON Schema website](https://json-schema.org/learn/miscellaneous-examples), or reading the free [Understanding JSON Schema](https://json-schema.org/understanding-json-schema/) ebook.
 
+#### Option Defaults
```
mdjermanovic, line 805:
```suggestion
### Option Defaults
```

I think this shouldn't be under Options Schemas, as default options can be specified without a schema.

docs/src/extend/custom-rules.md
```diff
@@ -800,6 +802,51 @@ module.exports = {
...
+        }],
+        schema: {
+            type: "object",
+            properties: {
+                alias: {
+                    type: "string"
+                }
+            },
+            additionalProperties: false
+        }
```
mdjermanovic, line 826:
```suggestion
        schema: [{
            type: "object",
            properties: {
                alias: {
                    type: "string"
                }
            },
            additionalProperties: false
        }]
```

mdjermanovic (changes requested):

JoshuaKGoldberg:
Odd failure in the browser test, I'm guessing flake:

```plaintext
Execution of 1 workers started at 2024-10-22T15:05:06.247Z

[0-0] RUNNING in chrome - file:///tests/lib/linter/linter.js
[0-0]  Error:  WebDriver Bidi command "browsingContext.navigate" failed with error: unknown error - navigation canceled, as new navigation is requested by reload
[0-0] /home/runner/work/eslint/eslint/node_modules/webdriver/build/index.js:1423
[0-0]       throw new WebDriverRequestError(err, url, opts);
[0-0]             ^
[0-0] 
[0-0] 
[0-0] WebDriverRequestError: WebDriverError: Request failed with error code UND_ERR_SOCKET when running "window" with method "GET"
```

mdjermanovic:
> Odd failure in the browser test, I'm guessing flake

Yeah, I ran it again and it's passing now.

lib/linter/linter.js
```diff
@@ -1022,7 +1036,7 @@ function runRules(
                 Object.create(sharedTraversalContext),
                 {
                     id: ruleId,
-                    options: getRuleOptions(configuredRules[ruleId]),
+                    options: getRuleOptions(configuredRules[ruleId], applyDefaultOptions && rule.meta?.defaultOptions),
```
mdjermanovic, line 1039:
```suggestion
                    options: getRuleOptions(configuredRules[ruleId], applyDefaultOptions ? rule.meta?.defaultOptions : void 0),
```

It's a bit difficult to follow what happens when `applyDefaultOptions` is `false` and then this passes `false` to `getRuleOptions`.

lib/rules/default-case.js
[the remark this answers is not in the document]
mdjermanovic: I agree with @JoshuaKGoldberg that we can't set the default `commentPattern` here (unless it's something like `"^[Nn][Oo] [Dd]...$"`, which would be equivalent I believe but doesn't seem quite useful). We can revisit aligning the website and types with the rule and/or adding an option for case-insensitive matching in a follow up.

tests/lib/linter/linter.js
```diff
@@ -1464,6 +1464,151 @@ describe("Linter", () => {
...
         });
 
+        it("rules use the rule's config when it is present", () => {
+            const config = {
+                rules: {
+                    "no-constant-condition": ["error", { checkLoops: "all" }]
+                }
+            };
+            const codeA = "/*eslint no-constant-condition: error */ while (true) {}";
+            const messages = linter.verify(codeA, config, filename, false);
```
mdjermanovic, line 1474:
```suggestion
            const messages = linter.verify(codeA, config, filename);
```

`Linter#verify` accepts only 3 arguments.

tests/lib/linter/linter.js
```diff
@@ -1464,6 +1464,151 @@ describe("Linter", () => {
...
+                        endColumn: 53
+                    }
+                ]
+            );
+        });
+
+        it("rules should apply meta.defaultOptions when the rule is not configured", () => {
+            const config = { rules: {} };
+            const codeA = "/*eslint no-constant-condition: error */ while (true) {}";
+            const messages = linter.verify(codeA, config, filename, false);
```
mdjermanovic, line 1497:
```suggestion
            const messages = linter.verify(codeA, config, filename);
```

mdjermanovic (commented):
Just one small suggestion, then LGTM.

mdjermanovic (approved):

I'll just convert it to draft so we don't accidentally merge it until `@eslint/eslintrc` is released and package.json is updated here.

package.json
```diff
@@ -101,7 +101,7 @@
     "@eslint-community/regexpp": "^4.11.0",
     "@eslint/config-array": "^0.18.0",
     "@eslint/core": "^0.7.0",
-    "@eslint/eslintrc": "^3.1.0",
+    "@eslint/eslintrc": "github:JoshuaKGoldberg/eslintrc#factor-in-default-options",
```
mdjermanovic, line 104:
```suggestion
    "@eslint/eslintrc": "^3.2.0",
```

Before merging.

lib/config/rule-validator.js
```diff
@@ -187,6 +189,10 @@ class RuleValidator {
                     }`);
                 }
             }
+
+            if (mergedOptions.length) {
+                config.rules[ruleId] = [ruleOptions[0], ...mergedOptions];
```
nzakas, line 194: We shouldn't be changing anything in the config here. This is a validation function so it should just be inspecting values, not overwriting them.

I think it would be better to do this inside of `Config` for flat config. Specifically this function:
https://github.com/eslint/eslint/blob/151c965aec1c46000ac7dfc67a1c04802112aafc/lib/config/config.js#L129

For eslintrc, it's probably better to just do this before calling `ruleValidator.validate()`.

nzakas (changes requested):

lib/config/rule-validator.js
mdjermanovic, line 1: Now that the default options are merged in the `Config` class for configured rules and the `Linter` class for inline config, the changes in this file can be reverted?

lib/linter/linter.js
```diff
@@ -1850,16 +1865,27 @@ class Linter {
...
-                                    });
-                                }
+                                    };
 
-                                mergedInlineConfig.rules[ruleId] = ruleOptions;
+                                    ruleValidator.validate(configToValidate);
+                                    mergedInlineConfig.rules[ruleId] = configToValidate.rules[ruleId];
+                                } else {
+                                    mergedInlineConfig.rules[ruleId] = ruleOptions;
+                                }
```
mdjermanovic, line 1888: I think this change can be reverted as well.

lib/linter/linter.js
```diff
@@ -1850,16 +1865,27 @@ class Linter {
                                     }
                                 }
 
+                                // Apply meta.defaultOptions
+                                const slicedOptions = ruleOptions.slice(1);
+                                const mergedOptions = deepMergeArrays(rule.meta?.defaultOptions, slicedOptions);
+
+                                if (mergedOptions.length) {
+                                    ruleOptions = [ruleOptions[0], ...mergedOptions];
+                                }
```
mdjermanovic, line 1874: This could be wrapped in `else` of the above `if`, since in the `if` branch we're using options from the config which already have default options applies?

lib/config/config.js
```diff
@@ -276,6 +256,37 @@ class Config {
...
+    #normalizeRulesConfig() {
+        for (const [ruleId, originalConfig] of Object.entries(this.rules)) {
+
+            // ensure rule config is an array
+            let ruleConfig = Array.isArray(originalConfig)
+                ? originalConfig
+                : [originalConfig];
+
+            // normalize severity
+            ruleConfig[0] = severities.get(this.rules[ruleId][0]);
```
mdjermanovic, line 275:
```suggestion
            ruleConfig[0] = severities.get(ruleConfig[0]);
```

mdjermanovic (dismissed):

I'll just convert it to draft to prevent accidental merging until we commit https://github.com/eslint/eslint/pull/17656#discussion_r1820984884.

nzakas (dismissed):
LGTM. Thanks!

mdjermanovic (approved):

PR #17657, opened 2023-10-17
docs: explained rule fixers and suggestions

JoshuaKGoldberg:
#### Prerequisites checklist

[x] Documentation update
[ ] Add autofix to a rule
[ ] Add a CLI option
[ ] Add something to the core
[ ] Other, please explain:

Per the issue, added _Report Fixers_ and _Report Suggestions_ as sections in the _Core Concepts_ page. Also switched the _Rules_ page link about suggestions to point to the new core conception subheading instead of the custom rules docs.

Fixes https://github.com/eslint/eslint.org/issues/479.

I personally call them rule "reports" as that's what the `context.report` API refers to them as. But nowhere else in the _Core Concepts_ page are they referred to by that name. What's the standard nomenclature we should use here?

### DIFF
--- a/docs/src/use/core-concepts.md
+++ b/docs/src/use/core-concepts.md
@@ -23,17 +23,20 @@ ESLint contains hundreds of built-in rules that you can use. You can also create
 
 For more information, refer to [Rules](../rules/).
 
-### Report Fixers
+### Rule Fixes
 
-Rule reports may optionally provide "fixes" that safely correct the reported issue without changing application logic.
+Rules may optionally provide fixes for violations that they find. Fixes safely correct the violation without changing application logic.
+
 Fixes may be applied automatically with the [`--fix` command line option](https://eslint.org/docs/latest/use/command-line-interface#--fix) and via editor extensions.
 
-Rules that may provide fixers are marked with 🔧 in [Rules](../rules/).
+Rules that may provide fixes are marked with 🔧 in [Rules](../rules/).
 
-### Report Suggestions
+### Rule Suggestions
 
-Rule reports may optionally provide "suggestions" that correct the reported issue but also change application logic.
-Suggestions cannot be applied automatically, but may be applied manually via editor extensions.
+Rules may optionally provide suggestions in addition to or instead of providing fixes. Suggestions differ from fixes in two ways:
+
+1. Suggestions may change application logic and so cannot be automatically applied.
+1. Suggestions cannot be applied through the ESLint CLI and are only available through editor integrations.
 
 Rules that may provide suggestions are marked with 💡 in [Rules](../rules/).
 
--- a/docs/src/use/core-concepts.md
+++ b/docs/src/use/core-concepts.md
@@ -23,6 +23,20 @@ ESLint contains hundreds of built-in rules that you can use. You can also create
 
 For more information, refer to [Rules](../rules/).
 
+### Report Fixers
+
+Rule reports may optionally provide "fixes" that safely correct the reported issue without changing application logic.
+Fixes may be applied automatically with the [`--fix` command line option](https://eslint.org/docs/latest/use/command-line-interface#--fix) and via editor extensions.
+
+Rules that may provide fixers are marked with 🔧 in [Rules](../rules/).
+
+### Report Suggestions
+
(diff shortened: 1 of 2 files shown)


docs/src/use/core-concepts.md
```diff
@@ -23,6 +23,20 @@ ESLint contains hundreds of built-in rules that you can use. You can also create
 
 For more information, refer to [Rules](../rules/).
 
+### Report Fixers
+
+Rule reports may optionally provide "fixes" that safely correct the reported issue without changing application logic.
+Fixes may be applied automatically with the [`--fix` command line option](https://eslint.org/docs/latest/use/command-line-interface#--fix) and via editor extensions.
+
+Rules that may provide fixers are marked with 🔧 in [Rules](../rules/).
```
nzakas, line 31:
```suggestion
### Rule Fixes

Rules may optionally provide fixes for violations that they find. Fixes safely correct the violation without changing application logic.

Fixes may be applied automatically with the [`--fix` command line option](https://eslint.org/docs/latest/use/command-line-interface#--fix) and via editor extensions.

Rules that may provide fixes are marked with 🔧 in [Rules](../rules/).
```

docs/src/use/core-concepts.md
```diff
@@ -23,6 +23,20 @@ ESLint contains hundreds of built-in rules that you can use. You can also create
...
+
+### Report Suggestions
+
+Rule reports may optionally provide "suggestions" that correct the reported issue but also change application logic.
+Suggestions cannot be applied automatically, but may be applied manually via editor extensions.
+
+Rules that may provide suggestions are marked with 💡 in [Rules](../rules/).
```
nzakas, line 38:
```suggestion
### Rule Suggestions

Rules may optionally provide suggestions in addition to or instead of providing fixes. Suggestions differ from fixes in two ways:

1. Suggestions may change application logic and so cannot be automatically applied.
1. Suggestions cannot be applied through the ESLint CLI and are only available through editor integrations.

Rules that may provide suggestions are marked with 💡 in [Rules](../rules/).
```

nzakas (changes requested):
Thanks for putting this together! I did a little bit of wording cleanup.

I've started to use the term "violations" to describe lint errors -- we have too many other words floating around in the docs right now, so definitely room to improve.

docs/src/use/core-concepts.md
```diff
@@ -23,6 +23,23 @@ ESLint contains hundreds of built-in rules that you can use. You can also create
 
 For more information, refer to [Rules](../rules/).
 
+### Rule Fixes
+
+Rules may optionally provide fixes for violations that they find. Fixes safely correct the violation without changing application logic.
```
JoshuaKGoldberg, line 28: https://github.com/eslint/eslint/pull/17657#pullrequestreview-1682571543:

> I've started to use the term "violations" to describe lint errors -- we have too many other words floating around in the docs right now, so definitely room to improve.

Interesting! Are there any other discussion places that come to mind for the consolidation? I can file an issue if not. IMO it'd be valuable to have shared vocab used as a standard in the docs. Even a glossary long-term.

In case it's useful: "violation" is not what I personally would lean towards right now. It's a bit more intense and opinionated of a word than "report".

nzakas: Nothing official, no. Just something I've been thinking about. A glossary is definitely of interest to me.

Our problem is that we already have clearly defined meanings for several terms that we could use:

1. problem - Type of rule
2. suggestion - Actually both a type of rule and an unsafe fix (need to disambiguate this in the future)
3. error - Too ambiguous vs. runtime errors and also a severity level
4. warning - Severity level

We've actually used "report" historically to represent the collection of violations for a given run, so I'm not in favor of repurposing that.

Thinking in terms of rule violations make sense to me, which is why I started adopting "violation" as the term. Definitely open to other suggestions.

mdjermanovic (approved):
LGTM, thanks! Leaving open for @nzakas to verify his changes.

nzakas (approved):
LGTM. Thanks!

docs/src/use/core-concepts.md
JoshuaKGoldberg: Realized I never back-linked: https://github.com/eslint/eslint/issues/17987 tracks adding a glossary.

Issue #17568, opened 2023-09-15
Rule Change: Add option to `no-unused-vars` that will report a variable that has been marked as ignored but is being used.


Pearce-Ropion:
### What rule do you want to change?

no-unused-vars

### What change to do you want to make?

Generate more warnings

### How do you think the change should be implemented?

A new option

### Example code

```js
/* eslint no-unused-vars: ["error", { "varsIgnorePattern": "^_" }] */

const _myUnusedVariable = 10;

export const myUsedVariable = _myUnusedVariable + 10;
```

### What does the rule currently do for this code?

Currently, the rule will not report this case because `_myUnusedVariable` is being used.

### What will the rule do after it's changed?

The rule would report that a variable that has been marked as an "ignored unused variable" is actually being used. 

I propose a new option, potentially called `reportUsedIgnorePattern`. Optionally, there could be a version of this option for each of the different ignore pattern options.

### Participation

- [X] I am willing to submit a pull request to implement this change.

### Additional comments

The main reason for this is that I've found that devs will name a variable with our ignore pattern since its a simple way to have another variable with a very similar name. I would like to instead encourage devs to rename their variables with a clear meaning by not allowing them to use the ignore pattern for a used variable.

I acknowledge that part of this is that we have a very simple ignore pattern, but I feel like this is still something that should be enforceable by the lint rule.

Rec0iL99:
Hi @Pearce-Ropion, thanks for the issue. Can you please provide a few more usage examples to support this proposal?

> Currently, the rule will not report this case because _myUnusedVariable is being used.

The rule will not check `_myUnusedVariable` for usage since it matches the ignore pattern ([docs](https://eslint.org/docs/latest/rules/no-unused-vars#varsignorepattern)).

Rec0iL99:
No response so closing.

Pearce-Ropion:
@Rec0iL99 Apologies, I had written out but I guess I never sent it. TIL GitHub caches draft comments a lot longer than expected. 

The idea, is that if an ignore pattern is set, users should only be able to use that variable naming pattern for ignored unused variables. 

This increases code readability as a quick scan of the code would indicate to the user that a variable that is marked with an ignore pattern is unused. 

In the current implementation, users are able to use variables that are marked with by the unused variable ignore pattern which decreases readability as you wonder why the variable was marked as "ignore-unused" given that it is actually being used. 

I think my original example snippet makes the most sense but I can see how further explanation would be helpful. 

Given
```js
[...]

PR #17662, opened 2023-10-18
feat: Add `reportUsedIgnorePattern` option to `no-unused-vars` rule

Pearce-Ropion:
#### Prerequisites checklist

- [x] Documentation update
- [ ] Bug fix ([template](https://raw.githubusercontent.com/eslint/eslint/HEAD/templates/bug-report.md))
- [x] Changes an existing rule ([template](https://raw.githubusercontent.com/eslint/eslint/HEAD/templates/rule-change-proposal.md))
- [ ] Add autofix to a rule
- [ ] Add a CLI option
- [ ] Other, please explain:

`no-unused-vars`

- [x] Generate more warnings
- [ ] Generate fewer warnings
- [ ] Implement autofix
- [ ] Implement suggestions

- [x] A new option
- [ ] A new default behavior
- [ ] Other

Example 1:
```js
/* eslint no-unused-vars: ["error", { "varsIgnorePattern": "^_" }] */

const _a = 5;
const _b = _a + 5
```

Example 2:
```js
/* eslint no-unused-vars: ["error", { "argsIgnorePattern": "^_" }] */

function foo(_a) {
  return _a + 5
}
```

Example 3:
```js
/* eslint no-unused-vars: ["error", { "destructuredArrayIgnorePattern": "^_" }] */

const [ a, _b ] = items;
console.log(a + _b);
```

Example 4:
```js
/* eslint no-unused-vars: ["error", { "caughtErrorsIgnorePattern": "^_" }] */

try {}
catch(_err) {
  console.error(_err);
}
```

In all examples, no warnings are reported.

When the new option `reportUsedIgnorePattern` is enabled, each of the examples above will report; indicating that a variable that has been marked as an ignored unused variable is being used.

This change adds a new option `reportUsedIgnorePattern` to the `no-unused-vars` rule. When enabled, the rule will report any instances of variables that match any of the valid ignore pattern options when that variable is actually being used. 

Resolves: #17568

When finding all references in order to report on the correct nodes, it is also finding the variable declaration. So a question for the team would be which nodes should this option report on:
* Should all references including the declaration be reported?
* Should only the variable declaration be reported? 
* Should only references not including the declaration be reported

### DIFF
--- a/lib/rules/no-unused-vars.js
+++ b/lib/rules/no-unused-vars.js
@@ -228,52 +228,51 @@ module.exports = {
         /**
          * Generate the warning message about a variable being used even though
          * it is marked as being ignored.
-         * @param {Variable} usedIgnoredVar eslint-scope variable object
+         * @param {Variable} variable eslint-scope variable object
+         * @param {"catch-clause" | "args" | "array-destructure" | "vars"} variableType
+         *  the type of the unused variable
+         * @throws {Error} (Unreachable)
          * @returns {UsedIgnoredVarMessageData} The message data to be used with
          * this used ignored variable.
          */
-        function getUsedIgnoredMessageData(usedIgnoredVar) {
-            const def = usedIgnoredVar.defs && usedIgnoredVar.defs[0];
-            let additional = "";
+        function getUsedIgnoredMessageData(variable, variableType) {
+            let pattern;
+            let variableDescription;
 
-            if (def) {
-                let type;
-                let pattern;
+            switch (variableType) {
+                case "catch-clause":
+                    pattern = config.caughtErrorsIgnorePattern;
+                    variableDescription = "args";
+                    break;
 
-                switch (def.type) {
-                    case "CatchClause":
-                        if (config.caughtErrorsIgnorePattern) {
-                            type = "args";
-                            pattern = config.caughtErrorsIgnorePattern;
-                        }
-                        break;
+                case "args":
+                    pattern = config.argsIgnorePattern;
+                    variableDescription = "args";
+                    break;
 
-                    case "Parameter":
-                        if (config.argsIgnorePattern) {
-                            type = "args";
-                            pattern = config.argsIgnorePattern;
-                        }
-                        break;
+                case "array-destructure":
+                    pattern = config.destructuredArrayIgnorePattern;
+                    variableDescription = "elements of array destructuring";
+                    break;
 
-                    default:
-                        if (def.name.parent.type === "ArrayPattern" && config.destructuredArrayIgnorePattern) {
-                            type = "elements of array destructuring";
-                            pattern = config.destructuredArrayIgnorePattern;
-                        } else if (config.varsIgnorePattern) {
-                            type = "vars";
-                            pattern = config.varsIgnorePattern;
-                        }
-                        break;
-                }
+                case "vars":
+                    pattern = config.varsIgnorePattern;
+                    variableDescription = "vars";
+                    break;
 
-                if (type) {
-                    additional = `. Used ${type} must not match ${pattern.toString()}`;
-                }
+                default:
+                    throw new Error(`Unexpected variable type: ${variableType}`);
+            }
+
+            let additionalMessageData = "";
+
+            if (pattern && variableDescription) {
+                additionalMessageData = `. Used ${variableDescription} must not match ${pattern.toString()}`;
             }
 
             return {
-                varName: usedIgnoredVar.name,
-                additional
+                varName: variable.name,
+                additional: additionalMessageData
             };
         }
 
@@ -671,13 +670,12 @@ module.exports = {
          * Gets an array of variables without read references.
          * @param {Scope} scope an eslint-scope Scope object.
          * @param {Variable[]} unusedVars an array containing unused variables that should be reported
-         * @param {Variable[]} usedIgnoredVars an array containing used variables that match an ignore pattern
          * @returns {[Variable[], Variable[]]} an array containing
          *   1. unused variables of the scope and descendant scopes.
          *   2. used variables that match an ignore pattern of the scope and descendant scope
          * @private
          */
-        function collectUnusedVariables(scope, unusedVars, usedIgnoredVars) {
+        function collectUnusedVariables(scope, unusedVars) {
             const variables = scope.variables;
             const childScopes = scope.childScopes;
             let i, l;
@@ -723,7 +721,11 @@ module.exports = {
                             config.destructuredArrayIgnorePattern.test(def.name.name)
                         ) {
                             if (config.reportUsedIgnorePattern && isUsedVariable(variable)) {
-                                usedIgnoredVars.push(variable);
+                                context.report({
+                                    node: def.name,
+                                    messageId: "usedIgnoredVar",
+                                    data: getUsedIgnoredMessageData(variable, "array-destructure")
+                                });
                             }
 
                             continue;
@@ -746,7 +748,11 @@ module.exports = {
                             // skip ignored parameters
                             if (config.caughtErrorsIgnorePattern && config.caughtErrorsIgnorePattern.test(def.name.name)) {
                                 if (config.reportUsedIgnorePattern && isUsedVariable(variable)) {
-                                    usedIgnoredVars.push(variable);
+                                    context.report({
+                                        node: def.name,
+                                        messageId: "usedIgnoredVar",
+                                        data: getUsedIgnoredMessageData(variable, "catch-clause")
+                                    });
                                 }
 
                                 continue;
@@ -766,7 +772,11 @@ module.exports = {
                             // skip ignored parameters
                             if (config.argsIgnorePattern && config.argsIgnorePattern.test(def.name.name)) {
                                 if (config.reportUsedIgnorePattern && isUsedVariable(variable)) {
-                                    usedIgnoredVars.push(variable);
+                                    context.report({
+                                        node: def.name,
+                                        messageId: "usedIgnoredVar",
+                                        data: getUsedIgnoredMessageData(variable, "args")
+                                    });
                                 }
 
                                 continue;
@@ -781,7 +791,11 @@ module.exports = {
                             // skip ignored variables
                             if (config.varsIgnorePattern && config.varsIgnorePattern.test(def.name.name)) {
                                 if (config.reportUsedIgnorePattern && isUsedVariable(variable)) {
-                                    usedIgnoredVars.push(variable);
+                                    context.report({
+                                        node: def.name,
+                                        messageId: "usedIgnoredVar",
+                                        data: getUsedIgnoredMessageData(variable, "vars")
+                                    });
                                 }
 
                                 continue;
@@ -796,10 +810,10 @@ module.exports = {
             }
 
             for (i = 0, l = childScopes.length; i < l; ++i) {
-                collectUnusedVariables(childScopes[i], unusedVars, usedIgnoredVars);
+                collectUnusedVariables(childScopes[i], unusedVars);
             }
 
-            return [unusedVars, usedIgnoredVars];
+            return unusedVars;
         }
 
         //--------------------------------------------------------------------------
@@ -808,7 +822,7 @@ module.exports = {
 
         return {
             "Program:exit"(programNode) {
-                const [unusedVars, usedIgnoredVars] = collectUnusedVariables(sourceCode.getScope(programNode), [], []);
+                const unusedVars = collectUnusedVariables(sourceCode.getScope(programNode), []);
 
                 for (let i = 0, l = unusedVars.length; i < l; ++i) {
                     const unusedVar = unusedVars[i];
@@ -843,18 +857,6 @@ module.exports = {
                             messageId: "unusedVar",
                             data: getDefinedMessageData(unusedVar)
                         });
-                    }
-                }
-
-                if (config.reportUsedIgnorePattern) {
-                    for (const usedIgnoredVar of usedIgnoredVars) {
-                        for (const definition of usedIgnoredVar.defs) {
-                            context.report({
-                                node: definition.name,
-                                messageId: "usedIgnoredVar",
-                                data: getUsedIgnoredMessageData(usedIgnoredVar)
-                            });
-                        }
                     }
                 }
             }
--- a/lib/rules/no-unused-vars.js
+++ b/lib/rules/no-unused-vars.js
@@ -670,9 +670,7 @@ module.exports = {
          * Gets an array of variables without read references.
          * @param {Scope} scope an eslint-scope Scope object.
          * @param {Variable[]} unusedVars an array containing unused variables that should be reported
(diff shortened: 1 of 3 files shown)


lib/rules/no-unused-vars.js
```diff
@@ -178,6 +191,47 @@ module.exports = {
...
+                } else if (defType === "Parameter" && config.argsIgnorePattern) {
+                    type = "args";
+                    pattern = config.argsIgnorePattern.toString();
+                } else if (defType !== "Parameter" && config.varsIgnorePattern) {
+                    type = "vars";
+                    pattern = config.varsIgnorePattern.toString();
+                } else if (def.name.parent.type === "ArrayPattern" && config.destructuredArrayIgnorePattern) {
+                    type = "elements of array destructuring";
+                    pattern = config.destructuredArrayIgnorePattern.toString();
+                }
```
costinsin, line 222:
```suggestion
                switch (true) {
                    case defType === "CatchClause" && config.caughtErrorsIgnorePattern:
                        type = "args";
                        pattern = config.caughtErrorsIgnorePattern.toString();
                        break;
                    case defType === "Parameter" && config.argsIgnorePattern:
                        type = "args";
                        pattern = config.argsIgnorePattern.toString();
                        break;
                    case defType !== "Parameter" && config.varsIgnorePattern:
                        type = "vars";
                        pattern = config.varsIgnorePattern.toString();
                    case def.name.parent.type === "ArrayPattern" && config.destructuredArrayIgnorePattern:
                        type = "elements of array destructuring";
                        pattern = config.destructuredArrayIgnorePattern.toString();
                        break;
                }
```

Pearce-Ropion: This seems like an anti pattern.

Also, there are 2 other functions that are very similar to this one in this rule already so making this change would be inconsistent with how the rest of the rule is formatted.

costinsin: This is clearly not an anti pattern, as TypeScript 5.3 [just added](https://devblogs.microsoft.com/typescript/announcing-typescript-5-3-beta/#switch-true-narrowing#switch-true-narrowing) type narrowing for this pattern. I feel it resembles the powerful `match` present usually in functional languages.

This being said, I agree with you that we should keep the same code style and not mix them. It was just a suggestion.

mdjermanovic:
> * Should all references including the declaration be reported?
> * Should only the variable declaration be reported?
> * Should only references not including the declaration be reported

Would it make sense to report only on the reasons the variable is considered used, which are usually its read references?

This is how it currently works:

```js
/* eslint no-unused-vars: ["error", {
    varsIgnorePattern: "^_",
    reportUsedIgnorePattern: true
}] */

var _a; // not reported

_a = 1; // this is reported but doesn't seem relevant?

foo(_a); // this is also reported, ok

```

But this variable is considered used only because of the read reference so it might make sense to report like this:

```js
/* eslint no-unused-vars: ["error", {
    varsIgnorePattern: "^_",
    reportUsedIgnorePattern: true
}] */

var _a; // not reported

_a = 1; // not reported

foo(_a); // reported

```

lib/rules/no-unused-vars.js
```diff
@@ -178,6 +191,47 @@ module.exports = {
...
+
+                if (type) {
+                    additional = `. Allowed unused ${type} must not match ${pattern}`;
```
mdjermanovic, line 225:
```suggestion
                    additional = `. Used ${type} must not match ${pattern}`;
```

Pearce-Ropion:
> > * Should all references including the declaration be reported?
> > * Should only the variable declaration be reported?
> > * Should only references not including the declaration be reported
> 
> Would it make sense to report only on the reasons the variable is considered used, which are usually its read references?
> 
> This is how it currently works:
> 
> ```js
> /* eslint no-unused-vars: ["error", {
>     varsIgnorePattern: "^_",
>     reportUsedIgnorePattern: true
> }] */
> 
> var _a; // not reported
> 
> _a = 1; // this is reported but doesn't seem relevant?
> 
> foo(_a); // this is also reported, ok
> ```
> 
> But this variable is considered used only because of the read reference so it might make sense to report like this:
> 
> ```js
> /* eslint no-unused-vars: ["error", {
>     varsIgnorePattern: "^_",
>     reportUsedIgnorePattern: true
> }] */
> 
> var _a; // not reported
> 
> _a = 1; // not reported
> 
> foo(_a); // reported
> ```

That makes sense to me. However, I would also expect that a user shouldn't be allowed to write to an ignored unused variable since it should be unused.

lib/rules/no-unused-vars.js
```diff
@@ -711,8 +789,25 @@ module.exports = {
                         });
                     }
                 }
+
+                if (config.reportUsedIgnorePattern) {
+                    for (const usedIgnoredVar of usedIgnoredVars) {
+                        if (usedIgnoredVar.defs.length > 0) {
+                            const references = usedIgnoredVar.references.filter(
+                                ref => ref.isRead() && ref.from.variableScope === usedIgnoredVar.scope.variableScope
```
mdjermanovic, line 797: Per this condition, `_a` wouldn't be reported at all in the following example, which seems wrong as it's used?

```js
/* eslint no-unused-vars: ["error", {
    varsIgnorePattern: "^_",
    reportUsedIgnorePattern: true
}]*/

var _a = 42;

foo(() => _a);

```

mdjermanovic (changes requested):

lib/rules/no-unused-vars.js
Pearce-Ropion: Do you think I should revert to just `ref => ref.from.variableScope === usedIgnoredVar.scope.variableScope`? Is there a better condition to use?

Pearce-Ropion:
I changed it to skip all variable initializations but otherwise report on all references. I also doesn't check whether the variable scope matches anymore so it can handle variables defined in the outer scope which are used in an inner scope.

Pearce-Ropion:
@mdjermanovic I think this is working as expected now. Can you think of any other cases that I could test for? Otherwise, I think we are good to go.

mdjermanovic:
@Pearce-Ropion apologies for the delay in reviewing this PR.

> I changed it to skip all variable initializations but otherwise report on all references.

I'm not sure what the common use cases for the `varsIgnorePattern` option are, but if user has a code like the following where the rule doesn't report any errors:

```js
/*eslint no-unused-vars: ["error", { varsIgnorePattern: "^_", reportUsedIgnorePattern: true }]*/

var _foo;

_foo = 5;

```

and then they add more code that causes the variable to be considered used, so that the rule now reports it:

```js
/*eslint no-unused-vars: ["error", { varsIgnorePattern: "^_", reportUsedIgnorePattern: true }]*/

var _foo;

_foo = 5;

bar(_foo);

```

then, I think it makes the most sense to report only on the references where this rule considers the variable to be used: `_foo` in `bar(_foo)`, because `_foo = 5` has no effect on whether the variable is considered used, and it was valid before `bar(_foo)` was added.

Alternatively, we could report only the declaration assuming that the variable name is wrong because it was intended to be used (as opposed to reporting references, in which case the assumption is that the variable that wasn't intended to be used is mistakenly used).

Curious what others think about this.

Pearce-Ropion:
No worries, eslint is a massive project and I don't want to burden anyone. 

> then, I think it makes the most sense to report only on the references where this rule considers the variable to be used: `_foo` in `bar(_foo)`, because `_foo = 5` has no effect on whether the variable is considered used, and it was valid before `bar(_foo)` was added.
 
> Alternatively, we could report only the declaration assuming that the variable name is wrong because it was intended to be used (as opposed to reporting references, in which case the assumption is that the variable that wasn't intended to be used is mistakenly used).

Using vscode's eslint extension as an example, on one hand it would be useful to see reported issues on the usages since that's what is considered invalid according to the new option. However, It would be more clear what the problem is and how to fix it at the declaration, so it'd probably be better to do it there.

nzakas:
> then, I think it makes the most sense to report only on the references where this rule considers the variable to be used: _foo in bar(_foo), because _foo = 5 has no effect on whether the variable is considered used, and it was valid before bar(_foo) was added.

I think this makes the most sense. :+1:

@Pearce-Ropion sorry for the delay, we've been busy working on v9. Can you please rebase and make any further changes? Thanks!

nzakas:
Oops, accidentally hit "close"

Pearce-Ropion:
> Oops, accidentally hit "close"

lol, I was going to reopen once I pushed the changes.

Pearce-Ropion:
This is still active. Waiting for a review.

lib/rules/no-unused-vars.js
```diff
@@ -178,6 +191,47 @@ module.exports = {
...
+        function getUsedIgnoredMessageData(usedIgnoredVar) {
+            const def = usedIgnoredVar.defs && usedIgnoredVar.defs[0];
+            let additional = "";
+
+            if (def) {
+                const defType = def.type;
+                let type;
+                let pattern;
+
+                if (defType === "CatchClause" && config.caughtErrorsIgnorePattern) {
```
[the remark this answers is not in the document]
Pearce-Ropion: As per https://github.com/eslint/eslint/pull/17662#discussion_r1365318959 we opted to leave the if statements as is to reduce the scope of the PR and to maintain consistency with other similar functions within the rule. If I make this change, should I also update the other functions that use a similar flow?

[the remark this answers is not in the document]
Pearce-Ropion: On second thought, given that this is the second time this suggestion has been made, I've gone ahead and refactored this and the other message getters in the rule for consistency to use `switch` statements.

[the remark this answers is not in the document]
nzakas: Please double-check who the person is who is leaving suggestions on a PR. Not everyone is a team member. :)

lib/rules/no-unused-vars.js
```diff
@@ -144,22 +144,40 @@ module.exports = {
...
+                            pattern = config.argsIgnorePattern;
+                        }
+                        break;
+
+                    default:
+                        if (config.varsIgnorePattern) {
+                            type = "vars";
+                            pattern = config.varsIgnorePattern;
+                        }
+                        break;
```
nzakas, line 174: Nice :+1:

nzakas (approved):
LGTM. Would like @mdjermanovic to re-review before merging.

mdjermanovic:
In this case, the message seems misleading:

```js
/*eslint no-unused-vars: ["error", {
    "reportUsedIgnorePattern": true,
    "destructuredArrayIgnorePattern": "^_",
    "varsIgnorePattern": "[iI]gnored"
}]*/

let _x; // '_x' is marked as ignored but is used. Used vars must not match /[iI]gnored/u.

[_x] = arr;

foo(_x);

```

Tanujkanti4441:
Hi @Pearce-Ropion, some changes are requested. you can check out.

tests/lib/rules/no-unused-vars.js
```diff
@@ -442,6 +460,22 @@ ruleTester.run("no-unused-vars", rule, {
         {
             code: "var a; a ??= 1;",
             languageOptions: { ecmaVersion: 2021 }
+        },
+
+        // https://github.com/eslint/eslint/issues/17568
+        {
+            code: "const a = 5; const _c = a + 5;",
+            options: [{ args: "all", varsIgnorePattern: "^_", reportUsedIgnorePattern: true }],
```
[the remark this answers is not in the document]
Pearce-Ropion: Yep, nice catch. I also apparently got the error output correct on the invalid cases.

lib/rules/no-unused-vars.js
```diff
@@ -92,7 +103,8 @@ module.exports = {
             vars: "all",
             args: "after-used",
             ignoreRestSiblings: false,
-            caughtErrors: "all"
+            caughtErrors: "all",
+            reportUsedIgnorePattern: false,
```
mdjermanovic, line 107:
```suggestion
            reportUsedIgnorePattern: false
```

To fix the lint error.

mdjermanovic:
> In this case, the message seems misleading:
> 
> ```js
> /*eslint no-unused-vars: ["error", {
>     "reportUsedIgnorePattern": true,
>     "destructuredArrayIgnorePattern": "^_",
>     "varsIgnorePattern": "[iI]gnored"
> }]*/
> 
> let _x; // '_x' is marked as ignored but is used. Used vars must not match /[iI]gnored/u.
> 
> [_x] = arr;
> 
> foo(_x);
> ```

I'm still getting the same error, looks like this case still hasn't been addressed?

mdjermanovic:
Here's a similar case that should also be fixed:

```js
/*eslint no-unused-vars: ["error", {
    "reportUsedIgnorePattern": true,
    "destructuredArrayIgnorePattern": "^_",
    "varsIgnorePattern": "[iI]gnored"
}]*/

const [ignored] = arr; // 'ignored' is marked as ignored but is used. Used elements of array destructuring must not match /^_/u.

foo(ignored);

```

The message should be: `'ignored' is marked as ignored but is used. Used vars must not match /[iI]gnored/u.`

lib/rules/no-unused-vars.js
```diff
@@ -610,6 +717,10 @@ module.exports = {
                             config.destructuredArrayIgnorePattern &&
                             config.destructuredArrayIgnorePattern.test(def.name.name)
                         ) {
+                            if (config.reportUsedIgnorePattern && isUsedVariable(variable)) {
+                                usedIgnoredVars.push(variable);
+                            }
```
mdjermanovic, line 722: Instead of collecting these vars and later figuring out why they were considered as used ignored, we could just report them right away with the data we have at this point:

```suggestion
                            if (config.reportUsedIgnorePattern && isUsedVariable(variable)) {
                                context.report({
                                    node: def.name,
                                    messageId: "usedIgnoredVar",
                                    data: {
                                        varName: variable.name,
                                        additional: `. Used elements of array destructuring must not match ${config.destructuredArrayIgnorePattern}`
                                    }
                                });
                            }
```

(the same for catch arguments, params and variables).

That would fix https://github.com/eslint/eslint/pull/17662#issuecomment-1910344522 and https://github.com/eslint/eslint/pull/17662#issuecomment-1959062359.

nzakas:
@Pearce-Ropion are you still working on this?

Pearce-Ropion:
> @Pearce-Ropion are you still working on this?

Yes, apologies. I've been slammed with work. I'll get the proposed suggestions in this weekend

Pearce-Ropion:
Ok, I've applied @mdjermanovic's suggestions to report as soon as we detect the used ignored variable. I also explicitly added the test cases that @mdjermanovic used.

I've also updated the message data handling. I had already updated the other message data handlers as per previous comments, but I noticed that all 3 of the message data getters were doing basically the same thing. So I've added a `getVariableDescription` helper function to get the required info for the message data base on the type of the variable that is currently being reported.

nzakas (approved):

mdjermanovic (approved):

PR #17690, opened 2023-10-28
feat: expose cli entrance

zanminkian:
#### Prerequisites checklist

[ ] Documentation update
[ ] Add autofix to a rule
[ ] Add a CLI option
[ ] Add something to the core
[x] Other, please explain:

Expose cli entrance. I have a cli tool want to reuse the eslint's cli entrance by using:
```ts
// ./bin/my-lint.js

import('eslint/bin')
```

See the similar [PR](https://github.com/lint-staged/lint-staged/pull/1237) for `lint-staged`.

No.

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,6 +9,7 @@
   "main": "./lib/api.js",
   "exports": {
     "./package.json": "./package.json",
+    "./bin": "./bin/eslint.js",
     ".": "./lib/api.js",
     "./use-at-your-own-risk": "./lib/unsupported-api.js"
   },


nzakas:
I'm not sure I understand why this would be helpful. The CLI entrance doesn't export anything, so you'd just be running the code without any arguments, which would fail. What is it you're trying to achieve that can't be done via `exec("npx eslint .")`?

zanminkian:
> The CLI entrance doesn't export anything, so you'd just be running the code without any arguments, which would fail. 

Other tools based on eslint can easily hook the cli.

```js
// ./bin/my-lint.js
process.argv.push('--fix') // always fix

import('eslint/bin')
```

```js
my-lint .
```

> What is it you're trying to achieve that can't be done via `exec("npx eslint .")`?

I have an npm package (a cli tool) based on eslint. If I use `exec("npx eslint .")` in my npm package, users who use my tool via `npx my-lint` will be asked to install `eslint` again, which is not friendly. If I use `process.argv.push('--fix'); import('eslint/bin')`, user will not be asked to install `eslint`.

nzakas:
> If I use exec("npx eslint .") in my npm package, users who use my tool via npx my-lint will be asked to install eslint again, which is not friendly.

I don't understand this. If your tool is based on ESLint, then isn't ESLint already installed? Why would they be asked to install it again?

nzakas:
And note: You'll need to sign the CLA for us to consider this PR.

zanminkian:
>  If your tool is based on ESLint, then isn't ESLint already installed? Why would they be asked to install it again?

This will appear when user do not install `my-lint` into package.json and just want to run `npx my-lint` for a quick test, because of `exec("npx eslint .")` in `my-lint`

nzakas:
I see. And what is the reason you don't use the `ESLint` class?

zanminkian:
> And what is the reason you don't use the `ESLint` class?

Because it's hard to use, and I have to rewrite the same code as `./bin/eslint` of eslint. Let's say I have a cli tool to run `prettier` and `eslint`, and this tool have the same cli interface with `prettier` and `eslint`.

```ts
// ./bin/my-lint.js
const origin = process.argv
process.argv = origin.filter(i=>!i.startsWith('--eslint')).map(i=>i.replace('--prettier', '')))
await import('prettier/bin')
process.argv = origin.filter(i=>!i.startsWith('--prettier')).map(i=>i.replace('--eslint', '')))
await import('eslint/bin')
```

```sh
my-lint . --eslint-f --prettier-w
```
In this example, user don't need to learn new api. And I, who created it, don't need to rewrite the same code as `./bin/eslint.js` of eslint.

nzakas:
I see, thanks for explaining.

I guess I'm still having a hard time understanding this use case. If your tool is bundling ESLint as a dependency, then you should be able to run it via `exec()` directly. It seems like using `import()` is jumping through a lot of unnecessary hoops. What am I missing?

zanminkian:
Hi.
1. Using `exec()` is unfriendly to user. They have to install eslint again. I think I've explain [here](https://github.com/eslint/eslint/pull/17690#issuecomment-1787779751). My tool is `git-validator`. You can 
   1. create a empty folder 
   2. and then run `npm init -y` to create a new `package.json`
   3. and then run `npx git-validator`. You will find that you need to install eslint again. See this (line)[https://github.com/zanminkian/git-validator/blob/f9a29909bfb5bc451de5f34a6169dc065b0f9af6/packages/git-validator/src/cli.js#L102]
image
2. Pnpm hoist eslint by default. If Pnpm do not hoist eslint by default in the future, using `exec('eslint . --fix')` will always fail. See https://pnpm.io/npmrc#public-hoist-pattern.
3. If you have time, please see my source code [here](https://github.com/zanminkian/git-validator/tree/main/packages/git-validator/src/bin) and the [PR](https://github.com/zanminkian/git-validator/pull/125/files). I'm not sure that I have explained completely. If you need more info, please let me know. I am glad to provide it.

zanminkian:
Hi @nzakas. Following your suggection, it can be executed. But the [code](https://github.com/zanminkian/git-validator/blob/main/packages/git-validator/src/bin/eslint.js) will be a little bit complex. Comparing with the implementaions about [prettier](https://github.com/zanminkian/git-validator/blob/main/packages/git-validator/src/bin/prettier.js), [lint-staged](https://github.com/zanminkian/git-validator/blob/main/packages/git-validator/src/bin/lint-staged.js) and [commitlint](https://github.com/zanminkian/git-validator/blob/main/packages/git-validator/src/bin/commitlint.js), integrating eslint one is the most the most complicated. Any one of the others is only one line of code. All of them can do it. Why cann't eslint? 🤔

nzakas:
See, you now have a script that will work for all CLIs regardless of whether or not the `package.json` exports a `bin`. This is a much more reliable way of working with any CLIs, as opposed to continuing to ask every CLI you use to update their `package.json`. I'd suggest you make that your preferred way of working with CLIs and continue to tweak the approach to make it easier for yourself. I don't think 14 lines of code to universally work with any CLI in the Node.js ecosystem is too complex.

Issue #17646, opened 2023-10-13
Bug: Docs site "return to top" button overlaps language dropdown


nzakas:
On the docs site, if you scroll all the way down on a page in a desktop browser, the "return to top" button overlaps the controls on the lower-right.
![Screenshot 2023-10-13 at 11-44-32 Getting Started with ESLint - ESLint - Pluggable JavaScript Linter](https://github.com/eslint/eslint/assets/38546/e979121d-e422-4570-8bbf-93ab08c8ddea)

I'm not sure how to fix this. Maybe the button should appear to the left of that vertical line? Maybe it should stop before all of the controls in the lower right? Other ideas?

Tanujkanti4441:
We may shift the button a bit above so it can be placed just between the theme switcher and language switcher.

nzakas:
@Tanujkanti4441 I think we would end up with the same problem, only the button would overlap the controls above it depending on the vertical scroll position.

Tanujkanti4441:
> @Tanujkanti4441 I think we would end up with the same problem, only the button would overlap the controls above it depending on the vertical scroll position.

Agreed! It will somehow overlap the controls. Then if we planning to make it appear on the left of the vertical line i am with @harish-sethuraman we first should check how it is appearing there. And for mobile we can keep it's position as it is currently.

nzakas:
Agreed. We always want to see how things look first before making a change.

Tanujkanti4441:
Here are the screenshots after shifting the return to top button to the left of the vertical line.
<details>
<summary>screenshot 1</summary>

![Annotation 2023-10-23 130725](https://github.com/eslint/eslint/assets/86398394/839fc658-b598-4d0f-876c-2017e4941816)
</details>

<details>
<summary>screenshot 2</summary>

![Annotation 2023-10-23 135328](https://github.com/eslint/eslint/assets/86398394/6097e40f-4147-41d1-aad5-27b10c51c7d9)
</details>

<details>
<summary>screenshot 3</summary>

![Screenshot 2023-10-23 at 13-29-16 Configuration Migration Guide - ESLint - Pluggable JavaScript Linter](https://github.com/eslint/eslint/assets/86398394/d04f8a73-e8d2-4e58-9837-10d900093d83)
</details>

screenshot for below 1024px screen size

<details>
<summary>screenshot 1</summary>

[...]

PR #17848, opened 2023-12-13
docs: place the playground button in the bottom right corner

kecrily:
#### Prerequisites checklist

[x] Documentation update
[ ] Add autofix to a rule
[ ] Add a CLI option
[ ] Add something to the core
[ ] Other, please explain:

Removes large white space from the code block.

<table>
  <tr>
    <th>Before</th>
    <th>After</th>
  </tr>

  <tr>
    <td>
image</td>
    <td>
image
</td>
  </tr>
</table>

### DIFF
--- a/docs/src/assets/scss/docs.scss
+++ b/docs/src/assets/scss/docs.scss
@@ -104,6 +104,11 @@ html {
 div.correct,
 div.incorrect {
     position: relative;
+    background-color: var(--lightest-background-color);
+    padding-top: 1.5rem;
+    padding-bottom: 1.5rem;
+    border-radius: var(--border-radius);
+    margin-bottom: 3rem;
 
     &::after {
         position: absolute;
@@ -111,6 +116,16 @@ div.incorrect {
         right: -22px;
         offset-inline-end: -22px;
         offset-block-start: -22px;
+    }
+
+    & > pre[class*="language-"] {
+        padding-top: unset;
+        padding-bottom: unset;
+        margin: 0;
+
+        & > .line-numbers-wrapper {
+            padding-top: unset;
+        }
     }
 }
 
--- a/docs/src/assets/scss/docs.scss
+++ b/docs/src/assets/scss/docs.scss
@@ -112,11 +112,6 @@ div.incorrect {
         offset-inline-end: -22px;
         offset-block-start: -22px;
     }
-
-    // Add space to the bottom if there is a Playground button.
-    .c-btn.c-btn--playground ~ pre.line-numbers-mode {
-        padding-bottom: 4.5rem;
-    }
 }
 
 div.correct {
@@ -147,13 +142,9 @@ pre[class*="language-"] {
 .c-btn.c-btn--playground {
     position: absolute;
     font-size: var(--step--1);
-    bottom: 1rem;
-    right: 1rem;
+    bottom: -1.5rem;
+    right: -1.25rem;
     z-index: 1;
-
-    @media all and (min-width: 768px) {
-        bottom: 1.5rem;
-    }
 }
 
 @media (hover: none) {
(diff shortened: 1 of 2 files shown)


nzakas:
It does look a bit better this way, but I think we'll need to adjust the styles in light mode:

Screenshot 2023-12-13 at 10-56-14 getter-return - ESLint - Pluggable JavaScript Linter

I think the button needs a shadow or border to help make the edges clearer

kecrily:
image

fasttime:
I can see that after this change there will be extra empty space after all code blocks, even if there is no Playground button:

| Now | After |
|--------|------|
| image | image |

Is this intended?

mdjermanovic:
IMO, it looks better when the button is in the code block. Also, if the button is over the border, when the horizontal scroll bar appears (usually it does on small screens, e.g. mobile), the button will overlap with it:

image

kecrily:
> Is this intended?

@fasttime Yes, to avoid overlapping with other content on a small screen. But it shouldn't be found in a normal code block. now adjusted.

> IMO, it looks better when the button is in the code block. Also, if the button is over the border, when the horizontal scroll bar appears (usually it does on small screens, e.g. mobile), the button will overlap with it:

@mdjermanovic I fixed it. The scrollbar now no longer overlaps the buttons.

Tanujkanti4441:
Screenshot 2023-12-17 222423

in my browser i have different scrollbar is it looking fine?

kecrily:
> in my browser i have different scrollbar is it looking fine?

There's nothing wrong with its position, but it doesn't look very good.

harish-sethuraman:
I feel like something like this on hover would be better with help text open in playground probably? 
Screenshot 2023-12-18 at 4 30 51 PM

we could use something like this to denote correct and in correct instead? 

Screenshot 2023-12-18 at 4 32 02 PM


end output would be somewhere close to this? (😅  edited entire thing in inspect tools design could be better). 
Why I'd prefer something like this is because,
-  we show a link to playground when we hover on code sample so it doesn't obstruct the view?   
- the correct in correct doesn't float anymore and lives more with the code block itself

Screenshot 2023-12-18 at 5 23 09 PM

Tanujkanti4441:
i like the last design and instead of keeping that playground button in the code block we can place it to the correct/incorrect block with the `playground` name.

nzakas:
This PR was just to move the button -- there's no good reason to change the overall design of the correct/incorrect code examples. If we can't agree on moving the button, then we should just close this PR.

Tanujkanti4441:
if we shift the button to the bottom right corner we will end up with the same overlapping problem which we had previously

Screenshot 2023-12-23 201015

amareshsm:
> if we shift the button to the bottom right corner we will end up with the same overlapping problem which we had previously
> 
> Screenshot 2023-12-23 201015

Yes we need to consider. Boz of this issue we decided to place the button inside the block in https://github.com/eslint/eslint/pull/17403

kecrily:
This button is floating. It does not interfere with reading. And even if you don't move the button to the bottom right corner, it obscures the correct/incorrect icon.

image

Tanujkanti4441:
> Back to top button is floating. It does not interfere with reading, just slide up or down. And even if don't move the button to the bottom right corner, it still overlaps the correct/incorrect icon.
> image

[comment](https://github.com/eslint/eslint/issues/17646#issuecomment-1780049570), as per this comment it is okay if button overlaps the things which  can't be clicked.

kecrily:
image

Is this OK?

nzakas:
Folks - this discussion has been going on for a while, so my vote is just to leave the button as-is. If it was an easy fix, that would be one thing, but we're continuing to run into edge cases and I just don't think it's worth the effort to continue down this path.

Thoughts?

harish-sethuraman:
I'd also vote on keeping the same layout 👍🏻

nzakas:
All right, let's just keep what we have. I appreciate the suggestions and work, but let's move on to more interesting problems. :)

PR #17862, opened 2023-12-15
feat: throw error when `.eslintignore` file detected with flat config

### LINKED ISSUE #17831: Change Request: Warn when `.eslintignore` detected with flat config
### ESLint version

v8.55.0

After migrating my repository from `.eslintrc.js` to flat config `eslint.config.js`, I was confused by some additional linting errors that showed up. I then realized these linting errors were in files that were supposed to be ignored by `.eslintignore`. After doing some digging through ESLint github issues/PRs, I found out `.eslintignore` no longer works with flat config.

1. Document that `.eslintignore` no longer works in the migration guide: https://eslint.org/docs/latest/use/configure/migration-guide
2. Most importantly, add an error or warning when `.eslintignore` is detected with flat config, similar to how we often show a warning when the user provides an option that has been removed/deprecated.

If we don't have automated detection for this, then inevitably a large percentage of users performing this migration will meet this same confusion, and some will forget to cleanup/delete their old `.eslintignore` file. Having an `.eslintignore` present with flat config is a recipe for confusion. I can imagine some users even trying to keep `.eslintignore` and `ignores` in `eslint.config.js` in sync because they aren't sure which they are supposed to use.

Ideally, we could throw an error to disallow this mistake entirely, especially since flat config isn't considered stable yet until ESLint v9. Otherwise, we can print a warning to the console. The error/warning message would describe the replacement for this file or link to documentation on it.

Concerns:
1. Is there any situation in which the user would need to ignore the presence of `.eslintignore`?

### Participation

### Additional comments

Are there any other old files we should warn about to ease the migration? Should we warn when both `.eslintrc.js` and `eslint.config.js` are present?

### Related

* https://github.com/eslint/eslint/issues/17822
* https://github.com/eslint/eslint/pull/16434
* https://github.com/eslint/eslint/pull/16355

bmish:
#### Prerequisites checklist

[ ] Documentation update
[ ] Add autofix to a rule
[ ] Add a CLI option
[x] Add something to the core
[ ] Other, please explain:

Throw an error when the old `.eslintignore` file is detected with flat config.

Fixes #17831. If that issue doesn't get accepted, we can close this PR.

TODO:

- [ ] Fix a lot of tests that are failing because they have flat config with an `.eslintignore` file present
- [ ] Update the error message to describe the replacement for `.eslintignore` or point to documentation for the replacement

1. Is this the right place to throw an error?

### DIFF
--- a/lib/eslint/flat-eslint.js
+++ b/lib/eslint/flat-eslint.js
@@ -594,6 +594,10 @@ class FlatESLint {
             });
         }
 
+        // Check for old .eslintignore file.
+        if (existsSync(path.resolve(processedOptions.cwd, ".eslintignore"))) {
+            throw new Error(".eslintignore file is unsupported in flat config.");
+        }
     }
 
     /**
--- a/tests/lib/eslint/flat-eslint.js
+++ b/tests/lib/eslint/flat-eslint.js
@@ -4408,6 +4408,14 @@ describe("FlatESLint", () => {
 
             await assert.rejects(() => eslint.isPathIgnored(null), /'filePath' must be a non-empty string/u);
         });
+
+        describe("with unsupported .eslintignore file", () => {
+            it("throws an error", async () => {
+                const cwd = getFixturePath("ignored-paths");
+
+                assert.throws(() => new FlatESLint({ cwd }), /\.eslintignore file is unsupported in flat config\./u);
+            });
+        });
     });
 
     describe("loadFormatter()", () => {


aladdin-add (commented):
I feel like it better be just a warning, not an error. users may just forgot to remove it, and it does not stop eslint's functionality.

bmish:
@aladdin-add that's fair, can you suggest that on the issue #17831 too, want to have the main discussion on the issue.

PR #17876, opened 2023-12-19
chore: migrate linting workflow to use trunk check metalinter

EliSchleifer:
#### Prerequisites checklist

- [x ] I have read the [contributing guidelines](https://github.com/eslint/eslint/blob/HEAD/CONTRIBUTING.md).

[ ] Documentation update
[ ] Add autofix to a rule
[ ] Add a CLI option
[ ] Add something to the core
[x ] Other, please explain: chore to migrate linting solution to trunk check - universal metalinter

Trunk is a universal meta-linter and therefore obfuscates the need for separate calls to markdownlint / eslint / etc...

- Removed markdownlintignore (trunk by default ignores all files in .gitignore and added separately ignore of changelog.md to trunk.yaml)
- All tools will run in batch mode where possible (eslint, markdownlint, svgo)

I've left in place the current git-hook implementation. I want to make sure the basic ergonomics work before migrating these scripts into trunk actions as well.

Make sure you like the ergonomics of the tool as currently integrated


nzakas:
Thanks for putting this together. It looks like the CI is failing with a message "trunk not found". Is there a way to get this working in CI to see what the output looks like?

We're working on getting our v9.0.0 alpha out this week, so will have more time to investigate this next week.

EliSchleifer:
Will get that fixed. I’m on vacation till end of the week
________________________________

.trunk/trunk.yaml
```diff
@@ -0,0 +1,59 @@
...
+        - node
+# This is the section where you manage your linters. (https://docs.trunk.io/check/configuration)
+lint:
+    definitions:
+        - name: eslint
+          commands:
+              - name: lint
+                run: ${workspace}/bin/eslint.js --output-file ${tmpfile} --format json ${target}
+    enabled:
+        - eslint@8.56.0
```
nzakas, line 29: I'm curious if this is needed given the previous definition?

.trunk/trunk.yaml
```diff
@@ -0,0 +1,59 @@
...
+          # Default `check` mode will not run on anything in docs
+          paths:
+              - docs/**
+actions:
+    disabled:
+        - trunk-announce
+        - trunk-check-pre-push
+        - trunk-fmt-pre-commit
+    enabled:
+        - trunk-upgrade-available
```
nzakas, line 59: Question: Will something like Renovate know to upgrade trunk? Or is this something that needs to happen manually?

package.json
```diff
@@ -30,7 +30,8 @@
     "test": "node Makefile.js test",
     "test:cli": "mocha",
     "test:fuzz": "node Makefile.js fuzz",
-    "test:performance": "node Makefile.js perf"
+    "test:performance": "node Makefile.js perf",
+    "trunk": "trunk"
```
nzakas, line 34: I'm thinking we'd probably want `"lint": "trunk"` if the intent is to replace the current linting task?

.trunk/trunk.yaml
EliSchleifer: Currently upgrading is manual - we are working on a PR to make renovate aware of our schema.

.trunk/trunk.yaml
EliSchleifer: Yeah - the definition teaches trunk how to run a tool - and the enabled list turns on the tools.

package.json
EliSchleifer: yeah - I just need to extract the json file check and then we're in business

EliSchleifer:
@nzakas - I think this cleans up all the issues and at least let's you kick tires completely. I didn't see an active GitHub action that ran ESLint on itself - is the lint step right now fully optional? 

I can add a `trunk check` step that would automatically run check on each PR and make sure all rules are being followed.

nzakas:
Because you're a first-time contributor, the lint check doesn't run automatically -- I just manually approved it so it will run now. It's titled "Verify Files" in the actions list.

d
nzakas, line 1: This looks like it was included accidentally?

.trunk/trunk.yaml
```diff
@@ -0,0 +1,70 @@
...
+        - markdownlint@0.38.0
+        - oxipng@9.0.0
+        - prettier@3.1.1
+        - renovate@37.122.0
+        - shellcheck@0.9.0
+        - shfmt@3.6.0
+        - svgo@3.2.0
+        - trivy@0.48.2
+        - trufflehog@3.63.7
+        - yamllint@1.33.0
```
nzakas, line 43: Do these versions correspond to the versions in `package.json`? I'm just concerned about needing to update two places when these change.

...or do these have no relation to `package.json` at all?

d
nzakas, line 1: Can we remove this file?

EliSchleifer: Geeze - that's a ridiculous file.

nzakas:
I've tried checking this out locally and on my Windows machine I'm getting the following:

```
$ npm run trunk

> eslint@9.0.0-alpha.0 trunk
> trunk

The system cannot find the path specified.
```

I tried running `npm install`, but it seems that doesn't help. Does trunk work on Windows? Is there something else I need to do get it to work locally?

EliSchleifer:
> I've tried checking this out locally and on my Windows machine I'm getting the following:
> 
> ```
> $ npm run trunk
> 
> > eslint@9.0.0-alpha.0 trunk
> > trunk
> 
> The system cannot find the path specified.
> ```
> 
> I tried running `npm install`, but it seems that doesn't help. Does trunk work on Windows? Is there something else I need to do get it to work locally?

The product does work on windows - but not sure we've tested the npm flow - I am not a node/windows expert - but we have those on staff. They're taking a look.

EliSchleifer:
two issues here:
1. I mistakenly changed call to `trunk check` from CI even though trunk was not on the path - so moved to use npm run
2. You uncovered a bug in our launcher for windows - having someone from the team fix that up. @nzakas - do you use powershell on your windows box or bash emulation?

.trunk/configs/.yamllint.yaml
JoshuaKGoldberg, line 1: There also exists an [`eslint-plugin-yml`](https://ota-meshi.github.io/eslint-plugin-yml). I suspect it would be good community-dogfooding for ESLint to use that rather than a separate tool. Is that something that can be reasonably done in metalinter without sacrificing features?

EliSchleifer:
on it - also looking into how we can better support the mode of running on docs only in CI and not locally. small product change to make that a better experience and map to what we're aiming for here

> @EliSchleifer looks like we have the CI failing again on this.

Tanujkanti4441:
hi @EliSchleifer, you can share the updates regarding this PR or if it is ready to review.

EliSchleifer:
I need to update the config - so the verification step is properly modeling the desire to handle CI flow differently from PR local flow. I should be able to update it this week

.trunk/configs/.yamllint.yaml
EliSchleifer: Let me take this as an action item for a follow up to use the eslint yaml plugin.

.github/workflows/ci.yml
```diff
@@ -33,9 +33,6 @@ jobs:
     - name: Check Licenses
       run: node Makefile checkLicenses
 
-    - name: Lint Docs JS Files
```
nzakas, line 36: We wanted this a separate step so we could more easily tell where a failure was on a PR.

EliSchleifer: The GitHub action from trunk will annotate the files inline with the issue. Is that not sufficient to get an understanding of the source of the problem. 

I can create a sub branch to demonstrate

nzakas: Will those show up on the Discussion tab of the PR? What I'm looking for is the "glance-ability" of a PR so I don't have to switch to the code tab to get a general understanding which CI checks failed. This is key when we are barraged with PRs and trying to triage.

Rec0iL99: ping @EliSchleifer

EliSchleifer:
Got a plan to make this work beautifully. Will post update tomorrow morning
________________________________

fasttime:
@EliSchleifer there are merge conflicts. Can you take a look?

det:
I have opened a new PR here: https://github.com/eslint/eslint/pull/18643

fasttime:
> I have opened a new PR here: #18643

Thanks @det! I'm going to close this PR in favor of the new one (and thanks @EliSchleifer for the initial work).

PR #17935, opened 2024-01-01
docs: missing word

brettz9:
#### Prerequisites checklist

[x] Documentation update
[ ] Add autofix to a rule
[ ] Add a CLI option
[ ] Add something to the core
[ ] Other, please explain:

Added a missing word.

I could be mistaken in my understanding here, but if so, I think the wording should be changed to "matched by all other configuration objects".

### DIFF
--- a/docs/src/use/configure/configuration-files.md
+++ b/docs/src/use/configure/configuration-files.md
@@ -56,7 +56,7 @@ module.exports = [
 
 Each configuration object contains all of the information ESLint needs to execute on a set of files. Each configuration object is made up of these properties:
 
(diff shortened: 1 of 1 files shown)


platinumazure (changes requested):
I think this change is incorrect, based on the introductory post here: https://eslint.org/blog/2022/08/new-config-system-part-2/#glob-based-configs-everywhere

I think the intent is that not specifying `files` is roughly equivalent to specifying the union of the other files specified in other config objects, so that they don't have to be specified again for common configuration settings.

brettz9:
Thanks, that helps make it more clear. I think that phrasing or at least "matched by all other configuration objects" rather than "any" makes it more clear.

brettz9:
I guess I'll review that post and re-submit if I think it can be clarified. Thanks!

Issue #18002, opened 2024-01-17
Docs: options section in `no-restricted-import` rule


Tanujkanti4441:
### Docs page(s)

https://eslint.org/docs/latest/rules/no-restricted-imports

### What documentation issue do you want to solve?

`no-restricted-imports` rule has lots of options and all are explained with a one line description in a row with no particular heading, i think this make it a little bit complex to understand at first and also  PR #16196 and issue #15261 may add new feature or option and will need to be arranged properly in the docs.

### What do you think is the correct solution?

every option can have it's own dedicated section with correct/incorrect examples.

### Participation

- [X] I am willing to submit a pull request for this change.

nzakas:
Sounds like a good idea. Go for it! :+1:

PR #18015, opened 2024-01-20
docs: update `no-restricted-imports` rule

Tanujkanti4441:
#### Prerequisites checklist

[x] Documentation update
[ ] Add autofix to a rule
[ ] Add a CLI option
[ ] Add something to the core
[ ] Other, please explain:

Putted everything under a particular heading in `no-restricted-imports` rule docs page

Fixes: #18002

### DIFF
--- a/docs/src/rules/no-restricted-imports.md
+++ b/docs/src/rules/no-restricted-imports.md
@@ -62,7 +62,6 @@ export * from 'fs';
 
 :::
 
-
 Examples of **correct** code for string option:
 
 ::: correct { "sourceType": "module" }
@@ -278,7 +277,6 @@ import pick from 'import1/private/someModule';
 In last example `"!import1/private/*"` is not reincluding the modules inside private directory because in `gitignore-style` negation mark (`!`) does not reinclude the files if it's parent directory is excluded by a pattern. In this case `import1/private` directory is already excluded by the `import1/*` pattern, but excluded directory can be reincluded like using `"!import1/private"`.
 
 Examples of **correct** code for `pattern` option:
-
 
 ::: correct { "sourceType": "module" }
 
--- a/docs/src/rules/no-restricted-imports.md
+++ b/docs/src/rules/no-restricted-imports.md
@@ -22,7 +22,7 @@ It applies to static imports only, not dynamic ones.
 
 This rule has both string and object options to specify the imported modules to restrict.
 
-Simplest way to specify a module that you want to restrict from being imported is to pass it's name as a string to the options:
+Using string option, you can specify  the name of a module that you want to restrict from being imported as a value in the rule options array:
 
 ```json
 "no-restricted-imports": ["error", "import1", "import2"]
@@ -40,7 +40,7 @@ import fs from 'fs';
 
 :::
 
-Specified string can also restrict the module from being exported:
+String options also restrict the module from being exported, as in this example:
 
 ::: incorrect { "sourceType": "module" }
 
@@ -75,7 +75,7 @@ export { foo } from "bar";
 
 :::
 
-You may also specify a custom message for a particular module simply using `name` and `message` properties inside an object, where the value of the `name` property will be the name of the module and `message` property will contain the custom message. The custom message will be appended to the default error message:
+You may also specify a custom message for a particular module using the `name` and `message` properties inside an object, where the value of the `name` is the name of the module and `message` property contains the custom message. (The custom message is appended to the default error message from the rule.)
 
 ```json
 "no-restricted-imports": ["error", {
@@ -104,7 +104,7 @@ import foo from 'disallowed-import';
 
 ### paths
 
-This is an object option whose value is an array and the name of the module you want to restrict can be specified inside this array, same as in string option above.
+This is an object option whose value is an array containing the names of the modules you want to restrict.
 
 ```json
 "no-restricted-imports": ["error", { "paths": ["import1", "import2"] }]
@@ -122,7 +122,7 @@ import cluster from 'cluster';
 
 :::
 
-Custom messages for a particular module can also be specified in `paths` array option using objects as above.
+Custom messages for a particular module can also be specified in `paths` array using objects with `name` and `message`.
 
 ```json
 "no-restricted-imports": ["error", {
@@ -138,9 +138,10 @@ Custom messages for a particular module can also be specified in `paths` array o
 
 #### importNames
 
-This option in `paths` is an array and can be used to specify certain imports to restrict that are being exported from a module. Import names specified inside `paths` array looks for the module specified in the `name` property of corresponding object, so it is required to specify the `name` key with both `importNames` and `message` options.
+This option in `paths` is an array and can be used to specify the names of certain bindings exported from a module. Import names specified inside `paths` array affect the module specified in the `name` property of corresponding object, so it is required to specify the `name` property first when you are using `importNames` or `message` option.
 
-Passing `"default"` string inside the `importNames` array will restrict the import that is being exported by default by the module.
+
+Specifying `"default"` string inside the `importNames` array will restrict the default export from being imported.
 
 ```json
 "no-restricted-imports": ["error", {
@@ -204,7 +205,7 @@ import * as Foo from "foo";
 
 Examples of **correct** code for `importNames` in `paths`:
 
-Other strings doesn't report default exported import even if the names are same:
+If the local name assigned to a default export is the same as a string in `importNames`, this will not cause an error.
 
 ::: correct { "sourceType": "module" }
 
@@ -232,9 +233,9 @@ import { AllowedObject as DisallowedObject } from "foo";
 
 ### patterns
 
-This is also an object option whose value is an array. This option comes handy when you need to restrict the modules present inside a perticular directory. Simply, you can specify the modules inside the directories you want to restrict as string in `gitignore-style` patterns.
+This is also an object option whose value is an array. This option allows you to specify multiple modules to restrict using `gitignore`-style patterns.
 
-Since the patterns follows the `gitignore-style` then if you want to reinclude any particular module this can be done by prefixing a negation (`!`) mark in front of the pattern and it's good to keep the negation patterns in last as `gitignore-style` follows order:
+Because the patterns follow the `gitignore`-style, if you want to reinclude any particular module this can be done by prefixing a negation (`!`) mark in front of the pattern. (Negated patterns should come last in the array because order is important.)
 
 ```json
 "no-restricted-imports": ["error", {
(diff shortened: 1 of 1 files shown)


docs/src/rules/no-restricted-imports.md
```diff
@@ -20,176 +20,139 @@ It applies to static imports only, not dynamic ones.
 
 ## Options
 
-The syntax to specify restricted imports looks like this:
+This rule has both string and object options to specify the imported modules to restrict.
+
+Simplest way to specify a module that you want to restrict from being imported is to pass it's name as a string to the options:
```
nzakas, line 25:
```suggestion
Using string options, you can specify the name of a module that you want to restrict from being imported as a value in the rule options array:
```

docs/src/rules/no-restricted-imports.md
```diff
@@ -20,176 +20,139 @@ It applies to static imports only, not dynamic ones.
...
-```json
-"no-restricted-imports": ["error", {
-    "name": "import-foo",
-    "message": "Please use import-bar instead."
-}, {
-    "name": "import-baz",
-    "message": "Please use import-quux instead."
-}]
-```
+Specified string can also restrict the module from being exported:
```
nzakas, line 43:
```suggestion
String options also restrict the module from being exported, as in this example:
```

docs/src/rules/no-restricted-imports.md
```diff
@@ -20,176 +20,139 @@ It applies to static imports only, not dynamic ones.
...
-    }]
-}]
+import crypto from 'crypto';
+export { foo } from "bar";
 ```
 
-Regex patterns can also be used to restrict specific import Name:
+:::
+
+You may also specify a custom message for a particular module simply using `name` and `message` properties inside an object, where the value of the `name` property will be the name of the module and `message` property will contain the custom message. The custom message will be appended to the default error message:
```
nzakas, line 78:
```suggestion
You may also specify a custom message for a particular module using the `name` and `message` properties inside an object, where the value of the `name` is the name of the module and `message` property contains the custom message. (The custom message is appended to the default error message from the rule.)
```

docs/src/rules/no-restricted-imports.md
```diff
@@ -20,176 +20,139 @@ It applies to static imports only, not dynamic ones.
...
 ```
 
 :::
 
-::: incorrect { "sourceType": "module" }
+### paths
 
-```js
-/*eslint no-restricted-imports: ["error", "fs"]*/
+This is an object option whose value is an array and the name of the module you want to restrict can be specified inside this array, same as in string option above.
```
nzakas, line 107:
```suggestion
This is an object option whose value is an array containing the names of the modules you want to restrict.
```

docs/src/rules/no-restricted-imports.md
```diff
@@ -20,176 +20,139 @@ It applies to static imports only, not dynamic ones.
...
+/*eslint no-restricted-imports: ["error", { "paths": ["cluster"] }]*/
 
-export * from 'fs';
+import cluster from 'cluster';
 ```
 
 :::
 
-::: incorrect { "sourceType": "module" }
+Custom messages for a particular module can also be specified in `paths` array option using objects as above.
```
nzakas, line 125:
```suggestion
Custom messages for a particular module can also be specified in `paths` array using objects with `name` and `message`.
```

docs/src/rules/no-restricted-imports.md
```diff
@@ -20,176 +20,139 @@ It applies to static imports only, not dynamic ones.
...
+        "message": "Please use import-quux instead."
+    }]
+}]
 ```
 
-:::
+#### importNames
 
-::: incorrect { "sourceType": "module" }
+This option in `paths` is an array and can be used to specify certain imports to restrict that are being exported from a module. Import names specified inside `paths` array looks for the module specified in the `name` property of corresponding object, so it is required to specify the `name` key with both `importNames` and `message` options.
```
nzakas, line 141:
```suggestion
This option in `paths` is an array and can be used to specify the names of certain bindings exported from a module. Import names specified inside `paths` array affect the module specified in the `name` property of corresponding object, so it is required to specify the `name` key with both `importNames` and `message` options.
```

Question: is `message` actually required when using `importNames`? It doesn't seem like it should be so just double-checking.

docs/src/rules/no-restricted-imports.md
```diff
@@ -20,176 +20,139 @@ It applies to static imports only, not dynamic ones.
...
 
-```js
-/*eslint no-restricted-imports: ["error", { "patterns": ["lodash/*"] }]*/
+Passing `"default"` string inside the `importNames` array will restrict the import that is being exported by default by the module.
```
nzakas, line 143:
```suggestion
Specifying `"default"` inside `importNames` array will restrict the default export from being imported.
```

docs/src/rules/no-restricted-imports.md
```diff
@@ -237,137 +202,147 @@ import * as Foo from "foo";
 
 :::
 
-::: incorrect { "sourceType": "module" }
+Examples of **correct** code for `importNames` in `paths`:
+
+Other strings doesn't report default exported import even if the names are same:
```
nzakas, line 207: I'm not entirely sure what this means. Maybe this?
```suggestion
If the local name assigned to a default export is the same as a string in `importNames`, this will not cause an error.
```

docs/src/rules/no-restricted-imports.md
```diff
@@ -237,137 +202,147 @@ import * as Foo from "foo";
...
-::: incorrect { "sourceType": "module" }
+### patterns
 
-```js
-/*eslint no-restricted-imports: ["error", { patterns: [{
-    group: ["utils/*"],
-    importNames: ['isEmpty'],
-    message: "Use 'isEmpty' from lodash instead."
-}]}]*/
+This is also an object option whose value is an array. This option comes handy when you need to restrict the modules present inside a perticular directory. Simply, you can specify the modules inside the directories you want to restrict as string in `gitignore-style` patterns.
```
nzakas, line 235:
```suggestion
This is also an object option whose value is an array. This option allows you to specify multiple modules to restrict using `gitignore`-style patterns.
```

docs/src/rules/no-restricted-imports.md
```diff
@@ -237,137 +202,147 @@ import * as Foo from "foo";
...
 
-import { isEmpty } from 'utils/collection-utils';
+Since the patterns follows the `gitignore-style` then if you want to reinclude any particular module this can be done by prefixing a negation (`!`) mark in front of the pattern and it's good to keep the negation patterns in last as `gitignore-style` follows order:
```
nzakas, line 237:
```suggestion
Because the patterns follow `gitignore`-style, if you want to reinclude any particular module this can be done by prefixing a negation (`!`) mark in front of the pattern. (Negated patterns should come last in the array because order is important.)
```

docs/src/rules/no-restricted-imports.md
```diff
@@ -237,137 +202,147 @@ import * as Foo from "foo";
...
+/*eslint no-restricted-imports: ["error", { "patterns": ["import1/*", "!import1/private/*"] }]*/
 
-import { bar, bazQux } from 'foo/quux';
+import pick from 'import1/private/someModule';
 ```
 
 :::
 
-Examples of **correct** code for this rule:
+In last example `"!import1/private/*"` is not reincluding the modules inside private directory because in `gitignore-style` negation mark (`!`) does not reinclude the files if it's parent directory is excluded by a pattern. In this case `import1/private` directory is already excluded by the `import1/*` pattern, but excluded directory can be reincluded like using `"!import1/private"`.
```
nzakas, line 277:
```suggestion
In this example, `"!import1/private/*"` is not reincluding the modules inside `private` because the negation mark (`!`) does not reinclude the files if it's parent directory is excluded by a pattern. In this case, `import1/private` directory is already excluded by the `import1/*` pattern. (The excluded directory can be reincluded using `"!import1/private"`.)
```

docs/src/rules/no-restricted-imports.md
```diff
@@ -237,137 +202,147 @@ import * as Foo from "foo";
...
 ```
 
 :::
 
-::: correct { "sourceType": "module" }
+#### group
+
+Alternatively you can use this object property which is an array to specify the `gitignore-style` patterns for restricting modules. Custom messages can also be specified like `paths`.
+
+This is a required property if you are using object options inside the `patterns` array:
```
nzakas, line 315:
```suggestion
The `patterns` array can also include objects. The `group` property is used to specify the `gitignore`-style patterns for restricting modules and the `message` property is used to specify a custom message.

The `group` property is required property when using objects inside the `patterns` array.
```

docs/src/rules/no-restricted-imports.md
```diff
@@ -381,6 +356,36 @@ import lodash from 'lodash';
 
 :::
 
+#### caseSensitive
+
+This is a boolean option and allows the patterns specified in the `group` array to be case-sensitive when sets to `true`. Default is `false`:
```
nzakas, line 361:
```suggestion
This is a boolean option and sets the patterns specified in the `group` array to be case-sensitive when `true`. Default is `false`.
```

docs/src/rules/no-restricted-imports.md
```diff
@@ -394,6 +399,38 @@ import pick from 'food';
 
 :::
 
+#### importNames
+
+`patterns` too has an `importNames` property and this works similar as in `paths`, it just looks for patterns specified inside the `group` array:
```
nzakas, line 404:
```suggestion
You can also specify `importNames` on objects inside of `patterns`. In this case, the specified names are applied only to the specified `group`.
```

docs/src/rules/no-restricted-imports.md
```diff
@@ -422,6 +518,14 @@ import isEmpty, { hasValue } from 'utils/collection-utils';
...
 :::
 
+To restrict the use of all Node.js core imports (via <https://github.com/nodejs/node/tree/master/lib>):
+
+```json
+    "no-restricted-imports": ["error",
+         "assert","buffer","child_process","cluster","crypto","dgram","dns","domain","events","freelist","fs","http","https","module","net","os","path","punycode","querystring","readline","repl","smalloc","stream","string_decoder","sys","timers","tls","tracing","tty","url","util","vm","zlib"
+    ],
+```
+
```
nzakas, line 528: I would leave this out. It's a very specific use case that it easy to figure out if you need it. (It would probably also be a better idea to use the `globals` package for this purpose.)

nzakas (changes requested):
Thanks for doing this! I've tried to clean up the language throughout, please take a look.

docs/src/rules/no-restricted-imports.md
Tanujkanti4441: > it is required to specify the `name` key with both `importNames` and `message` options.

For me this line is saying when using `importNames` or `message` property, then it is require to use `name` property first. I think i should clear it's language a bit.

docs/src/rules/no-restricted-imports.md
Tanujkanti4441: i this means we can remove this section?

docs/src/rules/no-restricted-imports.md
Tanujkanti4441: Done! till here

docs/src/rules/no-restricted-imports.md
Tanujkanti4441: Exactly this.

docs/src/rules/no-restricted-imports.md
nzakas: Yes, I think we can remove this section. It's not necessary.

Tanujkanti4441: Done!

nzakas (approved):
LGTM. Thanks!

PR #18041, opened 2024-01-27
docs: show red underlines in examples in rules docs

### LINKED ISSUE #16227: Change Request: Show red underlines in examples in rules docs
### ESLint version

N/A

* Make the example errors in rules docs easier to understand.
* Have the rules docs more directly demonstrate the user experience of enabling the rule (what you see in the docs mirrors what you will see in your editor)
* (optional) Surface actual lint error messages in the example code on hover.

Implement a custom syntax highlighter (inspired by [shiki-twoslash]) which, at build time, optionally runs code through ESLint and augments the syntax-highlighted HTML with diagnostics. This could be implemented as code directly in ESLint, or as a standalone npm module pulled in as a dependency.

### Participation

### Additional comments

I've been playing with a version of this locally and I have something which works, but it will need quite a bit more work to get it robust enough to be able to fully replace the existing Prism-based syntax highlighter.

Screen Shot 2022-08-22 at 2 04 57 PM

Thanks to @orta who has been looking into this with me.

ota-meshi:
#### Prerequisites checklist

[x] Documentation update
[ ] Add autofix to a rule
[ ] Add a CLI option
[ ] Add something to the core
[ ] Other, please explain:

This PR modifies `docs/src/_plugins/md-syntax-highlighter.js` to indicate where an error occurs with a red marker in the examples in the rules document.

close #16227

- This process is done using Prism's `after-tokenize` hook.
- ~~Content that starts with `/* eslint ` and whose language is js is considered an example of an eslint rule.~~
  It is now consider it an example of an ESLint rule only if it is wrapped in a custom container (that is, if parser options are passed from the custom container's rendering process).
  I would like to hear your opinion if this will work.
- I would like to hear your opinion on what style should be used.

### DIFF
--- a/docs/package.json
+++ b/docs/package.json
@@ -20,7 +20,8 @@
         "lint:scss": "stylelint \"**/*.{scss,html}\"",
         "lint:fix:scss": "npm run lint:scss -- --fix",
         "build:minify-images": "imagemin '_site/assets/images' --out-dir='_site/assets/images'",
-        "start": "npm-run-all build:sass build:postcss --parallel *:*:watch"
+        "start": "npm-run-all build:sass build:postcss --parallel *:*:watch",
+        "postinstall": "cd .. && npm install -f"
     },
     "devDependencies": {
         "@11ty/eleventy": "^2.0.1",
--- a/docs/package.json
+++ b/docs/package.json
@@ -20,8 +20,7 @@
         "lint:scss": "stylelint \"**/*.{scss,html}\"",
         "lint:fix:scss": "npm run lint:scss -- --fix",
         "build:minify-images": "imagemin '_site/assets/images' --out-dir='_site/assets/images'",
-        "start": "npm-run-all build:sass build:postcss --parallel *:*:watch",
-        "postinstall": "cd .. && npm install -f"
+        "start": "npm-run-all build:sass build:postcss --parallel *:*:watch"
     },
     "devDependencies": {
         "@11ty/eleventy": "^2.0.1",
--- a/docs/src/_plugins/prism-eslint-installer.js
+++ b/docs/src/_plugins/prism-eslint-installer.js
@@ -43,7 +43,7 @@ function install() {
             { filename: "code.js" },
         );
         if (messages.some((m) => m.fatal)) {
-            // If it contains a fatal error, change it to `ecmaVersion: 2015` and re-parse it.
+            // If it contains a fatal error, change it to remove `ecmaVersion: 2015` and re-parse it.
             messages = linter.verify(
                 code,
                 { languageOptions: { sourceType: "script" } },
--- a/docs/src/_plugins/prism-eslint-installer.js
+++ b/docs/src/_plugins/prism-eslint-installer.js
@@ -36,7 +36,15 @@ function install() {
         }
 
         // Run lint to extract the error range.
-        const messages = linter.verify(code, {}, { filename: "code.js" });
+        let messages = linter.verify(code, {}, { filename: "code.js" });
+        if (messages.some((m) => m.fatal)) {
+            // If it contains a parsing error, change it to `sourceType: 'script'` and re-parse it.
+            messages = linter.verify(
+                code,
+                { languageOptions: { sourceType: "script" } },
+                { filename: "code.js" },
+            );
+        }
         const ranges = messages.map((m) => {
             const start = getIndexFromLoc({
                 line: m.line,
--- a/docs/src/_plugins/prism-eslint-installer.js
+++ b/docs/src/_plugins/prism-eslint-installer.js
@@ -36,12 +36,25 @@ function install() {
         }
 
         // Run lint to extract the error range.
-        let messages = linter.verify(code, {}, { filename: "code.js" });
+        // Use `sourceType: "script"` and `ecmaVersion: 2015` to make the initial parsing loose.
+        let messages = linter.verify(
+            code,
+            { languageOptions: { sourceType: "script", parserOptions: { ecmaVersion: 2015 } } },
+            { filename: "code.js" },
+        );
         if (messages.some((m) => m.fatal)) {
-            // If it contains a parsing error, change it to `sourceType: 'script'` and re-parse it.
+            // If it contains a fatal error, change it to `ecmaVersion: 2015` and re-parse it.
             messages = linter.verify(
                 code,
                 { languageOptions: { sourceType: "script" } },
+                { filename: "code.js" },
+            );
+        }
+        if (messages.some((m) => m.fatal)) {
+            // If it contains a fatal error, change it to `sourceType: 'module'` and re-parse it.
+            messages = linter.verify(
+                code,
+                { languageOptions: { sourceType: "module" } },
                 { filename: "code.js" },
             );
         }
--- a/docs/src/_plugins/prism-eslint-installer.js
+++ b/docs/src/_plugins/prism-eslint-installer.js
@@ -1,0 +1,117 @@
+const Prism = require("prismjs");
+const { Linter } = require("../../../lib/api");
+const astUtils = require("../../../lib/shared/ast-utils");
+
+module.exports = { install };
+
+function install() {
+    const linter = new Linter({ configType: "flat" });
+    Prism.hooks.add("after-tokenize", (env) => {
+        // Content that starts with `/* eslint ` and whose language is js is considered an example of an eslint rule.
+        if (env.language !== "js" || !/^\s*\/\*\s*eslint\s/u.test(env.code)) {
+            return;
+        }
+
+        const code = env.code;
+        const tokens = env.tokens;
+
+        /** Copied from SourceCode constructor */
+        lineStartIndices = [0];
+        const lineEndingPattern = astUtils.createGlobalLinebreakMatcher();
+        let match;
+        while ((match = lineEndingPattern.exec(code))) {
+            lineStartIndices.push(match.index + match[0].length);
+        }
+        /**
+         * Converts a (line, column) pair into a range index.
+         * @param {Object} loc A line/column location
+         * @param {number} loc.line The line number of the location (1-indexed)
+         * @param {number} loc.column The column number of the location (1-indexed)
+         * Copied from SourceCode#getIndexFromLoc
+         * */
+        function getIndexFromLoc(loc) {
+            const lineStartIndex = lineStartIndices[loc.line - 1];
+            const positionIndex = lineStartIndex + loc.column - 1;
+            return positionIndex;
+        }
+
+        // Run lint to extract the error range.
+        const messages = linter.verify(code, {}, { filename: "code.js" });
+        const ranges = messages.map((m) => {
+            const start = getIndexFromLoc({
+                line: m.line,
+                column: m.column,
+            });
+            return [
+                start,
+                m.endLine == null
+                    ? start + 1
+                    : getIndexFromLoc({
+                          line: m.endLine,
+                          column: m.endColumn,
+                      }),
+            ];
+        });
+
+        let currentRange = ranges.shift();
+        if (!currentRange) {
+            return;
+        }
+        let start = 0;
+        env.tokens = [...convertMarked(tokens)];
+
+        /**
+         * Generates a token stream with the `marked` class assigned to the error range.
+         * @param {(string | Prism.Token)[]} tokens
+         */
+        function* convertMarked(tokens) {
+            for (const token of tokens) {
+                if (!currentRange) {
+                    yield token;
+                    return;
+                }
+                let content;
+                let buildToken;
+                if (typeof token === "string") {
+                    content = token;
+                    buildToken = (content) => content;
+                } else {
+                    if (typeof token.content !== "string") {
+                        token.content = [...convertMarked([token.content].flat())];
+                        yield token;
+                        continue;
+                    }
+                    content = token.content;
+                    buildToken = (content) =>
+                        content === token.content
+                            ? token
+                            : new Prism.Token(token.type, content, token.alias);
+                }
+                while (currentRange && content && currentRange[0] < start + content.length) {
+                    if (start < currentRange[0]) {
+                        const before = content.slice(0, currentRange[0] - start);
+                        yield buildToken(before);
+                    }
+                    const mark = content.slice(
+                        Math.max(currentRange[0] - start, 0),
+                        currentRange[1] - start
+                    );
+                    yield new Prism.Token(
+                        [typeof token === "string" ? "" : token.type, "eslint-marked"].join(" "),
+                        mark,
+                        token.alias
+                    );
+                    content = content.slice(currentRange[1] - start);
+                    start = start + mark.length;
+                    if (currentRange[1] <= start) {
+                        currentRange = ranges.shift();
+                    }
+                }
+                if (content.length > 0) {
+                    yield buildToken(content);
+                }
+                start = start + content.length;
+            }
+        }
+    });
+}
--- a/docs/src/_plugins/prism-eslint-installer.js
+++ b/docs/src/_plugins/prism-eslint-installer.js
@@ -88,21 +88,21 @@ function install() {
                             : new Prism.Token(token.type, content, token.alias);
                 }
                 while (currentRange && content && currentRange[0] < start + content.length) {
-                    if (start < currentRange[0]) {
-                        const before = content.slice(0, currentRange[0] - start);
+                    const before = content.slice(0, Math.max(currentRange[0] - start, 0));
+                    if (before) {
                         yield buildToken(before);
                     }
                     const mark = content.slice(
                         Math.max(currentRange[0] - start, 0),
-                        currentRange[1] - start
+                        currentRange[1] - start,
                     );
                     yield new Prism.Token(
                         [typeof token === "string" ? "" : token.type, "eslint-marked"].join(" "),
                         mark,
-                        token.alias
+                        token.alias,
                     );
                     content = content.slice(currentRange[1] - start);
-                    start = start + mark.length;
+                    start = start + before.length + mark.length;
                     if (currentRange[1] <= start) {
                         currentRange = ranges.shift();
                     }
--- a/docs/src/_plugins/prism-eslint-installer.js
+++ b/docs/src/_plugins/prism-eslint-installer.js
@@ -68,7 +68,7 @@ function install() {
             for (const token of tokens) {
                 if (!currentRange) {
                     yield token;
-                    return;
+                    continue;
                 }
                 let content;
                 let buildToken;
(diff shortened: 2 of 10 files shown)


ota-meshi:
It seems that CI does not install root dependencies, which causes the documentation to fail to build.
Can you temporarily change to install root dependencies to see how this PR changes the documentation?

ota-meshi:
It looks like this on my local.

image

fasttime:
Thanks for this PR! CI is failing because the job that builds the docs now requires ESLint dependencies to be installed, but only deps in the docs package (in the subfolder) are installed at that point:

https://github.com/eslint/eslint/blob/1bf2520c4166aa55596417bf44c567555bc65fba/.github/workflows/docs-ci.yml#L34-L36

We could run `npm install` in the parent folder before running the `build` script, or we could move this particular job to the main workflow in .github/workflows/ci.yml and have it running from there. But in this case, maybe we'll need to run `npm install` in the docs package as well. You should be able do all changes by yourself, I think.

docs/package.json
```diff
@@ -20,7 +20,8 @@
         "lint:scss": "stylelint \"**/*.{scss,html}\"",
         "lint:fix:scss": "npm run lint:scss -- --fix",
         "build:minify-images": "imagemin '_site/assets/images' --out-dir='_site/assets/images'",
-        "start": "npm-run-all build:sass build:postcss --parallel *:*:watch"
+        "start": "npm-run-all build:sass build:postcss --parallel *:*:watch",
+        "postinstall": "cd .. && npm install -f"
```
ota-meshi, line 24: I temporarily added postinstall so that I can check the results of this PR with netlify. I think you should probably replace this PR with another method before merging.
https://deploy-preview-18041--docs-eslint.netlify.app/rules/array-callback-return

docs/src/_plugins/prism-eslint-installer.js
```diff
@@ -0,0 +1,138 @@
...
+            );
+        }
+        if (messages.some((m) => m.fatal)) {
+            // If it contains a fatal error, change it to `sourceType: 'module'` and re-parse it.
+            messages = linter.verify(
+                code,
+                { languageOptions: { sourceType: "module" } },
+                { filename: "code.js" },
+            );
+        }
```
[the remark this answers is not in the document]
ota-meshi: Thank you for your advice!
Currently, we are only using `Prism` hooks, so we cannot retrieve that information. We will probably need to work with `markdown-it`. I'll look into that method.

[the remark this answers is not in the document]
ota-meshi: I changed this and the documentation works fine now!
However, the test CI is no longer able to resolve `prismjs` dependencies, so the CI now fails 😓 I will look for a way to solve this problem.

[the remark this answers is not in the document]
ota-meshi: I refactored it and solved the problem in ci test 😎

docs/tools/prism-eslint-hook.js
```diff
@@ -0,0 +1,170 @@
...
+            const positionIndex = lineStartIndex + loc.column - 1;
+
+            return positionIndex;
+        }
+
+        /*
+         * Run lint to extract the error range.
+         */
+        const messages = linter.verify(
+            code,
```
fasttime, line 74: In some examples of the `no-multiple-empty-lines` rule we use the character `⏎` to indicate the presence of a line break, e.g.:

https://github.com/eslint/eslint/blob/5a8efd5b7ad13eb320a1f468d1d4ab3c8ab99214/docs/src/rules/no-multiple-empty-lines.md?plain=1#L61-L70

This character is only useful as a visual guide, it should not be parsed. In fact, it is removed from the code when the rule example is validated or when the Playground link is generated:

https://github.com/eslint/eslint/blob/5a8efd5b7ad13eb320a1f468d1d4ab3c8ab99214/docs/tools/markdown-it-rule-example.js#L75-L78

To get correct results, I think we should apply the same kind of transformation to the `code` that is being linted here.

The fact that the column number could be off by one after this transformation should not be an issue at all, because the `no-multiple-empty-lines` rule only ever reports on empty lines.

ota-meshi: Thank you for checking this PR!
I fixed an issue related to `no-multiple-empty-lines`.
https://deploy-preview-18041--docs-eslint.netlify.app/rules/no-multiple-empty-lines

nzakas:
I think it's okay to update the CI in this PR.

This looks really good, thanks! Can we change the red underline to look more like the one VS Code uses? (A little squiggly.) I'm pretty sure it's just a repeating image that we can find in the VS Code repo.

docs/tools/prism-eslint-hook.js
```diff
@@ -0,0 +1,179 @@
...
+const contentsMustBeMarked = new Map();
+
+/**
+ * Add content that needs to be marked.
+ * @param {string} content Source code content that marks eslint errors.
+ * @param {ParserOptions} options The options used for validation.
+ * @returns {void}
+ */
+function addContentMustBeMarked(content, options) {
+    contentsMustBeMarked.set(content, options);
```
fasttime, line 22: I checked and we don't have any duplicate `content`, so this is okay. If we had any duplicates with different parser options, we could not rely just on the content as a key.

docs/tools/prism-eslint-hook.js
```diff
@@ -0,0 +1,179 @@
...
+            return positionIndex;
+        }
+
+        /*
+         * Run lint to extract the error range.
+         */
+        const messages = linter.verify(
+
+            // Remove trailing newline and presentational `⏎` characters
+            code.replace(/⏎(?=\n)/gu, "").replace(/\n$/u, ""),
```
fasttime, line 76: Shall we merge this expression with [this one](https://github.com/eslint/eslint/blob/5a8efd5b7ad13eb320a1f468d1d4ab3c8ab99214/docs/tools/markdown-it-rule-example.js#L75-L78) into a separate function, to ensure that they don't get out of sync if we make a change?

Alternatively, we could extend `addContentMustBeMarked` to accept `code` as a new argument [here](https://github.com/eslint/eslint/blob/94a090719015448b7ede6204225734020f02b152/docs/.eleventy.js#L198-L200), so we could store it to use it later.

docs/tools/prism-eslint-hook.js
```diff
@@ -0,0 +1,179 @@
...
+            { languageOptions: { sourceType: parserOptions.sourceType, parserOptions } },
+            { filename: "code.js" }
+        );
+
+        if (messages.some(m => m.fatal)) {
+
+            // ESLint fatal error.
+            return;
+        }
```
fasttime, line 85: I think it's fine to skip the fatals in `lines-around-comment`, but we should fix them in a separate PR.

docs/tools/prism-eslint-hook.js
ota-meshi: Prism.hooks only has code text in markdown, so I think it would be difficult to handle anything else as a key. For now, this is processed synchronously, so I don't think there is any problem with duplication. However, there may be problems after updating the document system 😓

docs/tools/prism-eslint-hook.js
```diff
@@ -0,0 +1,180 @@
...
+
+/**
+ * A set of content that needs to be marked.
+ * @type {Map<string, ParserOptions>}
+ */
+const contentsMustBeMarked = new Map();
+
+/**
+ * Add content that needs to be marked.
+ * @param {string} content Source code content that marks eslint errors.
```
amareshsm, line 18:
```suggestion
 * @param {string} content Source code content that marks ESlint errors.
```

docs/tools/prism-eslint-hook.js
fasttime: Every time execution hits this line, `contentsMustBeMarked` is empty, and every time execution hits the line with `contentsMustBeMarked.delete(env.code)` in the hook callback, the map has only one element: the content to be removed. So it seems that markdownIt is calling the highlighter immediately after calling the other plugin functions for the same block. If this is the case then we are safe: there can't be a duplication if there is just one potential match.

This makes me also think that probably we could just remember the last `content` and `options` found here instead of using a map. Do you think that would work?

ota-meshi: > This makes me also think that probably we could just remember the last `content` and `options` found here instead of using a map. Do you think that would work?

That's right. I think it works. I would replace Map :+1:

nzakas:
This is looking really nice. Is it possible to add a tooltip over the red-underlined code that includes the error message?

ota-meshi:
> Is it possible to add a tooltip over the red-underlined code that includes the error message?

Hmm. It seems difficult to do that with Prism's hooks for now.
It doesn't seem possible to add anything other than the `<span>` class and content for each Token. (For example, we cannot add the `popovertarget` attribute only to the error token.)
If we request the Prism team to pass the Token instance to Prism's `wrap` hook, we might be able to get a tooltip, but I'm not sure yet 🤔

ota-meshi:
> If we request the Prism team to pass the Token instance to Prism's wrap hook, we might be able to get a tooltip, but I'm not sure yet 🤔

I tried it but it didn't work well. During the `after-tokenize` hook and before, all Token instances created are replaced with new normalized Token instances before reaching the `wrap` hook.

nzakas:
Is it possible to add a `title` attribute to the `<span>`?

ota-meshi:
I don't know how to add `title` attribute (an attribute other than `class`) to a token using Prism's API 😓.
It might be possible to give a unique key to a `class` and later retrieve the message that maps to the unique key and set it to the `title`. But it may seem a bit like a hack.

ota-meshi:
> It might be possible to give a unique key to a class and later retrieve the message that maps to the unique key and set it to the title.

I tried that. If someone doesn't like it that way, we can change it back.
I also made major changes to `prism-eslint-hook.js` because we needed to organize the token structure more to display the `title` properly.

image

https://deploy-preview-18041--docs-eslint.netlify.app/rules/array-callback-return

docs/package.json
nzakas: We can add the top-level install to the workflow. :+1:

nzakas (commented):
This looks amazing. :heart: 

We can update the docs CI to install the top-level packages and I think this is ready to go.

docs/package.json
ota-meshi: I removed the `postinstall` and changed the workflow.
I think we probably need to modify the netlify script, but how should I do that?

docs/tools/prism-eslint-hook.js
fasttime: The fatals are fixed in #18078.

fasttime:
Nice work! LGTM when the build is fixed.

docs/tools/prism-eslint-hook.js
```diff
@@ -0,0 +1,334 @@
+"use strict";
+
+const Prism = require("prismjs");
+const { Linter } = require("../../lib/api");
+const astUtils = require("../../lib/shared/ast-utils");
```
mdjermanovic, line 5: How would this work for https://github.com/eslint/zh-hans.docs.eslint.org, where we don't have these modules?

ota-meshi: Oh... I don't think it will work as it is now.
@kecrily Do you have any ideas how to solve it?

ota-meshi: What do you think about changing this PR as below?

- `astUtils` is used only to get the regex, so it inlines the regex.
- Try `require("../../lib/api").Linter` and `require("eslint").Linter` and use the one that does not cause an error as the `Linter`.
- Change the install script of `zh-hans.docs.eslint.org` to install eslint as a dependency.

nzakas: Another option is to not do red underlines outside of the US site. We could check to see if these utilities exist, and if not, just don't do the linting.

snitin315: >Another option is to not do red underlines outside of the US site. We could check to see if these utilities exist, and if not, just don't do the lining.

I'd prefer this option.

ota-meshi: > Another option is to not do red underlines outside of the US site. We could check to see if these utilities exist, and if not, just don't do the linting.

I've made changes to this PR to make it so 👍

nzakas:
@christian-bromann looks like there's a problem with downloading the latest chromedriver.

docs/tools/prism-eslint-hook.js
```diff
@@ -0,0 +1,348 @@
+"use strict";
+
+const Prism = require("prismjs");
+const { docsExampleCodeToParsableCode } = require("./code-block-utils");
+
+let isAvailable = false;
+let Linter = null;
+let astUtils = null;
+
+try {
```
nzakas, line 10: Let's add a comment here explaining why we're testing.
```suggestion
/*
 * We can only do the syntax highlighting in the English-language
 * site because we need the Linter and astUtils. This same
 * code is run in translation sites where these utilities are
 * not available, so check to see if they are before attempting
 * to highlight the code.
 */
try {
```

docs/tools/prism-eslint-hook.js
```diff
@@ -0,0 +1,348 @@
+"use strict";
```
nzakas, line 1: Can you also add a fileoverview block on each new JS file?

christian-bromann:
> looks like there's a problem with downloading the latest chromedriver.

There was, however we fixed it on our side so it should be resolved with the latest WebdriverIO version.

nzakas:
I'll re-run now. Thanks!

nzakas (approved):
LGTM. Thanks! Would like another approval before merging.

fasttime:
Since ESLint dependencies are no longer installed when creating the rule docs pages, red underlines don't show up any more in the deploy preview (e.g. https://deploy-preview-18041--docs-eslint.netlify.app/rules/no-const-assign). Is this what we want?

ota-meshi:
I think we need to modify the netlify build script.
https://github.com/eslint/eslint/pull/18041#discussion_r1476814219

docs/package.json
nzakas: Hmm. There is no Netlify script. It just checks out `docs` and does `npm install`. 

I wonder, could we just list `"eslint": "file:../"` in `docs/package.json` and that will work?

ota-meshi: > I wonder, could we just list "eslint": "file:../" in docs/package.json and that will work?

I tried that locally but it doesn't seem to work. `eslint` dependencies were not installed 😓

ota-meshi: Do you think putting the `netlify.toml` and adding the command config will work fine?
https://docs.netlify.com/configure-builds/file-based-configuration/#build-settings

nzakas: Let's try it. I'm out of other ideas.

ota-meshi: I tried adding `netlify.toml` and it seems to be working fine :tada:

https://deploy-preview-18041--docs-eslint.netlify.app/rules/array-callback-return

image

nzakas (approved):
Looks like everything is working now. :tada:

Would like @mdjermanovic to verify his changes before merging.

docs/tools/prism-eslint-hook.js
```diff
@@ -0,0 +1,359 @@
...
+    /**
+     * A token type for marking the range reported by a rule.
+     * This is also used for the `class` attribute of `<span>`.
+     */
+    const TOKEN_TYPE_ESLINT_MARKED = "eslint-marked";
+
+    /**
+     * Use in the class attribute of `<span>` when an error is displayed in the line-feed, or BOM code.
+     */
+    const CLASS_ESLINT_MARKED_ON_ZERO_WIDTH = "eslint-marked-on-zero-width";
```
fasttime, line 70: In the Playground, the corresponding CSS class (`cm-lintPoint`) is also used to mark spans that are literally zero width, e.g. in this incorrect example for `unicode-bom` with option `always`: https://deploy-preview-18041--docs-eslint.netlify.app/rules/unicode-bom#always.

image

ota-meshi: Thank you for checking this PR!
I modified it so it looks the same as Playground :+1:

docs/src/assets/scss/syntax-highlighter.scss
```diff
@@ -99,6 +99,31 @@ pre[class*="language-"] {
...
+     * if we use it as is, we may see a shift in the border.
+     * To make the border shift less noticeable, draw it with a smaller width.
+     */
+    background-size: 4.4px auto;
+}
+
+.token.eslint-marked-on-zero-width {
+    border-left: 3px solid transparent;
+    border-right: 3px solid transparent;
+    border-bottom: 4px solid #d11;
```
fasttime, line 121: Maybe we can move these styles into an `::after` pseudoelement, so they don't cause the rest of the line to shift rightwards?

ota-meshi: I improved the style of the zero-width marker so that it appears more similar to playground 👍

fasttime: It's looking good now, thanks!

fasttime (approved):

snitin315 (approved):
This looks great ⭐ , thanks!

Leaving it open for @mdjermanovic

docs/src/assets/scss/syntax-highlighter.scss
```diff
@@ -99,6 +99,39 @@ pre[class*="language-"] {
...
+}
+
+.token.eslint-marked-on-zero-width {
+    position: relative;
+
+    /* Delete the wavy line. */
+    background: none;
+}
+
+.token.eslint-marked-on-zero-width::after {
```
mdjermanovic, line 125: For newlines, marks appear on the next line:

image

Maybe we should use `::before` instead of `::after`?

https://deploy-preview-18041--docs-eslint.netlify.app/rules/semi#always

ota-meshi: Thank you for finding the problem! I fixed it.

docs/tools/prism-eslint-hook.js
```diff
@@ -0,0 +1,360 @@
...
+            }
+
+            if (
+                env.content === "" ||
+                env.content === "\n" ||
+                env.content === "\r\n" ||
+                env.content === "\ufeff"
+            ) {
+                env.classes.push(CLASS_ESLINT_MARKED_ON_ZERO_WIDTH);
+            }
```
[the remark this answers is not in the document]
ota-meshi: I changed this PR to make it look like VSCode by extracting the newline code character by character into tokens 👍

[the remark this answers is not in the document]
mdjermanovic: Would it be possible to do this only for newline characters that appear in an error range and are either at the very start of the range or lone on the line?

image

image

nzakas (approved):

amareshsm:
nice 🎉

docs/tools/prism-eslint-hook.js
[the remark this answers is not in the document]
ota-meshi: I added CSS and changed it to look like that.

mdjermanovic (approved):
Great work, thanks!

PR #16612, opened 2022-12-04
docs: fix no-js dark theme bugs

amareshsm:
#### Prerequisites checklist

- [x] I have read the [contributing guidelines](https://github.com/eslint/eslint/blob/HEAD/CONTRIBUTING.md).

#### What is the purpose of this pull request? (put an "X" next to an item)

[ ] Documentation update
[ ] Bug fix ([template](https://raw.githubusercontent.com/eslint/eslint/HEAD/templates/bug-report.md))
[ ] New rule ([template](https://raw.githubusercontent.com/eslint/eslint/HEAD/templates/rule-proposal.md))
[ ] Changes an existing rule ([template](https://raw.githubusercontent.com/eslint/eslint/HEAD/templates/rule-change-proposal.md))
[ ] Add autofix to a rule
[ ] Add a CLI option
[ ] Add something to the core
[x] Other, please explain:
Fixed no-js (js disabled) dark theme bugs.

#### What changes did you make? (Give an overview)
When javascript is disabled and the prefers-color-scheme is dark 
    1. code block texts are not visible.
    2. Logo/tag text not inverted based on the theme.
    
| Before  | After |
| ------------- | ------------- |
|  image | image  |
| image  | image  | 
|  image |   image |
| image | image |
| image |  image |

PR #18071, opened 2024-02-01
docs: fix no-js dark theme bugs

amareshsm:
#### Prerequisites checklist

[ ] Documentation update
[ ] Add autofix to a rule
[ ] Add a CLI option
[ ] Add something to the core
[x] Other, please explain:
Fixed no-js (js disabled) dark theme bugs.

When javascript is disabled and the prefers-color-scheme is dark 
    1. code block texts are not visible.
    2. Logo/tag text not inverted based on the theme.
    
| Before  | After |
| ------------- | ------------- |
|   |   |

### DIFF
--- a/docs/src/assets/scss/no-js.scss
+++ b/docs/src/assets/scss/no-js.scss
@@ -1,0 +1,78 @@
+@media (prefers-color-scheme: dark) {
+    .no-js {
+        pre[class*="language-"] {
+            color: var(--color-neutral-100);
+        }
+
+        .token {
+            &.comment,
+            .prolog,
+            .doctype,
+            .cdata {
+                color: #8e9fae;
+            }
+        }
+
+        .logo-component {
+            fill: #fff;
+        }
+
+        #logo-center {
+            opacity: 0.6;
+        }
+
+        .rule__categories,
+        .rule__status {
+            background: var(--body-background-color);
+        }
+
+        .algolia-docsearch-suggestion--highlight {
+            background-color: var(--link-color);
+            color: var(--color-neutral-900);
+        }
+
+        .alert {
+            &.alert--warning {
+                color: var(--color-rose-300);
+                background-color: var(--color-rose-900);
+                border: 1px solid var(--color-rose-300);
+                .alert__type {
+                    color: var(--color-rose-200);
+                }
+            }
+
+            &.alert--important {
+                color: var(--color-warning-300);
+                background-color: var(--color-warning-900);
+                border: 1px solid var(--color-warning-300);
+                .alert__type {
+                    color: var(--color-warning-200);
+                }
+            }
+
+            &.alert--tip {
+                color: var(--color-success-300);
+                background-color: var(--color-success-900);
+                border: 1px solid var(--color-success-300);
+                .alert__type {
+                    color: var(--color-success-200);
+                }
+            }
+        }
+
+        .alert__type,
+        .alert__learn-more {
+            .alert--warning & {
+                color: var(--color-rose-200);
+            }
+
+            .alert--important & {
+                color: var(--color-warning-200);
+            }
+
+            .alert--tip & {
+                color: var(--color-success-200);
+            }
+        }
+    }
+}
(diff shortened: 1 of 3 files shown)


kecrily (approved):
LGTM. But I remember some of the problems we were supposed to have fixed at one point, I don't know if it's an illusion or not 🤷

amareshsm:
> LGTM. But I remember some of the problems we were supposed to have fixed at one point, I don't know if it's an illusion or not 🤷

Yes, PR was raised but not merged it was not completely done at that time.
https://github.com/eslint/eslint/pull/16612

docs/src/assets/scss/no-js.scss
JoshuaKGoldberg, line 1: [Refactor] This adds in a new set of styles that only get hit in dark mode, and may only be tested in production by users with JavaScript disabled. Even if many users are visiting the site with JS disabled, that's still an extra source of truth for these comments. IMO it's not great having an additional (so, at least 2 total) place to manage these styles.

Instead, would it be feasible to figure out why some of the existing styles aren't loading in? E.g. if dark-mode-only styles are only loaded by JS, maybe they can be loaded in with pure HTML/CSS instead?



Sorry if I missed a previous explanation. I looked through the past PR comments and didn't see anything. I also filed #18076 for tracking.

JoshuaKGoldberg (changes requested):
The functionality looks good to me on the preview deploy, thanks for getting this started up again! ✨ 

But, requesting changes that I think there might be a strategy that avoids code duplication.

docs/src/assets/scss/no-js.scss
amareshsm: Yes, this added a new set of styles that are only used when js is disabled and the preferred theme is dark. 

An additional place to these styles I am getting it? do you mean a new file specifically for handling no-js mode?

`
Instead, would it be feasible to figure out why some of the existing styles aren't loading in? E.g. if dark-mode-only styles are only loaded by JS, maybe they can be loaded in with pure HTML/CSS instead?
`

There is no existing style targeting drak mode with js disabled which is why it falls back to default styles.  The styles that are not applied are related to JS.

For example:
let's take the ESLint logo - it is an SVG element. The default style works well in light mode. 
   
https://github.com/eslint/eslint/blob/aadfb609f1b847e492fc3b28ced62f830fe7f294/docs/src/_includes/components/logo.html#L11-L20

For the dark mode styles specified explicitly based on the data-theme attribute (which gets added based on the user choice)
https://github.com/eslint/eslint/blob/aadfb609f1b847e492fc3b28ced62f830fe7f294/docs/src/_includes/components/logo.html#L2-L10

If JavaScript is turned off and the chosen theme is light, there is no problem as the default logo functions correctly. However, when the preferred mode is dark, the absence of styles for this scenario results in a fallback to the light mode style, leading to the issue. To resolve this, I have specifically included styles that target the dark theme when JavaScript is disabled.


Additionally, let's examine the behaviour of various elements in a dark theme when JavaScript is disabled. The code block background colour toggles flawlessly between dark and light modes even without JavaScript, Because property value is not hard-coded

https://github.com/eslint/eslint/blob/b844324e4e8f511c9985a96c7aca063269df9570/docs/src/assets/scss/syntax-highlighter.scss#L48-L50 

Variable `--lightest-background-color` value is toggled based on the theme

light theme:
https://github.com/eslint/eslint/blob/b844324e4e8f511c9985a96c7aca063269df9570/docs/src/assets/scss/tokens/themes.scss#L129

dark theme:
https://github.com/eslint/eslint/blob/b844324e4e8f511c9985a96c7aca063269df9570/docs/src/assets/scss/tokens/themes.scss#L152

preferred theme dark and js disabled:
https://github.com/eslint/eslint/blob/b844324e4e8f511c9985a96c7aca063269df9570/docs/src/assets/scss/tokens/themes.scss#L108

Preferred them light and js disabled:

https://github.com/eslint/eslint/blob/b844324e4e8f511c9985a96c7aca063269df9570/docs/src/assets/scss/tokens/themes.scss#L87

Using variables we can solve this issue. Refer: https://github.com/eslint/eslint/pull/18077

JoshuaKGoldberg: This makes sense, thanks! +1 to #18077's approach, I'll approve over there.

Issue #16620, opened 2022-12-06
 Standardize auto-labeling 


amareshsm:
### What problem do you want to solve?

Auto labeling Pull Request - Currently, we are using the ESLint bot to add labels based on the commit message. 

### What do you think is the correct solution?

In some cases the bot is not required we can use [labeler](https://github.com/actions/labeler) to automatically label new pull requests based on the paths of files being changed.

Eg: 
1. Adding a `documentation` label if any file changes in the docs folder.
2. Adding a `rule` label if any file changes in the rule folder.

We can check where else we can use the labeler. For the remaining cases, we can use the ESLint bot itself.

### Participation

- [X] I am willing to submit a pull request for this change.

### Additional comments

Thoughts?

nzakas:
I'm okay with this. Are there any other labels we can add in this way?

snitin315:
I'm ok with this change too. I believe the following labels can be added:
- `documentation`: changes in `docs/**`
- `rule`: changes in `lib/rules/*.js`
-  `cli`: changes in `lib/cli.js` `lib/cli-engine`
-  `core`: changes in `lib/{linter,eslint,configs}`
- `formatter`: changes in `lib/cli-engine/formatters`

harish-sethuraman:
I think we didn't want this for documentation?

https://github.com/eslint/eslint/pull/16615#discussion_r1041850113

nzakas:
@amareshsm are you still interested in pushing this forward?

amareshsm:
> @amareshsm are you still interested in pushing this forward?

I missed this one thanks for bumping up. I will look into this soon.

PR #18109, opened 2024-02-10
ci: add PR labeler action

snitin315:
#### Prerequisites checklist

[ ] Documentation update
[ ] Add autofix to a rule
[ ] Add a CLI option
[ ] Add something to the core
[x] Other, please explain: 

Fix https://github.com/eslint/eslint/issues/16620

> [Notes regarding `pull_request_target` event](https://github.com/actions/labeler?tab=readme-ov-file#notes-regarding-pull_request_target-event). It requires the workflow file to be present at the main branch.

For testing purposes, you can play create PRs in my forked repo - https://github.com/snitin315/eslint/pulls

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -13,6 +13,11 @@ cli:
   - changed-files:
     - any-glob-to-any-file: ['lib/cli.js', 'lib/options.js', 'lib/cli-engine/**']
 
+ESLint:
+- any:
+  - changed-files:
+    - any-glob-to-any-file: ['lib/eslint/**']
+
 core:
 - any:
   - changed-files:
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -28,7 +28,7 @@ formatter:
   - changed-files:
     - any-glob-to-any-file: ['lib/cli-engine/formatters/**']
 
-github_actions:
+"github actions":
 - any:
   - changed-files:
     - any-glob-to-any-file: ['.github/workflows/**']
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -1,0 +1,30 @@
+documentation:
+- any:
+  - changed-files:
+    - all-globs-to-all-files: ['docs/**']
+    - any-globs-to-any-files: ['docs/**', '!lib/rules/**']
+
+rule:
+- any:
+  - changed-files:
+    - any-globs-to-any-files: ['lib/rules/**']
+
+cli:
+- any:
+  - changed-files:
+    - any-globs-to-any-files: ['lib/cli.js', 'lib/options.js', 'lib/cli-engine/**']
+
+core:
+- any:
+  - changed-files:
+    - any-globs-to-any-files: ['lib/{config,eslint,linter,rule-tester,source-code}/**', 'lib/api.js']
+
+formatter:
+- any:
+  - changed-files:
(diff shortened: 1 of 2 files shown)


.github/labeler.yml
```diff
@@ -0,0 +1,29 @@
...
+
+rule:
+- any:
+  - changed-files:
+    - any-glob-to-any-file: ['lib/rules/**']
+
+cli:
+- any:
+  - changed-files:
+    - any-glob-to-any-file: ['lib/cli.js', 'lib/options.js', 'lib/cli-engine/**']
```
nzakas, line 14: We probably also want `lib/eslint/**` to capture `ESLint` changes.

.github/labeler.yml
```diff
@@ -0,0 +1,29 @@
...
+- any:
+  - changed-files:
+    - any-glob-to-any-file: ['lib/{config,eslint,linter,rule-tester,source-code}/**', 'lib/api.js']
+
+formatter:
+- any:
+  - changed-files:
+    - any-glob-to-any-file: ['lib/cli-engine/formatters/**']
+
+github_actions:
```
nzakas, line 26: Maybe to avoid being the only label with an underscore?

```suggestion
"github actions":
```

nzakas (changes requested):
Thanks for looking into this! Just a few suggestions.

.github/labeler.yml
snitin315: Actually, this label [already exists](https://github.com/eslint/eslint/pulls?q=is%3Apr+label%3Agithub_actions+is%3Aclosed), but maybe we can update it for depandabot too

nzakas: We aren't using dependabot anymore, we're using Renovate, so we should make sure that will use the same label.

.github/labeler.yml
```diff
@@ -0,0 +1,29 @@
+documentation:
```
harish-sethuraman, line 1: we do already have the same label coming from our github bot. Should we do this from both the places or keep one alive?

snitin315: The bot attaches label as per the PR title, I think it's fine to keep it at both places.

.github/labeler.yml
snitin315: Added 👍🏻

nzakas: Sorry, I meant that `lib/eslint/**` should be under `cli` because that really is the driver of the CLI.

.github/labeler.yml
```diff
@@ -0,0 +1,34 @@
...
+- any:
+  - changed-files:
+    - any-glob-to-any-file: ['lib/rules/**']
+
+cli:
+- any:
+  - changed-files:
+    - any-glob-to-any-file: ['lib/cli.js', 'lib/options.js', 'lib/cli-engine/**']
+
+ESLint:
```
nzakas, line 16: I don't think we need this as long as we the files labeled `core`

nzakas (changes requested):

.github/labeler.yml
snitin315: Oh! Updated 👍🏻

nzakas (approved):
LGTM. Let's try it out and see how it goes. :)

PR #18198, opened 2024-03-13
docs: Clarify the description of `sort-imports` options

gwBear1:
#### Prerequisites checklist

- [x] Documentation update

스크린샷 2024-03-13 오전 10 51 55

---

Made 4 test cases of the below. It could be more straightforward.
```
incorrect with default option ignore false
correct   with default option ignore false
correct   with ignore true
incorrect with ignore true
```
- ignoreCase
- ignoreDeclarationSort
- ignoreMemberSort

- Should be checked whether test cases are correct.

### DIFF
--- a/docs/src/rules/sort-imports.md
+++ b/docs/src/rules/sort-imports.md
@@ -193,29 +193,53 @@ import {b, a, c} from 'foo.js'
 ### `ignoreCase`
 
 When `true` the rule ignores the case-sensitivity of the imports local name.
+Uppercase letters of the alphabet must always precede lowercase letters.
 
-Examples of **incorrect** code for this rule with the `{ "ignoreCase": false }` option:
+Examples of **incorrect** code for this rule with the default `{ "ignoreCase": false }` option:
 
 ::: incorrect
 
 ```js
 /*eslint sort-imports: ["error", { "ignoreCase": false }]*/
-
-import B from 'foo.js';
 import a from 'bar.js';
+import B from 'foo.js';
+import c from 'baz.js';
 ```
 
 :::
 
-Examples of **correct** code for this rule with the `{ "ignoreCase": true }` option:
+Examples of **correct** code for this rule with the default `{ "ignoreCase": false }` option:
+
+::: correct
+
+```js
+/*eslint sort-imports: ["error", { "ignoreCase": false }]*/
+import B from 'bar.js';
+import a from 'foo.js';
+import c from 'baz.js';
+```
+
+:::
+
+Examples of **correct** code for this rule with `{ "ignoreCase": true }` option:
 
 ::: correct
 
 ```js
 /*eslint sort-imports: ["error", { "ignoreCase": true }]*/
+import a from 'bar.js';
+import B from 'foo.js';
+import c from 'baz.js';
+```
 
-import a from 'foo.js';
+:::
+
+::: correct
+
+```js
+/*eslint sort-imports: ["error", { "ignoreCase": true }]*/
 import B from 'bar.js';
+import a from 'foo.js';
 import c from 'baz.js';
 ```
 
@@ -235,6 +259,18 @@ Examples of **incorrect** code for this rule with the default `{ "ignoreDeclarat
 /*eslint sort-imports: ["error", { "ignoreDeclarationSort": false }]*/
 import b from 'foo.js'
 import a from 'bar.js'
+```
+
+:::
+
+Examples of **correct** code for this rule with the default `{ "ignoreDeclarationSort": false }` option:
+
+::: correct
+
+```js
+/*eslint sort-imports: ["error", { "ignoreDeclarationSort": false }]*/
+import a from 'bar.js'
+import b from 'foo.js'
 ```
 
 :::
@@ -278,6 +314,17 @@ import {b, a, c} from 'foo.js'
 
 :::
 
+Examples of **correct** code for this rule with the default `{ "ignoreMemberSort": false }` option:
+
+::: incorrect
+
+```js
+/*eslint sort-imports: ["error", { "ignoreMemberSort": false }]*/
+import {a, b, c} from 'foo.js'
+```
+
+:::
+
 Examples of **correct** code for this rule with the `{ "ignoreMemberSort": true }` option:
 
 ::: correct
@@ -285,6 +332,15 @@ Examples of **correct** code for this rule with the `{ "ignoreMemberSort": true 
 ```js
 /*eslint sort-imports: ["error", { "ignoreMemberSort": true }]*/
 import {b, a, c} from 'foo.js'
+```
+
+:::
+
+::: correct
+
+```js
+/*eslint sort-imports: ["error", { "ignoreMemberSort": true }]*/
+import {a, b, c} from 'foo.js'
 ```
 
 :::
--- a/docs/src/rules/sort-imports.md
+++ b/docs/src/rules/sort-imports.md
@@ -192,8 +192,6 @@ import {b, a, c} from 'foo.js'
 
 ### `ignoreCase`
 
-Default is `false`.
-
 When `false` (default), uppercase letters of the alphabet must always precede lowercase letters.
 
 When `true`, the rule ignores the case-sensitivity of the imports local name.
@@ -251,9 +249,7 @@ import a from 'bar.js';
 
 ### `ignoreDeclarationSort`
 
-Default is `false`.
-
-Ignores the sorting of import declaration statements.
+When `true`, the rule ignores the sorting of import declaration statements. Default is `false`.
 
 Examples of **incorrect** code for this rule with the default `{ "ignoreDeclarationSort": false }` option:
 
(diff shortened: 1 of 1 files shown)


Tanujkanti4441:
Hi @gwBear1, thanks for the PR, but the docs are correct because it shows both incorrect and correct examples when 
`{ "ignoreCase": true }` as you can see the examples written under the `incorrect` section by opening them in playground.

Tanujkanti4441:
sorry for closing too early.
after looking around a bit i found
```js
/*eslint sort-imports: "error"*/

import A from 'foo.js';
import b from 'bar.js';
```
by default rule doesn't report this arrangement of import names.
```js
/*eslint sort-imports: "error"*/

import a from 'foo.js';
import B from 'bar.js';
```
but report this

@mdjermanovic is it intended

if yes then we need to show this in docs too

gwBear1:
@Tanujkanti4441 Yeah I was confused too. You right, fixing test case would be better choice. I'll rewrite test cases.

gwBear1:
@Tanujkanti4441 And I also think whole document's test case is written wrong way. For example in ignoreDeclarationSort...

스크린샷 2024-03-13 오후 12 48 33

Testing with `{ "ignoreDeclarationSort": true }` is a very useless use case...

I'll update the whole section.

gwBear1:
I added test cases more straightforward. Each options has 4 test cases
- incorrect with default option `ignore false`
- correct with default option `ignore false`
- correct with `ignore true`
- correct2 with `ignore true`

However, I think `ignore true` cases are not useful examples. Those could be deleted if you guys want.

docs/src/rules/sort-imports.md
```diff
@@ -239,18 +252,20 @@ import a from 'bar.js'
...
 
 ::: correct
 
 ```js
-/*eslint sort-imports: ["error", { "ignoreDeclarationSort": true }]*/
-import a from 'foo.js'
-import b from 'bar.js'
+/*eslint sort-imports: ["error", { "ignoreDeclarationSort": false }]*/
+import a from 'bar.js'
+import b from 'foo.js'
```
Tanujkanti4441, line 262:
```suggestion
import a from 'bar.js';
import b from 'foo.js';
```

there are other examples that don't have semicolon can you please add semicolon to them too?

and also i think its better if we add an incorrect section for `{ "ignoreDeclarationSort": true }` with these examples that shows multiple members import are not allowed by this option:
```js
import {b, a, c} from 'foo.js';
```

docs/src/rules/sort-imports.md
```diff
@@ -278,6 +293,17 @@ import {b, a, c} from 'foo.js'
 
 :::
 
+Examples of **correct** code for this rule with the default `{ "ignoreMemberSort": false }` option:
+
+::: correct
+
+```js
+/*eslint sort-imports: ["error", { "ignoreMemberSort": false }]*/
+import {a, b, c} from 'foo.js'
```
Tanujkanti4441, line 302:
```suggestion
import {a, b, c} from 'foo.js';
```

and an `incorrect` section can also be added for { "ignoreMemberSort": true } with following example:
```js
import b from 'foo.js';
import a from 'bar.js';
```

docs/src/rules/sort-imports.md
```diff
@@ -193,29 +193,42 @@ import {b, a, c} from 'foo.js'
...
-import a from 'foo.js';
+/*eslint sort-imports: ["error", { "ignoreCase": false }]*/
 import B from 'bar.js';
+import a from 'foo.js';
+import c from 'baz.js';
+```
+
+:::
+
+Examples of **correct** code for this rule with `{ "ignoreCase": true }` option:
```
Tanujkanti4441, line 224: its fine to keep the `incorrect` example for `{ "ignoreCase": true } ` that shows which case is not allowed by the option.

Tanujkanti4441 (changes requested):

docs/src/rules/sort-imports.md
gwBear1: Good suggestion. Thanks for the review!

gwBear1:
@Tanujkanti4441 Can you check my PR again?

docs/src/rules/sort-imports.md
```diff
@@ -193,34 +193,59 @@ import {b, a, c} from 'foo.js'
 ### `ignoreCase`
 
 When `true` the rule ignores the case-sensitivity of the imports local name.
+Uppercase letters of the alphabet must always precede lowercase letters.
```
Tanujkanti4441, line 196:
```suggestion
When `true` the rule ignores the case-sensitivity of the imports local name.

Uppercase letters of the alphabet must always precede lowercase letters with default settings.
```

better to clear this

Tanujkanti4441 (changes requested):

mdjermanovic:
> ```js
> /*eslint sort-imports: "error"*/
> 
> import A from 'foo.js';
> import b from 'bar.js';
> ```
> 
> 
>     
>       
>     
> 
>       
>     
> 
>     
>   
> by default rule doesn't report this arrangement of import names.
> ```js
> /*eslint sort-imports: "error"*/
> 
> import a from 'foo.js';
> import B from 'bar.js';
> ```
> 
> 
>     
>       
>     
> 
>       
>     
> 
>     
>   
> but report this
> @mdjermanovic is it intended?
> ```

This is intended because: any uppercase letter < any lowercase letter. A < b, B < a.

mdjermanovic:
I'm confused by the PR title. "Fixed incorrect doc description" sounds like the docs were incorrect, but it looks like the docs were correct and we just want to add more examples to further clarify options?

docs/src/rules/sort-imports.md
```diff
@@ -194,33 +194,59 @@ import {b, a, c} from 'foo.js'
 
 When `true` the rule ignores the case-sensitivity of the imports local name.
 
-Examples of **incorrect** code for this rule with the `{ "ignoreCase": true }` option:
+Uppercase letters of the alphabet must always precede lowercase letters with default settings.
```
mdjermanovic, line 197: If we want to describe both, I think the default behavior should be first.

```md
When `false` (default), uppercase letters of the alphabet must always precede lowercase letters.

When `true`, the rule ignores the case-sensitivity of the imports local name.
```

gwBear1 (commented):
fix ignoreCase description

gwBear1:
> I'm confused by the PR title. "Fixed incorrect doc description" sounds like the docs were incorrect, but it looks like the docs were correct and we just want to add more examples to further clarify options?

I changed the PR title and description

gwBear1:
@Tanujkanti4441 @mdjermanovic Followed your comments. Could you check this PR again?

docs/src/rules/sort-imports.md
```diff
@@ -192,35 +192,61 @@ import {b, a, c} from 'foo.js'
...
+
+```js
+/*eslint sort-imports: ["error", { "ignoreCase": true }]*/
+import B from 'foo.js';
+import a from 'bar.js';
+```
+
+:::
+
 Default is `false`.
```
Tanujkanti4441, line 250: The default values are mentioned in the bottom of the option's section i feel this info should be on top with option's description can you shift them to the top for every option?

Tanujkanti4441 (commented):
everything looks good just a last thing.

docs/src/rules/sort-imports.md
gwBear1: I'm checking how other rules describe options but all rules have different formats 🤔. Anyway in this document, I'll address your comment.

---
1. `prefer-arrow-callback` - (allowUnboundThis)
They are telling default option as a sentence.
스크린샷 2024-03-21 오후 3 37 04

2. `no-unused-expressions` (allowTernary)
They doesn't talk about default option. It's just defined at the top of the document.
스크린샷 2024-03-21 오후 3 36 50

docs/src/rules/sort-imports.md
```diff
@@ -289,10 +341,22 @@ import {b, a, c} from 'foo.js'
...
+/*eslint sort-imports: ["error", { "ignoreMemberSort": true }]*/
+import b from 'foo.js';
+import a from 'bar.js';
+```
+
+:::
 
 ### `memberSyntaxSortOrder`
 
+Default is `["none", "all", "multiple", "single"]`.
```
Tanujkanti4441, line 358:
```suggestion
This option takes an array with four predefined elements, the order of elements specifies the order of import styles.

Default order is `["none", "all", "multiple", "single"]`.
```

Tanujkanti4441 (changes requested):

docs/src/rules/sort-imports.md
```diff
@@ -341,10 +405,10 @@ import {a, b} from 'foo.js';
 
 :::
 
-Default is `["none", "all", "multiple", "single"]`.
-
 ### `allowSeparatedGroups`
 
+Default is `false`.
+
 When `true` the rule checks the sorting of import declaration statements only for those that appear on consecutive lines.
```
Tanujkanti4441, line 412:
```suggestion
When `true`, the rule checks the sorting of import declaration statements only for those that appear on consecutive lines. Default is `false`.
```

Tanujkanti4441 (changes requested):

docs/src/rules/sort-imports.md
Tanujkanti4441: Ya! but there is not a particular format for the description, only concept must be clear to the readers.

gwBear1:
@Tanujkanti4441 I hope this would be the last request 😅 Thanks for reviewing be thorough

Tanujkanti4441 (approved):
LGTM! thanks.

mdjermanovic (approved):
LGTM, thanks!

PR #18256, opened 2024-04-01
feat: Show config names in error messages

### LINKED ISSUE #18231: Support `name` field in flat config and recommend as an convention

nzakas:
#### Prerequisites checklist

[ ] Documentation update
[ ] Add autofix to a rule
[ ] Add a CLI option
[x] Add something to the core
[ ] Other, please explain:

Updated `@humanwhocodes/config-array` so that we see config names in config validations errors.

fixes #18231
refs #18252

### DIFF
--- a/package.json
+++ b/package.json
@@ -69,7 +69,7 @@
     "@eslint-community/regexpp": "^4.6.1",
     "@eslint/eslintrc": "^3.0.2",
     "@eslint/js": "9.0.0-rc.0",
-    "@humanwhocodes/config-array": "^0.11.14",
+    "@humanwhocodes/config-array": "^0.12.1",
     "@humanwhocodes/module-importer": "^1.0.1",
     "@nodelib/fs.walk": "^1.2.8",
     "ajv": "^6.12.4",


mdjermanovic (approved):
LGTM, thanks!

PR #18306, opened 2024-04-09
fix: --inspect-config only for flat config and respect -c

nzakas:
#### Prerequisites checklist

[ ] Documentation update
[ ] Add autofix to a rule
[ ] Add a CLI option
[ ] Add something to the core
[ ] Other, please explain:

### Environment

Environment Info:

Node version: v20.11.0
npm version: v10.4.0
Local ESLint version: v9.0.0 (Currently used) 

Espree

### What did you do?

1. Ran `--inspect-config` using `.eslintrc.js`
2. Ran `--inspect-config` with `-c other.config.js`

1. ESLint should report an error and not open config inspector because this command is not supported.
2. ESLint should open config inspector with the file specified by `-c`

### What actually happened?

1. Config inspector opened with an error message.
2. Config inspector opened with the default `eslint.config.js` file.

1. Moved `--inspect-config` resolutions into `cli`
2. Calculate the correct config file to pass to config inspector before calling it.

### DIFF
--- a/lib/cli.js
+++ b/lib/cli.js
@@ -448,12 +448,6 @@ const cli = {
 
         if (options.inspectConfig) {
 
-            // we can only use this option with flat config
-            if (!usingFlatConfig) {
-                log.error("The --inspect-config option is only available when using an eslint.config.js file.");
-                return 2;
-            }
-
             log.info("You can also run this command directly using 'npx @eslint/config-inspector' in the same directory as your configuration file.");
 
             try {
--- a/lib/cli.js
+++ b/lib/cli.js
@@ -19,7 +19,7 @@ const fs = require("fs"),
     path = require("path"),
     { promisify } = require("util"),
     { LegacyESLint } = require("./eslint"),
-    { ESLint, shouldUseFlatConfig } = require("./eslint/eslint"),
+    { ESLint, shouldUseFlatConfig, locateConfigFileToUse } = require("./eslint/eslint"),
     createCLIOptions = require("./options"),
     log = require("./shared/logging"),
     RuntimeInfo = require("./shared/runtime-info"),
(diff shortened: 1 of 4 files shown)


aladdin-add:
> What did you do?
> 1. Ran --inspect-config using .eslintrc.js

flat config has been enabled by default since v9. how did you using `.eslintrc.js`?

here is what i'm getting using `ESLINT_USE_FLAT_CONFIG=false`:

```bash
➜  eslint git:(config-inspector-fix) ESLINT_USE_FLAT_CONFIG=false ./bin/eslint.js --inspect-config 
Invalid option '--inspect-config' - perhaps you meant '--inline-config'?
```

nzakas:
Here's what I get:

```
$ ESLINT_USE_FLAT_CONFIG=false npx eslint --inspect-config
You can also run this command directly using 'npx @eslint/config-inspector' in the same directory as your configuration file.
ℹ Starting ESLint config inspector at http://127.0.0.1:7777 
```

When in eslintrc mode, it should not start config inspector.

aladdin-add:
I'm seeing the same output when using eslint v9.0.0. What I had in my last comment was the output at the current branch and it doesn't seem to be the expected result?

nzakas:
The output you mention in https://github.com/eslint/eslint/pull/18306#issuecomment-2046584306 is what you should get with the change in this pull request.

aladdin-add:
fine, I see. we are on the same page now!

mdjermanovic (dismissed):
LGTM, thanks! Leaving open for a few days for others to review.

lib/cli.js
```diff
@@ -425,6 +446,30 @@ const cli = {
             return 0;
         }
 
+        if (options.inspectConfig) {
+
+            // we can only use this option with flat config
+            if (!usingFlatConfig) {
+                log.error("The --inspect-config option is only available when using an eslint.config.js file.");
```
aladdin-add, line 453: the actual result is:
```bash
Invalid option '--inspect-config' - perhaps you meant '--inline-config'?
```

seems the code was never excuted?

nzakas: Oops, good call. Forgotten during refactoring.

aladdin-add (approved):
LGTM.

PR #18346, opened 2024-04-15
feat: Add `ignore-directives` option to `require-await`

jorgezreik:
#### Prerequisites checklist

[ ] Documentation update
[ ] Add autofix to a rule
[ ] Add a CLI option
[ ] Add something to the core
[ ] Other, please explain:

`require-await`

[ ] Generate more warnings
[X] Generate fewer warnings
[ ] Implement autofix
[ ] Implement suggestions

[X] A new option
[ ] A new default behavior
[ ] Other

```js
/*eslint require-await: ["error", { "ignoreDirectives": ["use server"] }]*/

async function handler() {
    "use server"
    console.log("Handled")
}

async function handler() {
    'use server'
    console.log("Handled")
}
```

It will flag the `async` function as requiring an `await` operator.

With the `ignoreDirectives` set to `["use server"]`, no errors/warning will be thrown for these functions.

Added an option to `require-await` to ignore functions with the given directives. This is useful for frameworks that require functions with certain directives to always be asynchronous. The biggest example of this is the upcoming React `"use server"` changes which require all functions with the `"use server"` directive to be async.

The code itself is simple! I would just like to point out the benefits this option provides in allowing better error behavior around certain framework requirements while not being prescriptive with what those requirements are. I.E., this change allows users to abide by React `"use server"` requirements, but it would also support similar features in other systems just as well.

### DIFF
--- a/docs/src/rules/require-await.md
+++ b/docs/src/rules/require-await.md
@@ -77,6 +77,69 @@ async function noop() {}
 
 :::
 
+## Options
+
+This rule has an object option, with one option:
+
+* `"ignoreDirectives"` - an optional array of directive values whose containing functions this rule will ignore.
+
+### ignore
+
+You can specify multiple directive values to ignore in the `"ignoreDirectives"` array.
+
+For example, a framework might require functions tagged with the `"use server"` directive to always be async. In this case, they should be async even if they don't contain an `await` operator. You can include `"use server"` in the `"ignoreDirectives"` array to make this rule ignore such functions.
+
+Examples of **incorrect** code for this rule with the `"ignoreDirectives"` option:
+
+::: incorrect
+
+```js
+/*eslint require-await: ["error", { "ignore": ["use server"] }]*/
+
+async function handler() {
+    console.log("Handled")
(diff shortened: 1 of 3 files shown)


nzakas:
Thanks for the suggestion, but this doesn't make sense to go in the core. First, `"use server"` is not actually a directive even though React may be using it as such. The only real directive is `"use strict"`, so we can't make general assumptions about anything else in the same format.

If you need this for a React project, then I'd suggest you create a custom rule for that purpose, or else look to `eslint-plugin-react` for how best to handle this.

jorgezreik:
> Thanks for the suggestion, but this doesn't make sense to go in the core. First, `"use server"` is not actually a directive even though React may be using it as such. The only real directive is `"use strict"`, so we can't make general assumptions about anything else in the same format.
> 
> If you need this for a React project, then I'd suggest you create a custom rule for that purpose, or else look to `eslint-plugin-react` for how best to handle this.

Thank you for the feedback! I'll look into other ways to get this behavior going for React projects :)

PR #18352, opened 2024-04-16
feat: add suggestions to `no-unused-vars`

Tanujkanti4441:
#### Prerequisites checklist

[ ] Documentation update
[ ] Add autofix to a rule
[ ] Add a CLI option
[ ] Add something to the core
[ ] Other, please explain:

added suggestion to `no-unused-vars` rule.

Fixes: #17545

### DIFF
--- a/lib/rules/no-unused-vars.js
+++ b/lib/rules/no-unused-vars.js
@@ -1152,28 +1152,8 @@ module.exports = {
                 return null;
             }
 
-            // skip fix when variable is reassigned
-            if (allWriteReferences.length > 1 && parentType !== "AssignmentPattern") {
-                return null;
-            }
-
-            // skip fix when variable in assignment pattern is reassigned
-            if (allWriteReferences.length > 2 && parentType === "AssignmentPattern") {
-                return null;
-            }
-
-            // skip fix when initialized variable is reassigned
-            if (allWriteReferences.length > 0 && parent.init === null && parentType !== "AssignmentPattern") {
-                return null;
-            }
-
-            // skip fix when function parameters are ressaigned
-            if (isFunction(parent) && allWriteReferences.length > 0 && parent.params.includes(id)) {
-                return null;
-            }
-
-            // skip fix when function parameters with default value are reassigned
-            if (isFunction(parent.parent) && parentType === "AssignmentPattern" && allWriteReferences.length > 1) {
+            // skip fix when variable has references that would be left behind
+            if (allWriteReferences.some(ref => ref.identifier.range[0] !== id.range[0])) {
                 return null;
             }
 
--- a/lib/rules/no-unused-vars.js
+++ b/lib/rules/no-unused-vars.js
@@ -1007,7 +1007,7 @@ module.exports = {
                 if (getTokenBeforeValue(node) === ":") {
                     if (parentNode.parent.type === "ObjectPattern") {
                         // eslint-disable-next-line no-use-before-define -- due to interdependency of functions
-                        return fixObjectWithValueSeperator(node);
+                        return fixObjectWithValueSeparator(node);
                     }
                 }
 
@@ -1100,7 +1100,7 @@ module.exports = {
              * @param {ASTNode} node parent node to check
              * @returns {Object} fixer object
              */
-            function fixObjectWithValueSeperator(node) {
+            function fixObjectWithValueSeparator(node) {
                 const parentNode = node.parent.parent;
 
                 // fix cases like [{a : { b }}] or [{a : [ b ]}]
@@ -1248,7 +1248,7 @@ module.exports = {
                 }
             }
 
-            // remove unused varibales inside an array
+            // remove unused variables inside an array
             if (parentType === "ArrayPattern") {
                 if (hasSingleElement(parent)) {
 
@@ -1478,7 +1478,7 @@ module.exports = {
                                         varName: unusedVar.name
                                     },
                                     fix(fixer) {
-                                        return handleFixes(fixer, unusedVar, writeReferences);
+                                        return handleFixes(fixer, unusedVar);
                                     }
                                 }
                             ]
--- a/lib/rules/no-unused-vars.js
+++ b/lib/rules/no-unused-vars.js
@@ -952,12 +952,14 @@ module.exports = {
 
                         // if next token is a string it could become a directive if node is removed -> no suggestion.
                         const nextToken = sourceCode.getTokenAfter(parentNode.parent);
+
                         if (nextToken?.type === "String") {
                             return null;
                         }
 
                         // if previous token exists and is not ";" or "{" not sure about ASI rules -> no suggestion.
                         const prevToken = sourceCode.getTokenBefore(parentNode.parent);
+
                         if (prevToken && !astUtils.isSemicolonToken(prevToken) && !astUtils.isOpeningBraceToken(prevToken)) {
                             return null;
                         }
@@ -994,6 +996,20 @@ module.exports = {
 
                     // remove complete declaration when there is an unused variable in 'const { a } = foo;', same for arrays.
                     if (parentNode.parent.declarations.length === 1) {
+                        // if next token is a string it could become a directive if node is removed -> no suggestion.
+                        const nextToken = sourceCode.getTokenAfter(parentNode.parent);
+
+                        if (nextToken?.type === "String") {
+                            return null;
+                        }
+
+                        // if previous token exists and is not ";" or "{" not sure about ASI rules -> no suggestion.
+                        const prevToken = sourceCode.getTokenBefore(parentNode.parent);
+
+                        if (prevToken && !astUtils.isSemicolonToken(prevToken) && !astUtils.isOpeningBraceToken(prevToken)) {
+                            return null;
+                        }
+
                         return fixer.removeRange(parentNode.parent.range);
                     }
 
@@ -1189,12 +1205,14 @@ module.exports = {
 
                     // if next token is a string it could become a directive if node is removed -> no suggestion.
                     const nextToken = sourceCode.getTokenAfter(parent.parent);
+
                     if (nextToken?.type === "String") {
                         return null;
                     }
 
                     // if previous token exists and is not ";" or "{" not sure about ASI rules -> no suggestion.
                     const prevToken = sourceCode.getTokenBefore(parent.parent);
+
                     if (prevToken && !astUtils.isSemicolonToken(prevToken) && !astUtils.isOpeningBraceToken(prevToken)) {
                         return null;
                     }
@@ -1262,7 +1280,7 @@ module.exports = {
                     }
 
                     /*
-                     * fix var foo = [a];
+                     * fix var [a] = foo;
                      * fix function foo([a]) {}
                      * fix const { a: [b] } = foo;
                      */
--- a/lib/rules/no-unused-vars.js
+++ b/lib/rules/no-unused-vars.js
@@ -1165,6 +1165,13 @@ module.exports = {
                         return null;
                     }
 
+                    // removes only variable in 'if (foo()) var bar;' not semicolon.
+                    if (parent.parent.parent.type === "IfStatement") {
+                        return fixer.removeRange(
+                            [sourceCode.getTokenBefore(parent.parent).range[1], parent.range[1]]
+                        );
+                    }
+
                     // // remove unused declared variable with single declaration like 'var a = b;'
                     return fixer.removeRange(parent.parent.range);
                 }
@@ -1334,7 +1341,7 @@ module.exports = {
                     !hasImportOfCertainType(parent.parent, "ImportSpecifier") &&
                     !hasImportOfCertainType(parent.parent, "ImportNamespaceSpecifier")
                 ) {
-                    return fixer.removeRange(parent.parent.range);
+                    return fixer.removeRange([parent.range[0], parent.parent.source.range[0]]);
                 }
 
                 // remove unused default import when there are other imports also
@@ -1351,7 +1358,7 @@ module.exports = {
                         return fixer.removeRange(parent.parent.range);
                     }
 
-                    // remove unused import when there is a default import
+                    // fixes "import foo from 'module';" to "import 'module';"
                     return fixer.removeRange([getPreviousTokenStart(parent, 1), tokenAfter.range[1]]);
                 }
 
@@ -1367,7 +1374,8 @@ module.exports = {
                     return fixer.removeRange([getPreviousTokenStart(parent), parent.range[1]]);
                 }
 
-                return fixer.removeRange(parent.parent.range);
+                // fixes "import * as foo from 'module';" to "import 'module';"
+                return fixer.removeRange([parent.range[0], parent.parent.source.range[0]]);
             }
 
             // skip error in catch(error) variable
--- a/lib/rules/no-unused-vars.js
+++ b/lib/rules/no-unused-vars.js
@@ -1146,7 +1146,7 @@ module.exports = {
                 return null;
             }
 
-            // skip fix when initialized variable in reassigned
+            // skip fix when initialized variable is reassigned
             if (writeReferences.length > 0 && parent.init === null && parentType !== "AssignmentPattern") {
                 return null;
             }
@@ -1171,9 +1171,7 @@ module.exports = {
                         isLoop(parent.parent.parent) ||
                         (parent.parent.parent.type === "WithStatement" && parent.parent.parent.body === parent.parent)
                     ) {
-                        return fixer.removeRange(
-                            [sourceCode.getTokenBefore(parent.parent).range[1], parent.range[1]]
-                        );
+                        return fixer.replaceText(parent.parent, ";");
                     }
 
                     // // remove unused declared variable with single declaration like 'var a = b;'
--- a/lib/rules/no-unused-vars.js
+++ b/lib/rules/no-unused-vars.js
@@ -836,6 +836,576 @@ module.exports = {
             return unusedVars;
         }
 
+        /**
+         * fixes unused variables
+         * @param {Object} fixer fixer object
+         * @param {Object} unusedVar unused variable to fix
+         * @param {Array} writeReferences collection of write references of unused variables
+         * @returns {Object} fixer object
+         */
+        function handleFixes(fixer, unusedVar, writeReferences) {
+            const id = unusedVar.identifiers[0];
+            const parent = id.parent;
+            const parentType = parent.type;
+            const tokenBefore = sourceCode.getTokenBefore(id);
+            const tokenAfter = sourceCode.getTokenAfter(id);
+            const isFunction = astUtils.isFunction;
+            const isLoop = astUtils.isLoop;
+
+            /**
+             * get range from token before of a given node
+             * @param {ASTNode} node node of identifier
+             * @param {number} skips number of token to skip
+             * @returns {number} start range of token before the identifier
+             */
+            function getPreviousTokenStart(node, skips) {
+                return sourceCode.getTokenBefore(node, skips).range[0];
+            }
+
+            /**
+             * get range to token after of a given node
+             * @param {ASTNode} node node of identifier
+             * @param {number} skips number of token to skip
+             * @returns {number} end range of token after the identifier
+             */
+            function getNextTokenEnd(node, skips) {
+                return sourceCode.getTokenAfter(node, skips).range[1];
+            }
+
+            /**
+             * get the value of token before of a given node
+             * @param {ASTNode} node node of identifier
+             * @returns {string} value of token before the identifier
+             */
+            function getTokenBeforeValue(node) {
+                return sourceCode.getTokenBefore(node).value;
+            }
+
+            /**
+             * get the value of token after of a given node
+             * @param {ASTNode} node node of identifier
+             * @returns {string} value of token after the identifier
+             */
+            function getTokenAfterValue(node) {
+                return sourceCode.getTokenAfter(node).value;
+            }
+
+            /**
+             * Check if an array has a single element with null as other element.
+             * @param {ASTNode} node ArrayPattern node
+             * @returns {boolean} true if array has single element with other null elements
+             */
+            function hasSingleElement(node) {
+                return node.elements.filter(e => e !== null).length === 1;
+            }
+
+            /**
+             * check whether import specifier has an import of particular type
+             * @param {ASTNode} node ImportDeclaration node
+             * @param {string} type type of import to check
+             * @returns {boolean} true if import specifier has import of specified type
+             */
+            function hasImportOfCertainType(node, type) {
+                return node.specifiers.some(e => e.type === type);
+            }
+
+            /**
+             * give fixes for unused variables in function parameters
+             * @param {ASTNode} node node to check
+             * @returns {Object} fixer object
+             */
+            function fixFunctionParameters(node) {
+                const parentNode = node.parent;
+
+                if (isFunction(parentNode)) {
+
+                    // remove unused function parameter if there is only a single parameter
+                    if (parentNode.params.length === 1) {
+                        return fixer.removeRange(node.range);
+                    }
+
+                    // remove first unused function parameter when there are multiple parameters
+                    if (getTokenBeforeValue(node) === "(" && getTokenAfterValue(node) === ",") {
+                        return fixer.removeRange([node.range[0], getNextTokenEnd(node)]);
+                    }
+
+                    // remove unused function parameters except first one when there are multiple parameters
+                    return fixer.removeRange([getPreviousTokenStart(node), node.range[1]]);
+                }
+
+                return null;
+            }
+
+            /**
+             * Gets fixes in variable declarations and function parameters
+             * @param {ASTNode} node parent node of identifier
+             * @returns {Object} fixer object
+             */
+            function fixVariables(node) {
+                const parentNode = node.parent;
+
+                // remove unused declared variables such as var a = b; or var a = b, c;
+                if (parentNode.type === "VariableDeclarator") {
+
+                    // remove unused declared variable with single declaration such as 'var a = b;'
+                    if (parentNode.parent.declarations.length === 1) {
+                        return fixer.removeRange(parentNode.parent.range);
+                    }
+
+                    // remove unused declared variable with multiple declaration except first one such as 'var a = b, c = d;'
+                    if (getTokenBeforeValue(parentNode) === ",") {
+                        return fixer.removeRange([getPreviousTokenStart(parentNode), parentNode.range[1]]);
+                    }
+
+                    // remove first unused declared variable when there are multiple declarations
+                    return fixer.removeRange([parentNode.range[0], getNextTokenEnd(parentNode)]);
+                }
+
+                return fixFunctionParameters(node);
+            }
+
+            /**
+             * fix unused variables inside object and array pattern
+             * @param {ASTNode} node parent node to check
+             * @returns {Object} fixer object
+             */
+            function fixObjectAndArrayVariableVariable(node) {
+                const parentNode = node.parent;
+
+                if (parentNode.type === "VariableDeclarator") {
+
+                    // skip variable in for (const [ foo ] of bar);
+                    if (isLoop(parentNode.parent.parent)) {
+                        return null;
+                    }
+
+                    // remove complete declaration when there is an unused variable in 'const { a } = foo;', same for arrays.
+                    if (parentNode.parent.declarations.length === 1) {
+                        return fixer.removeRange(parentNode.parent.range);
+                    }
+
+                    // fix 'let bar = "hello", { a } = foo;' to 'let bar = "hello";' if 'a' is unused, same for arrays.
+                    if (getTokenBeforeValue(parentNode) === ",") {
+                        return fixer.removeRange([getPreviousTokenStart(parentNode), parentNode.range[1]]);
+                    }
+
+                    // fix 'let { a } = foo, bar = "hello";' to 'let  bar = "hello";' if 'a' is unused, same for arrays.
+                    return fixer.removeRange([parentNode.range[0], getNextTokenEnd(parentNode)]);
+                }
+
+                // fixes [{a: {k}}], [{a: [k]}]
+                if (getTokenBeforeValue(node) === ":") {
+                    if (parentNode.parent.type === "ObjectPattern") {
+                        // eslint-disable-next-line no-use-before-define -- due to interdependency of functions
+                        return fixObjectWithValueSeperator(node);
+                    }
+                }
+
+                return fixFunctionParameters(node);
+            }
+
+            /**
+             * fix nested object like { a: { b } }
+             * @param {ASTNode} node parent node to check
+             * @returns {Object} fixer object
+             */
+            function fixNestedObjectVariable(node) {
+                const parentNode = node.parent;
+
+                // fix for { a: { b: c: { d } } }
+                if (
+                    parentNode.parent.parent.parent.type === "ObjectPattern" &&
+                        parentNode.parent.properties.length === 1
+                ) {
+                    return fixNestedObjectVariable(parentNode.parent);
+                }
+
+                // fix for { a: { b } }
+                if (parentNode.parent.type === "ObjectPattern") {
+
+                    // fix for unused variables in dectructured object with single property in variable decalartion and function parameter
+                    if (parentNode.parent.properties.length === 1) {
+                        return fixVariables(parentNode.parent);
+                    }
+
+                    // fix for first unused property when there are multiple properties such as '{ a: { b }, c }'
+                    if (getTokenBeforeValue(parentNode) === "{") {
+                        return fixer.removeRange(
+                            [parentNode.range[0], getNextTokenEnd(parentNode)]
+                        );
+                    }
+
+                    // fix for unused property except first one when there are multiple properties such as '{ k, a: { b } }'
+                    return fixer.removeRange([getPreviousTokenStart(parentNode), parentNode.range[1]]);
+                }
+
+                return null;
+            }
+
+            /**
+             * fix unused variables in array and nested array
+             * @param {ASTNode} node parent node to check
+             * @returns {Object} fixer object
+             */
+            function fixNestedArrayVariable(node) {
+                const parentNode = node.parent;
+
+                // fix for nested arrays [[ a ]]
+                if (parentNode.parent.type === "ArrayPattern" && hasSingleElement(parentNode)) {
+                    return fixNestedArrayVariable(parentNode);
+                }
+
+                if (hasSingleElement(parentNode)) {
+
+                    // fixes { a: [{ b }] } or { a: [[ b ]] }
+                    if (getTokenBeforeValue(parentNode) === ":") {
+                        return fixObjectAndArrayVariableVariable(parentNode);
+                    }
+
+                    // fixes [a, ...[[ b ]]] or [a, ...[{ b }]]
+                    if (parentNode.parent.type === "RestElement") {
+                        // eslint-disable-next-line no-use-before-define -- due to interdependency of functions
+                        return fixRestInPattern(parentNode.parent);
+                    }
+
+                    // fix unused variables in destructured array in variable declaration or function parameter
+                    return fixVariables(parentNode);
+                }
+
+                // remove last unused array element
+                if (
+                    getTokenBeforeValue(node) === "," &&
+                        getTokenAfterValue(node) === "]"
+                ) {
+                    return fixer.removeRange([getPreviousTokenStart(node), node.range[1]]);
+                }
+
+                // remove unused array element
+                return fixer.removeRange(node.range);
+            }
+
+            /**
+             * fix cases like {a: {k}} or {a: [k]}
+             * @param {ASTNode} node parent node to check
+             * @returns {Object} fixer object
+             */
+            function fixObjectWithValueSeperator(node) {
+                const parentNode = node.parent.parent;
+
+                // fix cases like [{a : { b }}] or [{a : [ b ]}]
+                if (
+                    parentNode.parent.type === "ArrayPattern" &&
+                        parentNode.properties.length === 1
+                ) {
+                    return fixNestedArrayVariable(parentNode);
+                }
+
+                // fix cases like {a: {k}} or {a: [k]}
+                return fixNestedObjectVariable(node);
+            }
+
+            /**
+             * fix [...[a]] or [...{a}] like patterns
+             * @param {ASTNode} node parent node to check
+             * @returns {Object} fixer object
+             */
+            function fixRestInPattern(node) {
+                const parentNode = node.parent;
+
+                // fix [...[a]] or [...{a}] in function parameters
+                if (isFunction(parentNode)) {
+                    if (parentNode.params.length === 1) {
+                        return fixer.removeRange(node.range);
+                    }
+
+                    return fixer.removeRange([getPreviousTokenStart(node), node.range[1]]);
+                }
+
+                // fix rest in nested array pattern like [[a, ...[b]]]
+                if (parentNode.type === "ArrayPattern") {
+
+                    // fix [[...[b]]]
+                    if (hasSingleElement(parentNode)) {
+                        if (parentNode.parent.type === "ArrayPattern") {
+                            return fixNestedArrayVariable(parentNode);
+                        }
+
+                        // fix 'const [...[b]] = foo; and function foo([...[b]]) {}
+                        return fixVariables(parentNode);
+                    }
+
+                    // fix [[a, ...[b]]]
+                    return fixer.removeRange([getPreviousTokenStart(node), node.range[1]]);
+                }
+
+                return null;
+            }
+
+            // skip fix when variable is reassigned
+            if (writeReferences.length > 1 && parentType !== "AssignmentPattern") {
+                return null;
+            }
+
+            // skip fix when initialized variable in reassigned
+            if (writeReferences.length > 0 && parent.init === null && parentType !== "AssignmentPattern") {
+                return null;
+            }
+
+            // skip fix when function parameters are ressaigned
+            if (isFunction(parent) && writeReferences.length > 0 && parent.params.includes(id)) {
+                return null;
+            }
+
+            // remove declared variables such as var a; or var a, b;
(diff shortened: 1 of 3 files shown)


lib/rules/no-unused-vars.js
```diff
@@ -5,6 +5,8 @@
 
 "use strict";
 
+// const { parents } = require("cheerio/lib/api/traversing");
+// const { remove } = require("../linter/rule-fixer");
```
nzakas, line 9: Looks like your editor might have added some unintentional imports?

nzakas (commented):
It looks like the whole file has been reformatted, making it difficult to view the diff. Can you fix the formatting?

Tanujkanti4441:
@eslint/eslint-team, code in this PR seems to work fine but having linting error that `Method 'fix' expected no return value` can i get some help in figuring out what is wrong or is there something i am not aware of?

aladdin-add:
It's always required to return a value in fixers (to catch possible errors). if you don't want to fix in some cases, please use `return null;`

lib/rules/no-unused-vars.js
```diff
@@ -858,10 +861,341 @@ module.exports = {
...
+                                                }
+
+                                                if (sourceCode.getTokenBefore(node).value === "(" && sourceCode.getTokenAfter(node).value === ",") {
+                                                    return fixer.removeRange([node.range[0], sourceCode.getTokenAfter(node).range[1]]);
+                                                }
+
+                                                return fixer.removeRange([sourceCode.getTokenBefore(node).range[0], node.range[1]]);
+                                            }
+
+                                            return;
```
fasttime, line 907:
```suggestion
                                            return null;
```

To fix the lint errors.

lib/rules/no-unused-vars.js
```diff
@@ -858,10 +861,341 @@ module.exports = {
...
+                                         * @returns {boolean} true if array has single element with other null elements
+                                         */
+                                        function hasSingleElement(node) {
+                                            return node.elements.filter(e => e !== null).length === 1;
+                                        }
+
+                                        if (parent.type === "VariableDeclarator") {
+                                            if (parent.parent.declarations.length === 1) {
+                                                if (parent.parent.parent.type === "ForOfStatement" || parent.parent.parent.type === "ForInStatement") {
+                                                    return;
```
fasttime, line 922:
```suggestion
                                                    return null;
```

We need to explicitly return something here, otherwise the no-unused-vars rule will assume that this function does not return a value, and it will flag all following `return` statements that have an expression.

lib/rules/no-unused-vars.js

Tanujkanti4441:
I think it's ready for review now.

lib/rules/no-unused-vars.js
```diff
@@ -858,10 +861,357 @@ module.exports = {
...
+                                        const tokenBefore = sourceCode.getTokenBefore(id);
+                                        const tokenAfter = sourceCode.getTokenAfter(id);
+
+                                        /**
+                                         * Gets fixes in variable declarations and function parameters
+                                         * @param {ASTNode} node parent node of identifier
+                                         * @returns {Object} fixer object
+                                         */
+                                        function fixVariables(node) {
+                                            if (node.parent.type === "VariableDeclarator") {
```
nzakas, line 883: It looks like you're using `node.parent` a lot, which requires looking up that property each time. I'd suggest saving a reference to both the parent and parent type so you don't have to keep evaluating it each time:

```js
const parent  = node.parent;
const parentType = parent.type;

File: lib/rules/no-unused-vars.js [hunk 5]
```diff
@@ -858,10 +861,357 @@ module.exports = {
...
+
+                                                return fixer.removeRange([node.parent.range[0], sourceCode.getTokenAfter(node.parent).range[1]]);
+                                            }
+
+                                            if (node.parent.type === "FunctionDeclaration" || node.parent.type === "FunctionExpression") {
+                                                if (node.parent.params.length === 1) {
+                                                    return fixer.removeRange(node.range);
+                                                }
+
+                                                if (sourceCode.getTokenBefore(node).value === "(" && sourceCode.getTokenAfter(node).value === ",") {
```
nzakas, line 900: You're also doing `getTokenBefore(node)` and `getTokenAfter(node)` a few times here. It's best to store those values in a variable to save the function call.

lib/rules/no-unused-vars.js
```diff
@@ -858,10 +861,357 @@ module.exports = {
...
+                                        /**
+                                         * Check if an array has a single element with null as other element.
+                                         * @param {ASTNode} node ArrayPattern node
+                                         * @returns {boolean} true if array has single element with other null elements
+                                         */
+                                        function hasSingleElement(node) {
+                                            return node.elements.filter(e => e !== null).length === 1;
+                                        }
+
+                                        if (parent.type === "VariableDeclarator") {
```
nzakas, line 919: Same comment about references throughout. We want to avoid a.b.c.d as much as possible.

nzakas (commented):
Left some comments related to storing variables for later use.

Also, there are a lot of `if` statements that aren't obvious what they're doing. Can we add some comments describing what each is accomplishing?

lib/rules/no-unused-vars.js
Tanujkanti4441: Here `node.parent` is just used in `fixVariable` function and parent is already declared as
```js
const parent = id.parent
```
is it ok if i don't replace `node.parent` for now because it will help me if i use it later also?

lib/rules/no-unused-vars.js

lib/rules/no-unused-vars.js

lib/rules/no-unused-vars.js
```diff
@@ -98,7 +100,8 @@ module.exports = {
 
         messages: {
             unusedVar: "'{{varName}}' is {{action}} but never used{{additional}}.",
-            usedIgnoredVar: "'{{varName}}' is marked as ignored but is used{{additional}}."
+            usedIgnoredVar: "'{{varName}}' is marked as ignored but is used{{additional}}.",
+            removeVar: "remove unused variable '{{varName}}'."
```
mdjermanovic, line 104:
```suggestion
            removeVar: "Remove unused variable '{{varName}}'."
```

lib/rules/no-unused-vars.js
```diff
@@ -858,10 +861,439 @@ module.exports = {
...
+
+                                        // remove vaiables in object patterns
+                                        if (parent.parent.type === "ObjectPattern") {
+                                            if (parent.parent.properties.length === 1) {
+
+                                                // var {a} = foo;
+                                                if (parent.parent.parent.type === "VariableDeclarator") {
+                                                    if (parent.parent.parent.parent.declarations.length === 1) {
+                                                        return fixer.removeRange(parent.parent.parent.parent.range);
+                                                    }
```
mdjermanovic, line 1001: Patterns can also appear in for-in/for-of loops:

```js
for (const { foo } of bar);
```

After the fix:

```js
for ( of bar);
```

lib/rules/no-unused-vars.js
```diff
@@ -858,10 +861,439 @@ module.exports = {
                             messageId: "unusedVar",
                             data: unusedVar.references.some(ref => ref.isWrite())
                                 ? getAssignedMessageData(unusedVar)
-                                : getDefinedMessageData(unusedVar)
+                                : getDefinedMessageData(unusedVar),
+                            suggest: [
```
mdjermanovic, line 865: I'm not sure if we should remove the declaration if there are references to that variable.

```js
var x;

x = 1;
```

After the fix:

```js


x = 1;
```

lib/rules/no-unused-vars.js
```diff
@@ -858,10 +861,439 @@ module.exports = {
...
+
+                                                    return fixer.removeRange([getBeforeToken(parent.parent), parent.parent.range[1]]);
+                                                }
+
+                                                // fix const { a: { b } } = foo;
+                                                if (getTokenBeforeValue(parent.parent) === ":") {
+                                                    if (parent.parent.parent.parent.type === "ObjectPattern") {
+                                                        if (parent.parent.parent.parent.properties.length === 1) {
+                                                            return fixVariables(parent.parent.parent.parent);
+                                                        }
```
mdjermanovic, line 1035: There seems to be a lot of repeated code for different combinations of patterns in patterns. Can we maybe unify the code? For example, when we determine that something can be removed, recursively check the parent and so on until we find the uppermost node that can be removed.

lib/rules/no-unused-vars.js
Tanujkanti4441: Done till here!

lib/rules/no-unused-vars.js
Tanujkanti4441: is this statement suggesting to use loops? may be i don't get it properly.
how about if we use function to reduce the repetition.

mdjermanovic: Yes, loops or recursion. That would avoid code repetition and cover more cases.

For example, the current implementation provides a suggestion to fix `const [[foo]] = bar;` because it has a branch that specifically covers the array pattern in array pattern case. But it doesn't provide a suggestion to fix `const [[[foo]]] = bar;` although it's basically the same case, just one more level deep. The same for any combination of patterns, for example the current implementation provides a suggestion to fix `const { a: [b] } = c;`, but doesn't provide a suggestion to fix `const [{ a: [b] }] = c;`.

Tanujkanti4441: fix for some nested patterns has implemented, working on some others, so marked it as draft.

lib/rules/no-unused-vars.js
```diff
@@ -858,10 +861,499 @@ module.exports = {
...
+                                        const tokenBefore = sourceCode.getTokenBefore(id);
+                                        const tokenAfter = sourceCode.getTokenAfter(id);
+
+                                        /**
+                                         * get range from token before of a given node
+                                         * @param {ASTNode} node node of identifier
+                                         * @param {number} skips number of token to skip
+                                         * @returns {number} start range of token before the identifier
+                                         */
+                                        function getBeforeToken(node, skips) {
```
nzakas, line 884: This doesn't seem to describe what the function is actually doing. Maybe `getPreviousTokenStart` would be a better name?

lib/rules/no-unused-vars.js
```diff
@@ -858,10 +861,499 @@ module.exports = {
...
+                                            return sourceCode.getTokenBefore(node, skips).range[0];
+                                        }
+
+                                        /**
+                                         * get range to token after of a given node
+                                         * @param {ASTNode} node node of identifier
+                                         * @param {number} skips number of token to skip
+                                         * @returns {number} end range of token after the identifier
+                                         */
+                                        function getAfterToken(node, skips) {
```
nzakas, line 894: Same here. Maybe `getNextTokenEnd` would make more sense?

lib/rules/no-unused-vars.js
```diff
@@ -858,10 +861,499 @@ module.exports = {
...
+                                        function getTokenAfterValue(node) {
+                                            return sourceCode.getTokenAfter(node).value;
+                                        }
+
+                                        /**
+                                         * check whether a given node is function
+                                         * @param {ASTNode} node parent of identifier
+                                         * @returns {boolean} true if node is function
+                                         */
+                                        function isTypeFunction(node) {
```
nzakas, line 921: We have a utility function for this:
https://github.com/eslint/eslint/blob/80747d23dec69b30ea2c3620a1198f7d06b012b8/lib/rules/utils/ast-utils.js#L125

lib/rules/no-unused-vars.js
```diff
@@ -858,10 +861,499 @@ module.exports = {
...
+                                                node.type === "ArrowFunctionExpression"
+                                            );
+                                        }
+
+                                        /**
+                                         * check if given node is forOf or forIn loop
+                                         * @param {ASTNode} node node to check
+                                         * @returns {boolean} true if node is forOf or forIn loop
+                                         */
+                                        function isLoop(node) {
```
nzakas, line 934: We have a utility function for this:
https://github.com/eslint/eslint/blob/80747d23dec69b30ea2c3620a1198f7d06b012b8/lib/rules/utils/ast-utils.js#L141

lib/rules/no-unused-vars.js
nzakas: I'm not sure I understand what you're saying. I see `node.parent` multiple times in the following lines, and there's just no need to keep doing the lookup.

lib/rules/no-unused-vars.js
```diff
@@ -858,10 +861,499 @@ module.exports = {
...
+                                                }
+
+                                                if (getTokenBeforeValue(node.parent) === ",") {
+                                                    return fixer.removeRange([getBeforeToken(node.parent), node.parent.range[1]]);
+                                                }
+
+                                                return fixer.removeRange([node.parent.range[0], getAfterToken(node.parent)]);
+                                            }
+
+                                            // remove unused varibles inside functions arguments
```
nzakas, line 967:
```suggestion
                                            // remove unused variables inside functions arguments
```

lib/rules/no-unused-vars.js
```diff
@@ -858,10 +861,499 @@ module.exports = {
...
+                                            if (isTypeFunction(node.parent)) {
+                                                if (node.parent.params.length === 1) {
+                                                    return fixer.removeRange(node.range);
+                                                }
+
+                                                if (getTokenBeforeValue(node) === "(" && getTokenAfterValue(node) === ",") {
```
nzakas, line 973: Please add a comment describing what this does.

lib/rules/no-unused-vars.js
```diff
@@ -858,10 +861,499 @@ module.exports = {
...
+
+                                            return null;
+                                        }
+
+                                        /**
+                                         * fix unused variables inside object and array pattern
+                                         * @param {ASTNode} node parent node to check
+                                         * @returns {Object} fixer object
+                                         */
+                                        function fixObjectAndArrayVariable(node) {
```
nzakas, line 988: For clarity

```suggestion
                                        function fixObjectAndArrayPatternVariable(node) {
```

lib/rules/no-unused-vars.js
```diff
@@ -858,10 +861,499 @@ module.exports = {
...
+                                            if (node.parent.type === "VariableDeclarator") {
+
+                                                // skip variable in for (const [ foo ] of bar);
+                                                if (isLoop(node.parent.parent.parent)) {
```
nzakas, line 992: Again, a lot of `node.parent` and `node.parent.parent` in the following lines. These should be pulled up into variables to avoid constant lookup.

lib/rules/no-unused-vars.js
```diff
@@ -858,10 +861,499 @@ module.exports = {
...
+                                                    return null;
+                                                }
+
+                                                if (node.parent.parent.declarations.length === 1) {
```
nzakas, line 996: Please add a comment describing what this does.

lib/rules/no-unused-vars.js
```diff
@@ -858,10 +861,499 @@ module.exports = {
...
+                                                    return fixer.removeRange(node.parent.parent.range);
+                                                }
+
+                                                if (getTokenBeforeValue(node.parent) === ",") {
```
nzakas, line 1000: Please add a comment describing what this does.

lib/rules/no-unused-vars.js
```diff
@@ -858,10 +861,499 @@ module.exports = {
...
+                                                    return fixer.removeRange([getBeforeToken(node.parent), node.parent.range[1]]);
+                                                }
+
+                                                return fixer.removeRange([node.parent.range[0], getAfterToken(node.parent)]);
+                                            }
+
+                                            if (isTypeFunction(node.parent)) {
```
nzakas, line 1007: Please add a comment describing what this does.

Also note, this is almost exactly the same as earlier in the file. Maybe there's a way to extract into a function?

lib/rules/no-unused-vars.js
```diff
@@ -858,10 +861,499 @@ module.exports = {
...
+                                                }
+
+                                                if (getTokenBeforeValue(node) === "(" && getTokenAfterValue(node) === ",") {
+                                                    return fixer.removeRange([node.range[0], getAfterToken(node)]);
+                                                }
+
+                                                return fixer.removeRange([getBeforeToken(node), node.range[1]]);
+                                            }
+
+                                            if (getTokenBeforeValue(node) === ":") {
```
nzakas, line 1019: Please add a comment describing what this does.

nzakas (commented):
I've added some comments, but my previous comments weren't addressed: There are a lot of `if` statements that aren't obvious as to what they do. Please add comments for all of them.

Also, we need to cut down on the number of times we have `node.parent` and `node.parent.parent` throughout. These references should be stored in variables to avoid the cost of property lookup over and over again.

lib/rules/no-unused-vars.js

lib/rules/no-unused-vars.js
Tanujkanti4441: Done!

aladdin-add:

nzakas (approved):
LGTM. Thanks!

Leaving open for @mdjermanovic and @fasttime to verify their comments.

fasttime:
Thanks for working on this @Tanujkanti4441. After running `npm test`, I checked the code coverage in `./coverage/lcov-report/lib/rules/no-unused-vars.js.html`, and it looks like some parts of the new code are not covered by unit tests. You could see this by yourself if you open that file in a browser:

image

There are 99 lines never hit by a test case, all but one in the new code. I haven't looked into the detail of what the missing tests would need to check, but I think you could figure that out quickly since you are already familiar with the logic. Sometimes the relevant code is already indicated in the comments, like this one:

```js
// fix 'let { a } = foo, bar = "hello";' to 'let  bar = "hello";' if 'a' is unused, same for arrays.
```

Those should be a good start to add invalid test cases.

Tanujkanti4441:
sorry! wasn't active due to some work and health issues, working on it.

Tanujkanti4441:
@fasttime, have written most of the test cases, but not sure about how to write the test where i have to return null, do you have any suggestions?
Screenshot 2024-06-30 212307

fasttime:
Thanks for the update @Tanujkanti4441, I think it's okay to leave that line uncovered if it cannot be reached in a test case. Probably a pattern like `...[a]` or `...{a}` can only appear inside an array destructuring pattern, and so the previous conditional path `if (parentNode.type === "ArrayPattern")` is always taken. You could leave a comment on `return null;` saying that that line should never be reached when using the default parser.

fasttime:
@Tanujkanti4441 Sorry for the delay, I will review this PR soon.

lib/rules/no-unused-vars.js
```diff
@@ -860,7 +864,563 @@ module.exports = {
...
+                                        if (parentType === "VariableDeclarator") {
+                                            if (parent.parent.declarations.length === 1) {
+
+                                                // prevent fix of variable in forOf and forIn loops.
+                                                if (isLoop(parent.parent.parent)) {
+                                                    return null;
+                                                }
+
+                                                // // remove unused declared variable with single declaration like 'var a = b;'
+                                                return fixer.removeRange(parent.parent.range);
```
fasttime, line 1187: If the code is like:

```js
let a;
(() => a = 1)();
```

The rule now suggests to remove `let a;`, which seems incorrect, since `a` is assigned a value.

lib/rules/no-unused-vars.js
```diff
@@ -860,7 +864,563 @@ module.exports = {
...
+                                            return null;
+                                        }
+
+                                        // remove unused declared classes
+                                        if (parentType === "ClassDeclaration") {
+                                            return fixer.removeRange(parent.range);
+                                        }
+
+                                        // remove unused varible that is in a sequence [a,b] fixes to [a]
+                                        if (tokenBefore.value === ",") {
```
fasttime, line 1402:
```suggestion
                                        if (tokenBefore?.value === ",") {
```

`tokenBefore` could be `null`, here and in other parts of the `fix` function.

Maybe we could add a test with code like `a => a = 1;` to make sure that the rule doesn't crash (currently it does).

fasttime (changes requested):
There is a merge conflict.

lib/rules/no-unused-vars.js
```diff
@@ -860,7 +864,563 @@ module.exports = {
...
                                 ? getAssignedMessageData(unusedVar)
-                                : getDefinedMessageData(unusedVar)
+                                : getDefinedMessageData(unusedVar),
+                            suggest: [
+                                {
+                                    messageId: "removeVar",
+                                    data: {
+                                        varName: unusedVar.name
+                                    },
+                                    fix(fixer) {
```
fasttime, line 874: Can we extract the `fix` method into a function before the line `context.report({` for better readability, and unindent it by 12 spaces?

lib/rules/no-unused-vars.js
Tanujkanti4441: Done! rule won't give suggestion for this case and even when a function parameter is reassigned.

lib/rules/no-unused-vars.js
Tanujkanti4441: Done for this particular part, example like `a => a = 1;` is already there, i think here `a` should not be removed in this case and in other cases where a is not reassigned it will get replaced by `()`.

lib/rules/no-unused-vars.js
Tanujkanti4441: Done!

fasttime:
@christian-bromann the browser test is failing again with a timeout sometimes: https://github.com/eslint/eslint/actions/runs/10132246872/job/28015986969.

christian-bromann:
@fasttime taking a look 👀

tests/lib/rules/no-unused-vars.js
```diff
@@ -1792,6 +2034,460 @@ _ => { _ = _ + 1 };
...
+        {
+            code: "import a from 'module';",
+            languageOptions: { ecmaVersion: 6, sourceType: "module" },
+            errors: [definedError("a", [{ output: "", messageId: "removeVar", data: { varName: "a" } }])]
+        },
+        {
+            code: "import * as foo from 'module'",
+            languageOptions: { ecmaVersion: 6, sourceType: "module" },
+            errors: [definedError("foo", [{ output: "", messageId: "removeVar", data: { varName: "foo" } }])]
+        },
```
fasttime, line 2426: In those cases it seems more correct to keep the import and just drop the variable. So the suggestion output should be probably `import 'module';`.

tests/lib/rules/no-unused-vars.js
fasttime, line 1: This is an edge case:

```js
if (foo()) var bar;
```

Currently, the suggestion output is `if (foo())` which is a syntax error. Should be `if (foo()) ;` I guess.

tests/lib/rules/no-unused-vars.js
Tanujkanti4441: Done!

tests/lib/rules/no-unused-vars.js
Tanujkanti4441: Fixed!

lib/rules/no-unused-vars.js
fasttime, line 1: We have edge cases where removing a declaration could suppress automatic semicolon insertion and change program semantics. For example:

```js
console.log('foo')
var a
+b > 0 ? bar() : baz()
```

If `var a` is removed as per suggestion, this becomes:

```js
console.log('foo')

+b > 0 ? bar() : baz()
```

which is the same as

```js
(console.log(foo) + b) > 0 ? bar() : baz()
```

At this moment, I'm not sure what the best way to handle those edge cases is, but I think we will have to get that sorted out before we can merge this PR.

fasttime (commented):
Thanks for the update and sorry for the delays in reviewing this... I'm being a little busy at the moment. I left some feedback if you would like to take a look.

lib/rules/no-unused-vars.js
```diff
@@ -832,6 +836,584 @@ module.exports = {
...
+            if (parentType === "VariableDeclarator") {
+                if (parent.parent.declarations.length === 1) {
+
+                    // prevent fix of variable in forOf and forIn loops.
+                    if (isLoop(parent.parent.parent)) {
+                        return null;
+                    }
+
+                    // removes only variable in 'if (foo()) var bar;' not semicolon.
+                    if (parent.parent.parent.type === "IfStatement") {
```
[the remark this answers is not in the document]
Tanujkanti4441: Done till here!

lib/rules/no-unused-vars.js
Tanujkanti4441: How about if we first check if there is a semi colon just before the variable declaration if so then we can safely remove the unused variable otherwise we can replace the variable declaration with a semicolon.
```js
console.log('foo')
var a
+b > 0 ? bar() : baz()
 // fixes to 

console.log('foo');
+b > 0 ? bar() : baz()
```

fasttime: > How about if we first check if there is a semi colon just before the variable declaration if so then we can safely remove the unused variable otherwise we can replace the variable declaration with a semicolon.

This would occasionally add semicolons in places where they aren't needed, like at the beginning of a block, but I think it's better that providing no suggestion at all. We could fine-tune that a little until we have an algorithm that detects accurately when a semicolon is necessary. I wonder what other this about this @eslint/eslint-team.

nzakas: If you're unsure what to do, then just don't generate the suggestion. Producing unexpected results is counter-productive to providing suggestions.

fasttime (commented):
I've left some comments to incorporate @nzakas' feedback and let us move on with this PR.

lib/rules/no-unused-vars.js
```diff
@@ -832,6 +836,588 @@ module.exports = {
...
+                    if (
+                        parent.parent.parent.type === "IfStatement" ||
+                        isLoop(parent.parent.parent) ||
+                        (parent.parent.parent.type === "WithStatement" && parent.parent.parent.body === parent.parent)
+                    ) {
+                        return fixer.removeRange(
+                            [sourceCode.getTokenBefore(parent.parent).range[1], parent.range[1]]
+                        );
+                    }
+
```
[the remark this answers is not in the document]
Tanujkanti4441:
```js
const prevToken = sourceCode.getTokenBefore(parent.parent);
if (prevToken && !astUtils.isSemicolonToken(prevToken) && !astUtils.isOpeningBraceToken(prevToken)) {
    return null;
}
```

this code also prevents suggestions in these cases
```js
for (;;) var foo;
for (a in b) var foo;
for (a of b) var foo;
while (a) var foo;
do var foo; while (b);
with (a) var foo;
```
should we modify the conditions to check if previous node is a `loop` or an `if` or a `with` statement as we are replacing the declaration with a semicolon?

lib/rules/no-unused-vars.js
```diff
@@ -832,6 +836,588 @@ module.exports = {
...
+                // remove unused declared variables such as var a = b; or var a = b, c;
+                if (parentNode.type === "VariableDeclarator") {
+
+                    // remove unused declared variable with single declaration such as 'var a = b;'
+                    if (parentNode.parent.declarations.length === 1) {
+                        return fixer.removeRange(parentNode.parent.range);
+                    }
+
+                    // remove unused declared variable with multiple declaration except first one such as 'var a = b, c = d;'
+                    if (getTokenBeforeValue(parentNode) === ",") {
```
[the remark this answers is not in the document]
Tanujkanti4441: Here in this suggestion these conditions will only be applied when there are multiple declarations `(var a, b;)` as they are under this condition:
```js
if (getTokenBeforeValue(parentNode) === ",") {// code}
```

i guess intend here is to prevent suggestions when there is only one declaration, could you please clarify that?

lib/rules/no-unused-vars.js
```diff
@@ -832,6 +836,588 @@ module.exports = {
...
+
+                    // removes only variable not semicolon in 'if (foo()) var bar;' or in 'loops' or in 'with' statement.
+                    if (
+                        parent.parent.parent.type === "IfStatement" ||
+                        isLoop(parent.parent.parent) ||
+                        (parent.parent.parent.type === "WithStatement" && parent.parent.parent.body === parent.parent)
+                    ) {
+                        return fixer.removeRange(
+                            [sourceCode.getTokenBefore(parent.parent).range[1], parent.range[1]]
+                        );
```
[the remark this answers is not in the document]

lib/rules/no-unused-vars.js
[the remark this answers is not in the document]
fasttime: > ```js
> for (;;) var foo;
> for (a in b) var foo;
> for (a of b) var foo;
> while (a) var foo;
> do var foo; while (b);
> with (a) var foo;
> ```

These cases are already handled in the previous lines of code, so they should not be affected by the additional checks:

https://github.com/eslint/eslint/blob/dfc5b4e39408002a331a1ce1d11f90c213d8a92c/lib/rules/no-unused-vars.js#L1168-L1175

lib/rules/no-unused-vars.js
```diff
@@ -832,6 +836,586 @@ module.exports = {
...
+             * @returns {Object} fixer object
+             */
+            function fixVariables(node) {
+                const parentNode = node.parent;
+
+                // remove unused declared variables such as var a = b; or var a = b, c;
+                if (parentNode.type === "VariableDeclarator") {
+
+                    // remove unused declared variable with single declaration such as 'var a = b;'
+                    if (parentNode.parent.declarations.length === 1) {
```
fasttime, line 951: The same check should be added here:

```suggestion
                    if (parentNode.parent.declarations.length === 1) {

                        // if next token is a string it could become a directive if node is removed -> no suggestion.
                        const nextToken = sourceCode.getTokenAfter(parentNode.parent);
                        if (nextToken?.type === "String") {
                            return null;
                        }

                        // if previous token exists and is not ";" or "{" not sure about ASI rules -> no suggestion.
                        const prevToken = sourceCode.getTokenBefore(parentNode.parent);
                        if (prevToken && !astUtils.isSemicolonToken(prevToken) && !astUtils.isOpeningBraceToken(prevToken)) {
                            return null;
                        }
```

lib/rules/no-unused-vars.js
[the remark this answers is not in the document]
fasttime: You are correct! These checks should be added to the previous if-block, where we test that there is exactly 1 declaration. I've added this suggestion at the correct line now. Sorry for the confusion.

lib/rules/no-unused-vars.js
Tanujkanti4441: Done.

lib/rules/no-unused-vars.js
```diff
@@ -832,6 +836,630 @@ module.exports = {
...
+                    if (nextToken?.type === "String") {
+                        return null;
+                    }
+
+                    // if previous token exists and is not ";" or "{" not sure about ASI rules -> no suggestion.
+                    const prevToken = sourceCode.getTokenBefore(parent.parent);
+
+                    if (prevToken && !astUtils.isSemicolonToken(prevToken) && !astUtils.isOpeningBraceToken(prevToken)) {
+                        return null;
+                    }
```
fasttime, line 1219: To fix the failing tests in `tests/lib/linter/linter.js`. Alternatively, we could revert the changes in that file, but it seems actually better to improve the suggestion logic a little bit to consider safe those cases where a declaration is removed at the end of file.

```suggestion
                    if (nextToken) {
                        if (nextToken.type === "String") {
                            return null;
                        }

                        // if previous token exists and is not ";" or "{" not sure about ASI rules -> no suggestion.
                        const prevToken = sourceCode.getTokenBefore(parent.parent);

                        if (prevToken && !astUtils.isSemicolonToken(prevToken) && !astUtils.isOpeningBraceToken(prevToken)) {
                            return null;
                        }
                    }
```

fasttime (commented):
Now the updated Linter tests are failing. I left a suggestion on how to fix that.

lib/rules/no-unused-vars.js
Tanujkanti4441: have changed the logic but linter still failing.

fasttime: Thanks! That's not a linter error, it's a problem with GitHub Actions that is happening with other PRs as well, for example [here](https://github.com/eslint/eslint/actions/runs/11529133487/job/32097195940) or [here](https://github.com/eslint/eslint/actions/runs/11528700815/job/32096167483). I hope it's temporary.

fasttime: @Tanujkanti4441 can you try to rebase this branch on top of origin/main? Hopefully that will fix the build error.

image

lib/rules/no-unused-vars.js
```diff
@@ -832,6 +836,636 @@ module.exports = {
...
+                            return null;
+                        }
+
+                        // if previous token exists and is not ";" or "{" not sure about ASI rules -> no suggestion.
+                        const prevToken = sourceCode.getTokenBefore(parent.parent);
+
+                        if (prevToken && !astUtils.isSemicolonToken(prevToken) && !astUtils.isOpeningBraceToken(prevToken)) {
+                            return null;
+                        }
+                    }
```
fasttime, line 1225: This block is repeated in `fixVariables` and `fixObjectAndArrayVariable`. Maybe we could extract the logic into a new function and leave here just a simple check like:

```js
if (!isDeclarationSafeToRemove(parent, nextToken)) {
    return null;
}
```

fasttime (commented):
I think this is looking good already. Just a couple of suggestions to remove duplicated code.

lib/rules/no-unused-vars.js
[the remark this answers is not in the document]
Tanujkanti4441: Done! I replaced the content of `fixVariables` with `fixObjectAndArrayVariable`, now we only have `fixVariables`.

lib/rules/no-unused-vars.js
```diff
@@ -832,6 +836,607 @@ module.exports = {
...
+
+            /**
+             * fix nested object like { a: { b } }
+             * @param {ASTNode} node parent node to check
+             * @returns {Object} fixer object
+             */
+            function fixNestedObjectVariable(node) {
+                const parentNode = node.parent;
+
+                // fix for { a: { b: c: { d } } }
```
fasttime, line 1026:
```suggestion
                // fix for { a: { b: { c: { d } } }
```

lib/rules/no-unused-vars.js
```diff
@@ -832,6 +836,607 @@ module.exports = {
...
+                ) {
+                    return fixNestedArrayVariable(parentNode);
+                }
+
+                // fix cases like {a: {k}} or {a: [k]}
+                return fixNestedObjectVariable(node);
+            }
+
+            /**
+             * fix [...[a]] or [...{a}] like patterns
```
fasttime, line 1119:
```suggestion
             * fix ...[[a]] or ...[{a}] like patterns
```

lib/rules/no-unused-vars.js
```diff
@@ -832,6 +836,607 @@ module.exports = {
...
+             * @param {ASTNode} node parent node to check
+             * @returns {Object} fixer object
+             */
+            function fixRestInPattern(node) {
+                const parentNode = node.parent;
+
+                // fix [...[a]] or [...{a}] in function parameters
```
fasttime, line 1126: Also here:

```suggestion
                // fix ...[[a]] or ...[{a}] in function parameters
```

lib/rules/no-unused-vars.js
```diff
@@ -832,6 +836,607 @@ module.exports = {
...
+
+                    /*
+                     * fix var [a] = foo;
+                     * fix function foo([a]) {}
+                     * fix const { a: [b] } = foo;
+                     */
+                    return fixVariables(parent);
+                }
+
+                // if "a" is unused in [a, b ,c] fixes to [, b c]
```
fasttime, line 1268:
```suggestion
                // if "a" is unused in [a, b ,c] fixes to [, b, c]
```

fasttime (approved):
LGTM, thanks! There are still duplications in the code but I think those can be cleaned up later, as we will probably revisit details of the suggestion logic in a near future. Leaving open for @mdjermanovic to verify that his feedback has been addressed.

lib/rules/no-unused-vars.js
```diff
@@ -860,7 +1465,18 @@ module.exports = {
...
-                                : getDefinedMessageData(unusedVar)
+                                : getDefinedMessageData(unusedVar),
+                            suggest: [
+                                {
+                                    messageId: "removeVar",
+                                    data: {
+                                        varName: unusedVar.name
+                                    },
+                                    fix(fixer) {
+                                        return handleFixes(fixer, unusedVar, writeReferences);
```
mdjermanovic, line 1476: `writeReferences` contains only write references in the same execution context (same `variableScope`, e.g., same function):

https://github.com/eslint/eslint/blob/4f08332ac03e51002f8de6da5c5a362608205437/lib/rules/no-unused-vars.js#L850



If we don't want to remove a variable that has references, then in this case it wouldn't work as intended:

```js
let x; // suggestion removes this

() => x = 1;

```

lib/rules/no-unused-vars.js
```diff
@@ -832,6 +836,607 @@ module.exports = {
...
+                    return fixer.removeRange([getPreviousTokenStart(node), node.range[1]]);
+                }
+
+                return null;
+            }
+
+            // skip fix when variable is reassigned
+            if (writeReferences.length > 1 && parentType !== "AssignmentPattern") {
+                return null;
+            }
```
mdjermanovic, line 1158: Variables initialized in an assignment pattern would be removed even if they have write references, which is probably not what we want:

```js
let [a = 1] = arr; // suggestion removes this

a = 2;
```

Tanujkanti4441: @mdjermanovic, should we skip suggestion here also for `b`?
```js
/*eslint no-unused-vars: ["error", { "destructuredArrayIgnorePattern": "^_" }]*/

let _a, b; foo.forEach(item => { [a, b] = item; });

//'_a' is defined but never used.
//'b' is assigned a value but never used.
```

mdjermanovic: If the suggestion is going to remove only the declaration, then yes we should skip the suggestion because it would leave unresolved references in the code.

Tanujkanti4441: Done! thanks.

lib/rules/no-unused-vars.js
```diff
@@ -860,7 +1470,18 @@ module.exports = {
...
-                                : getDefinedMessageData(unusedVar)
+                                : getDefinedMessageData(unusedVar),
+                            suggest: [
+                                {
+                                    messageId: "removeVar",
+                                    data: {
+                                        varName: unusedVar.name
+                                    },
+                                    fix(fixer) {
+                                        return handleFixes(fixer, unusedVar, writeReferences);
```
mdjermanovic, line 1481:
```suggestion
                                        return handleFixes(fixer, unusedVar);
```

lib/rules/no-unused-vars.js
```diff
@@ -1153,17 +1153,22 @@ module.exports = {
...
             // skip fix when variable is reassigned
-            if (writeReferences.length > 1 && parentType !== "AssignmentPattern") {
+            if (allWriteReferences.length > 1 && parentType !== "AssignmentPattern") {
+                return null;
+            }
+
+            // skip fix when variable in assignment pattern is reassigned
+            if (allWriteReferences.length > 2 && parentType === "AssignmentPattern") {
                 return null;
             }
```
mdjermanovic, line 1163: When assignment pattern represents default param value, it has only one write ref for that assignment. For example, in this case there should be no suggestion to remove `x`:

```js
function foo(x = 1) {
    x = 2;
}
```

mdjermanovic: There's also a bug when params with default values are being removed:

```js
function foo(x = 1) {} // fixed to: function foo( = 1) {}
```

mdjermanovic: I think we can simplify this by just checking if all references have the same `range[0]` as `id`.

Tanujkanti4441: >I think we can simplify this by just checking if all references have the same range[0] as id.

This way aren't we leaving suggestion for following code: 
```js
function foo(x = 1) {}  // no suggestion for x
```
is this the intention?

Though i tried a condition to skip fix for `function foo(x = 1) {x = 2;}` please check if this seems correct.

lib/rules/no-unused-vars.js
```diff
@@ -832,6 +836,612 @@ module.exports = {
...
+                // remove unused array element
+                return fixer.removeRange(node.range);
+            }
+
+            /**
+             * fix cases like {a: {k}} or {a: [k]}
+             * @param {ASTNode} node parent node to check
+             * @returns {Object} fixer object
+             */
+            function fixObjectWithValueSeperator(node) {
```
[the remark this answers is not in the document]

lib/rules/no-unused-vars.js
mdjermanovic: > This way aren't we leaving suggestion for following code:
> 
> ```js
> function foo(x = 1) {}  // no suggestion for x
> ```

Yes, this code should have a suggestion to remove `x = 1`.

lib/rules/no-unused-vars.js
```diff
@@ -832,6 +836,622 @@ module.exports = {
...
+
+            // skip fix when function parameters are ressaigned
+            if (isFunction(parent) && allWriteReferences.length > 0 && parent.params.includes(id)) {
+                return null;
+            }
+
+            // skip fix when function parameters with default value are reassigned
+            if (isFunction(parent.parent) && parentType === "AssignmentPattern" && allWriteReferences.length > 1) {
+                return null;
+            }
```
mdjermanovic, line 1178:
```suggestion
            // skip fix when variable has references that would be left behind
            if (allWriteReferences.some(ref => ref.identifier.range[0] !== id.range[0])) {
                return null;
            }
```

To simplify.

lib/rules/no-unused-vars.js
mdjermanovic: I was thinking about something like this: https://github.com/eslint/eslint/pull/18352#discussion_r1864932728

lib/rules/no-unused-vars.js

nzakas:
Hi folks, just wanted to check in because this PR has now been open for almost eight months. Do we feel like we're getting enough value out of this change for the amount of working being put into it?

mdjermanovic (approved):
LGTM, thanks!

PR #18356, opened 2024-04-17
docs: include notes about globals in migration-guide

Grohden:
#### Prerequisites checklist

I'm trying to not get too much frustrated with the migration & migration guides.. so I thought I could contribute on the issues I have faced, hopefully they're welcome 🙏 

[x] Documentation update
[ ] Add autofix to a rule
[ ] Add a CLI option
[ ] Add something to the core
[ ] Other, please explain:

Added a note about the required instalation of the `globals` package (docs don't [make it clear that globals is not a magic package](https://github.com/sindresorhus/globals/issues/244) and that we need to make it a direct dev dependency)

### DIFF
--- a/docs/src/use/configure/migration-guide.md
+++ b/docs/src/use/configure/migration-guide.md
@@ -330,7 +330,9 @@ export default [
 ];
 ```
 
-Note that for this to work or if you get `Global "AudioWorkletGlobalScope " has leading or trailing whitespace`, you'll need to install the latest `globals` package.
+::: tip
+You'll need to install the `globals` package from npm for this example to work. It is not automatically installed by ESLint.
+:::
 
 ### `eslint-env` Configuration Comments
 
--- a/docs/src/use/configure/migration-guide.md
+++ b/docs/src/use/configure/migration-guide.md
@@ -330,6 +330,8 @@ export default [
 ];
 ```
 
+Note that for this to work or if you get `Global "AudioWorkletGlobalScope " has leading or trailing whitespace`, you'll need to install the latest `globals` package.
+
 ### `eslint-env` Configuration Comments
 
 In the eslintrc config system it was possible to use `eslint-env` configuration comments to define globals for a file.


aladdin-add:
> This was fixed 2 years ago, https://github.com/sindresorhus/globals/pull/184. Please upgrade your package.

https://github.com/sindresorhus/globals/issues/239#issuecomment-1968731224

When installing it, you will always get the latest version. This change may be more distracting to users.

aladdin-add:
I would suggest a more appropriate place to be: https://eslint.org/docs/latest/use/troubleshooting/. Let's hear from the others. @eslint/eslint-team

Grohden:
@aladdin-add 

> When installing it, you will always get the latest version. This change may be more distracting to users.

You have to remember that some devs (like me) are stupid 😅, I thought this was some magic global from node or something ("it worked before, I should just change the code to what the guide tells me right?"), so I just imported the package and saw that it was already installed in `node_modules`, so I thought that `eslint` included it as a dependency on its own to facilitate things, then I got that error.

About the troubleshooting:  I'm already trying to find a way to solve the global problem.. why do I have to hunt in another **potentially** unrelated place the solution? For the specific string search on docs (whitespace issue) I guess it makes sense to be there, but saying "hey, this is not magic, we included the package for you before, but now you'll need to install and use it"

aladdin-add:
> About the troubleshooting: I'm already trying to find a way to solve the global problem.. why do I have to hunt in another potentially unrelated place the solution? For the specific string search on docs (whitespace issue) I guess it makes sense to be there, but saying "hey, this is not magic, we included the package for you before, but now you'll need to install and use it yourself" would be helpful

you should always install the package - even if you have not encountered this problem. It may be working without explicitly installing as it may be an indirect dependency, while  your package manager (like npm v3+) just flattened it. But it should be avoided.

docs/src/use/configure/migration-guide.md
```diff
@@ -330,6 +330,8 @@ export default [
 ];
 ```
 
+Note that for this to work or if you get `Global "AudioWorkletGlobalScope " has leading or trailing whitespace`, you'll need to install the latest `globals` package.
```
nzakas, line 333: This is way too specific an error message to include here. We can also format this a bit nicer.

```suggestion
::: tip
You'll need to install the `globals` package from npm for this example to work. It is not automatically installed by ESLint.
:::
```

nzakas (approved):
LGTM. Thanks!

PR #18472, opened 2024-05-18
docs: refactor `prefer-destructuring` rule

Tanujkanti4441:
#### Prerequisites checklist

[x] Documentation update
[ ] Add autofix to a rule
[ ] Add a CLI option
[ ] Add something to the core
[ ] Other, please explain:

refactor `prefer-destructuring` rule docs page to avoid repetition of option's description.

### DIFF
--- a/docs/src/rules/prefer-destructuring.md
+++ b/docs/src/rules/prefer-destructuring.md
@@ -16,20 +16,28 @@ With JavaScript ES6, a new syntax was added for creating variables from an array
 
 This rule takes two sets of configuration objects. The first object parameter determines what types of destructuring the rule applies to.
 
-The two properties, `array` and `object`, can be used to turn on or off the destructuring requirement for each of those types independently. By default, both are true.
+In first set of configuration object the two properties, `array` and `object`, can be used to turn on or off the destructuring requirement for each of those types independently. By default, both are true.
 
-Alternatively, you can use separate configurations for different assignment types. It accepts 2 other keys instead of `array` and `object`.
+```json
+{
+  "rules": {
+    "prefer-destructuring": ["error", {
+      "array": true,
+      "object": true
+    }]
+  }
+}
+```
 
-One key is `VariableDeclarator` and the other is `AssignmentExpression`, which can be used to control the destructuring requirement for each of those types independently. Each property accepts an object that accepts two properties, `array` and `object`, which can be used to control the destructuring requirement for each of `array` and `object` independently for variable declarations and assignment expressions.  By default, `array` and `object` are set to true for both `VariableDeclarator` and `AssignmentExpression`.
+For example, the following configuration enforces only object destructuring, but not array destructuring:
 
-The rule has a second object with a single key, `enforceForRenamedProperties`, which determines whether the `object` destructuring applies to renamed variables.
-
-**Note**: It is not possible to determine if a variable will be referring to an object or an array at runtime. This rule therefore guesses the assignment type by checking whether the key being accessed is an integer. This can lead to the following possibly confusing situations:
-
-* Accessing an object property whose key is an integer will fall under the category `array` destructuring.
-* Accessing an array element through a computed index will fall under the category `object` destructuring.
-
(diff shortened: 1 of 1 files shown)


docs/src/rules/prefer-destructuring.md
```diff
@@ -16,20 +16,28 @@ With JavaScript ES6, a new syntax was added for creating variables from an array
 
 This rule takes two sets of configuration objects. The first object parameter determines what types of destructuring the rule applies to.
```
nzakas, line 17: I think the use of the word "sets" is incorrect throughout. What we are really saying is that this rule takes two arguments, both of which are objects. 

Then we can say, "The first object contains two properties, `array` and `object`..."

And, "The second object..."

What do you think?

Tanujkanti4441: i think it explains the option well,
updated the description!

docs/src/rules/prefer-destructuring.md
```diff
@@ -14,9 +14,9 @@ With JavaScript ES6, a new syntax was added for creating variables from an array
 
 ### Options
 
-This rule takes two sets of configuration objects. The first object parameter determines what types of destructuring the rule applies to.
+This rule takes two arguments, both of which are objects. The first object parameter determines what types of destructuring the rule applies to.
 
-In first set of configuration object the two properties, `array` and `object`, can be used to turn on or off the destructuring requirement for each of those types independently. By default, both are true.
+In first object the two properties, `array` and `object`, can be used to turn on or off the destructuring requirement for each of those types independently. By default, both are true.
```
nzakas, line 19:
```suggestion
In the first object, there are two properties, `array` and `object`, that can be used to turn on or off the destructuring requirement for each of those types independently. By default, both are true.
```

Tanujkanti4441: Updated!

docs/src/rules/prefer-destructuring.md
```diff
@@ -73,138 +81,112 @@ let bar;
 
 :::
 
-Examples of **incorrect** code when `enforceForRenamedProperties` is enabled:
+Alternatively, you can use separate configurations for different assignment types. It accepts 2 other keys instead of `array` and `object`.
```
nzakas, line 84:
```suggestion
Alternatively, you can use separate configurations for different assignment types. The first argument accepts two other keys instead of `array` and `object`.
```

docs/src/rules/prefer-destructuring.md
```diff
@@ -73,138 +81,112 @@ let bar;
...
 
-::: incorrect
+One key is `VariableDeclarator` and the other is `AssignmentExpression`, which can be used to control the destructuring requirement for each of those types independently. Each property accepts an object that accepts two properties, `array` and `object`, which can be used to control the destructuring requirement for each of `array` and `object` independently for variable declarations and assignment expressions.  By default, `array` and `object` are set to true for both `VariableDeclarator` and `AssignmentExpression`.
```
nzakas, line 86:
```suggestion
One key is `VariableDeclarator` and the other is `AssignmentExpression`, which can be used to control the destructuring requirement for each of those types independently. Each property is an object containing two properties, `array` and `object`, which can be used to control the destructuring requirement for each of `array` and `object` independently for variable declarations and assignment expressions.  By default, `array` and `object` are set to `true` for both `VariableDeclarator` and `AssignmentExpression`.
```

docs/src/rules/prefer-destructuring.md
```diff
@@ -73,138 +81,112 @@ let bar;
...
+/* eslint prefer-destructuring: ["error", {AssignmentExpression: {array: true}}] */
+[bar] = array;
 ```
 
 :::
 
-An example configuration, with the defaults `array` and `object` filled in, looks like this:
+#### enforceForRenamedProperties
+
+The rule has a second object with a single key, `enforceForRenamedProperties`, which determines whether the `object` destructuring applies to renamed variables.
```
nzakas, line 129:
```suggestion
The rule has a second object argument with a single key, `enforceForRenamedProperties`, which determines whether the `object` destructuring applies to renamed variables.
```

nzakas (commented):
Just a few last suggestions to clarify wording.

docs/src/rules/prefer-destructuring.md
Tanujkanti4441: Done!

PR #18657, opened 2024-07-05
build: allow any Node 18 LTS

turadg:

[X] Documentation update
[x] Add something to the core

Expand the `engines` range to include the LTS release of Node 18. In dropping support for old versions, https://github.com/eslint/eslint/pull/17725/ dropped support for a versions 18.12-18.17 are part of the Node 18 LTS. 18.18 was the latest version at the time but [18.12 began LTS](https://nodejs.org/en/blog/release/v18.12.0). (as pointed out by @koddsson in https://github.com/eslint/eslint/pull/17725/files#r1447342271 ) 

Lowered the minimum from 18.18 to 18.12. 

What changes should be made to `migrate-to-9.0.0.md` ?


nzakas:
We decided on 18.18.0, and we've propagated that throughout all of our packages. There's no reason to revisit that decision.

turadg:
Thanks for the quick attention.

> We decided on 18.18.0

In the PRs I read it wasn't apparent to me how this was decided. It appeared to me to be an oversight.

> There's no reason to revisit that decision.

Maybe I wan't clear about the motivation. I have a downstream package that has a policy of supporting LTS and that depends on eslint. Eslint's engine won't allow it to install on 18.12 (an LTS release in Maintenance for about another year). Correct me if I'm wrong, but doesn't that make it impossible for a package requiring ESlint (as is) to support LTS releases?

I can understand if you don't think this reason is good enough to allow the PR. I think that implies that it's not the policy of ESlint to support LTS releases. Am I correct in that?

PR #18718, opened 2024-07-26
refactor: Drop dependency on `text-table` to improve performance

jonkoops:


lib/cli-engine/formatters/stylish.js
```diff
@@ -60,21 +104,19 @@ module.exports = function(results) {
...
-                    return stripAnsi(str).length;
+
+                if (message.ruleId) {
+                    row.push(chalk.dim(message.ruleId));
                 }
-            }
+
+                return row;
+            })
         ).split("\n").map(el => el.replace(/(\d+)\s+(\d+)/u, (m, p1, p2) => chalk.dim(`${p1}:${p2}`))).join("\n")}\n\n`;
```
JoshuaKGoldberg, line 120: > `.split("\n")`

The last thing `table()` does is effectively `.join("\n")`...  which is then immediately reverted by this `.split("\n")`. Remove both and the code will be both shorter & less runtime-heavy. ✂️

lib/cli-engine/formatters/stylish.js
```diff
@@ -22,6 +21,51 @@ function pluralize(word, count) {
...
+            if (columnIndex === 0) {
+                result += spacing + column;
+            } else {
+                result += column + (columnIndex === row.length - 1 ? "" : spacing);
+            }
+        }
+
+        if (rowIndex < rows.length - 1) {
+            result += "\n";
+        }
```
JoshuaKGoldberg, line 63: This looks very similar to what an Array `.join("\n")` would look like. You could achieve the same thing with roughly...


```js
return rows
  .map((row, rowIndex) => {
    // ...
  })
  .join("\n")
```

I think doing this would make the code more clear - such as making it more apparent that the `.join("\n")` below is unnecessary. WDYT?

JoshuaKGoldberg (changes requested):
Cool, a great start! Thanks for tackling this so quickly! 🚀 

I think we can get even more runtime performance improvements by removing the `.join("\n") ... .split("\n")` equivalent.

jonkoops:
I thought I opened this up on my fork, got some more enhancements coming up so I'll convert this to a draft.

lib/cli-engine/formatters/stylish.js
jonkoops: Yeah, I am planning to in-line the table code even more and switch this logic over to pure string concatenation for performance reasons.

lib/cli-engine/formatters/stylish.js
jonkoops: I'd like to avoid using `map()` altogether as it introduces a new array under the hood. Handling the newlines by appending them directly increases performance, even if it's not as readable.

That said, I am considering always unconditionally appending a newline at the end stripping the last one after the loop, that removes a condition in the `for` loop, which might boost performance even more.

aladdin-add:
I know this is a wip, but since this is a performance optimization, could you please provide a comparison of performance data before and after the change when you are done so that team members can evaluate it, thank you!🙏

nzakas:
@aladdin-add we're just looking at removing a dependency, so perf isn't the key factor here.

PR #18754, opened 2024-08-05
feat: report duplicate allowed flags in `no-invalid-regexp`

### LINKED ISSUE #18748: Bug: `no-invalid-regexp` not reporting repeted flag allowed by `allowConstructorFlags`
### Environment

Node version: latest
npm version: latest
Local ESLint version: latest
Global ESLint version: no
Operating System: windows

Default (Espree)

### What did you do?

```js
/*eslint no-invalid-regexp: ["error", { "allowConstructorFlags": ["a"] }]*/

new RegExp('.', 'aa');
```

report `aa` as invalid flag.

### What actually happened?

rule doesn't report this as invalid supplied flag as it does with other predefined flags.

https://eslint.org/play/#eyJ0ZXh0IjoiLyplc2xpbnQgbm8taW52YWxpZC1yZWdleHA6IFtcImVycm9yXCIsIHsgXCJhbGxvd0NvbnN0cnVjdG9yRmxhZ3NcIjogW1wiYVwiXSB9XSovXG5cbm5ldyBSZWdFeHAoJy4nLCAnYWEnKTsiLCJvcHRpb25zIjp7InJ1bGVzIjp7fSwibGFuZ3VhZ2VPcHRpb25zIjp7InNvdXJjZVR5cGUiOiJtb2R1bGUiLCJwYXJzZXJPcHRpb25zIjp7ImVjbWFGZWF0dXJlcyI6e319fX19

### Participation

### Additional comments

_No response_

Tanujkanti4441:
#### Prerequisites checklist

[ ] Documentation update
[ ] Add autofix to a rule
[ ] Add a CLI option
[ ] Add something to the core
[ ] Other, please explain:

`no-invalid-regexp` rule  doesn't report duplicate flags that are allowed by the `allowConstructorFlags` option.
```js
/*eslint no-invalid-regexp: ["error", { "allowConstructorFlags": ["a"] }]*/

new RegExp('.', 'aa'); // not reported
```
made the rule to report these flags
Fixes #18748

### DIFF
--- a/lib/rules/no-invalid-regexp.js
+++ b/lib/rules/no-invalid-regexp.js
@@ -55,7 +55,7 @@ module.exports = {
             const temp = options.allowConstructorFlags.join("").replace(new RegExp(`[${validFlags}]`, "gu"), "");
 
             if (temp) {
-                allowedFlags = temp.split("");
+                allowedFlags = [...new Set(temp)];
             }
         }
 
--- a/lib/rules/no-invalid-regexp.js
+++ b/lib/rules/no-invalid-regexp.js
@@ -55,7 +55,7 @@ module.exports = {
             const temp = options.allowConstructorFlags.join("").replace(validFlags, "");
 
             if (temp) {
-                allowedFlags = new RegExp(`[${temp}]`, "gu");
+                allowedFlags = temp.split("");
             }
         }
 
@@ -154,13 +154,16 @@ module.exports = {
                     return;
                 }
 
-                let flags = getFlags(node);
+                const flags = getFlags(node);
+                let flagsToCheck = flags;
 
-                if (flags && allowedFlags) {
-                    flags = flags.replace(allowedFlags, "");
+                if (flags && allowedFlags && allowedFlags.length > 0) {
+                    allowedFlags.forEach(flag => {
+                        flagsToCheck = flagsToCheck.replace(new RegExp(flag, "u"), "");
+                    });
                 }
 
-                let message = validateRegExpFlags(flags);
+                let message = validateRegExpFlags(flagsToCheck);
 
                 if (message) {
                     report(node, message);
--- a/lib/rules/no-invalid-regexp.js
+++ b/lib/rules/no-invalid-regexp.js
@@ -159,7 +159,7 @@ module.exports = {
 
                 if (flags && allowedFlags && allowedFlags.length > 0) {
                     allowedFlags.forEach(flag => {
-                        flagsToCheck = flagsToCheck.replace(new RegExp(flag, "u"), "");
+                        flagsToCheck = flagsToCheck.replace(flag, "");
                     });
                 }
 
--- a/lib/rules/no-invalid-regexp.js
+++ b/lib/rules/no-invalid-regexp.js
@@ -10,7 +10,7 @@
 
 const RegExpValidator = require("@eslint-community/regexpp").RegExpValidator;
 const validator = new RegExpValidator();
-const validFlags = /[dgimsuvy]/gu;
+const validFlags = "dgimsuvy";
 const undefined1 = void 0;
 
 //------------------------------------------------------------------------------
@@ -49,10 +49,10 @@ module.exports = {
     create(context) {
 
         const options = context.options[0];
-        let allowedFlags = null;
+        let allowedFlags = [];
 
         if (options && options.allowConstructorFlags) {
-            const temp = options.allowConstructorFlags.join("").replace(validFlags, "");
+            const temp = options.allowConstructorFlags.join("").replace(new RegExp(`[${validFlags}]`, "gu"), "");
 
             if (temp) {
                 allowedFlags = temp.split("");
@@ -125,27 +125,25 @@ module.exports = {
         /**
          * Check syntax error in a given flags.
          * @param {string|null} flags The RegExp flags to validate.
+         * @param {string|null} flagsToCheck The RegExp invalid flags.
          * @returns {string|null} The syntax error.
          */
-        function validateRegExpFlags(flags) {
-            if (!flags) {
-                return null;
-            }
-            try {
-                validator.validateFlags(flags);
-            } catch {
-                return `Invalid flags supplied to RegExp constructor '${flags}'`;
-            }
+        function validateRegExpFlags(flags, flagsToCheck) {
 
             /*
              * `regexpp` checks the combination of `u` and `v` flags when parsing `Pattern` according to `ecma262`,
              * but this rule may check only the flag when the pattern is unidentifiable, so check it here.
              * https://tc39.es/ecma262/multipage/text-processing.html#sec-parsepattern
              */
-            if (flags.includes("u") && flags.includes("v")) {
+            if (flags && flags.includes("u") && flags.includes("v")) {
                 return "Regex 'u' and 'v' flags cannot be used together";
             }
-            return null;
+
+            if (!flagsToCheck) {
+                return null;
+            }
+
+            return `Invalid flags supplied to RegExp constructor '${flagsToCheck}'`;
         }
 
         return {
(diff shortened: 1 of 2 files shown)


lib/rules/no-invalid-regexp.js
```diff
@@ -154,13 +154,16 @@ module.exports = {
...
 
-                let flags = getFlags(node);
+                const flags = getFlags(node);
+                let flagsToCheck = flags;
 
-                if (flags && allowedFlags) {
-                    flags = flags.replace(allowedFlags, "");
+                if (flags && allowedFlags && allowedFlags.length > 0) {
+                    allowedFlags.forEach(flag => {
+                        flagsToCheck = flagsToCheck.replace(new RegExp(flag, "u"), "");
```
aladdin-add, line 162:
```suggestion
                        flagsToCheck = flagsToCheck.replace(flag, "");
```

tests/lib/rules/no-invalid-regexp.js
```diff
@@ -213,6 +213,51 @@ ruleTester.run("no-invalid-regexp", rule, {
...
+        },
+        {
+            code: "new RegExp('.', 'aga');",
+            options: [{ allowConstructorFlags: ["a"] }],
+            errors: [{
+                messageId: "regexMessage",
+                data: { message: "Invalid flags supplied to RegExp constructor 'ga'" },
+                type: "NewExpression"
+            }]
+        },
```
aladdin-add, line 260: can you add a test like:
```js
            code: "new RegExp('.', 'auu');",
            options: [{ allowConstructorFlags: ["a"] }],
```

lib/rules/no-invalid-regexp.js
Tanujkanti4441: Done!

tests/lib/rules/no-invalid-regexp.js
Tanujkanti4441: Added!

aladdin-add (dismissed):

Tanujkanti4441:
@aladdin-add, right now it is allowed to pass duplicate strings to `allowConstructorFlags` which can also lead to allow duplicate flags in regex constructor i think this should also be fixed or is it intentional?

aladdin-add:
IMHO, it seems a bug - I didn't see it's covered by the tests. it can be easily fixed by adding `uniqueItems: true`, however it's a breaking change, so we have to wait until eslint v10. Can you create another issue?

Tanujkanti4441:
> IMHO, it seems a bug - I didn't see it's covered by the tests. it can be easily fixed by adding `uniqueItems: true`, however it's a breaking change, so we have to wait until eslint v10. Can you create another issue?

have created the issue => #18755

tests/lib/rules/no-invalid-regexp.js
```diff
@@ -213,6 +213,60 @@ ruleTester.run("no-invalid-regexp", rule, {
...
+                data: { message: "Invalid flags supplied to RegExp constructor 'z'" },
+                type: "NewExpression"
+            }]
+        },
+        {
+            code: "new RegExp('.', 'aga');",
+            options: [{ allowConstructorFlags: ["a"] }],
+            errors: [{
+                messageId: "regexMessage",
+                data: { message: "Invalid flags supplied to RegExp constructor 'ga'" },
```
fasttime, line 257: I'm not sure this error message is clear enough. `"g"` is a valid flag and `"a"` was allowed in the options, so it's not obvious why `'ga'` would be invalid. If `a` is replaced with a valid flag like `u`, the error message includes the flag duplication but also the valid flag `"g"`, which is still inconsistent:

```js
new RegExp('.', 'ugu');
// Invalid flags supplied to RegExp constructor 'ugu'
```

Maybe there should be a new message for duplicated flags?

Tanujkanti4441: seems this case is not only with the duplicate flags:
```js
/*eslint no-invalid-regexp: ["error", { "allowConstructorFlags": ["a"] }]*/

new RegExp('.', 'aumo');
// Invalid flags supplied to RegExp constructor 'umo.'
```

[palyground](https://eslint.org/play/#eyJ0ZXh0IjoiLyplc2xpbnQgbm8taW52YWxpZC1yZWdleHA6IFtcImVycm9yXCIsIHsgXCJhbGxvd0NvbnN0cnVjdG9yRmxhZ3NcIjogW1wiYVwiXSB9XSovXG5cbm5ldyBSZWdFeHAoJy4nLCAnYXVtbycpOyIsIm9wdGlvbnMiOnsicnVsZXMiOnt9LCJsYW5ndWFnZU9wdGlvbnMiOnsic291cmNlVHlwZSI6Im1vZHVsZSIsInBhcnNlck9wdGlvbnMiOnsiZWNtYUZlYXR1cmVzIjp7fX19fX0=)

lib/rules/no-invalid-regexp.js
```diff
@@ -154,13 +154,16 @@ module.exports = {
...
-                if (flags && allowedFlags) {
-                    flags = flags.replace(allowedFlags, "");
+                if (flags && allowedFlags && allowedFlags.length > 0) {
+                    allowedFlags.forEach(flag => {
+                        flagsToCheck = flagsToCheck.replace(flag, "");
+                    });
                 }
 
-                let message = validateRegExpFlags(flags);
+                let message = validateRegExpFlags(flagsToCheck);
```
fasttime, line 166: Maybe we should pass both the original and the filtered flags to the validation function:

```suggestion
                let message = validateRegExpFlags(flags, flagsToCheck);
```

In `validateRegExpFlags()` we could then validate the `flagsToCheck`, but if the validation fails, we could include all the original flags in the error message.

```js
        function validateRegExpFlags(flags, flagsToCheck) {
            if (!flagsToCheck) {
                return null;
            }
            try {
                validator.validateFlags(flagsToCheck);
            } catch {
                return `Invalid flags supplied to RegExp constructor '${flags}'`;
            }
```

This would retain the current error message in the case of predefined flags, and also remove the inconsistency when the message includes custom flags.

An additional error message for duplicated flags would be nice, but that could be added as a feature in another PR.

Tanujkanti4441: sounds good, but how about if we only show those flags in error message that are invalid? i think that can be helpful to find which flags we need to remove.

fasttime: > sounds good, but how about if we only show those flags in error message that are invalid? i think that can be helpful to find which flags we need to remove.

Makes sense, as long as we also show duplicated flags.

tests/lib/rules/no-invalid-regexp.js
```diff
@@ -213,6 +213,60 @@ ruleTester.run("no-invalid-regexp", rule, {
...
+        },
+        {
+            code: "new RegExp('.', 'uu');",
+            options: [{ allowConstructorFlags: ["u"] }],
+            errors: [{
+                messageId: "regexMessage",
+                data: { message: "Invalid flags supplied to RegExp constructor 'u'" },
+                type: "NewExpression"
+            }]
+        },
```
fasttime, line 269: The current error message for this invalid case is better, because it shows the duplication:

> Invalid flags supplied to RegExp constructor 'uu'.

Can we keep the duplicated flags in the error message until we have a more specific message?

[**Playground link**](https://eslint.org/play/#eyJ0ZXh0IjoiLyplc2xpbnQgbm8taW52YWxpZC1yZWdleHA6IFwiZXJyb3JcIiovXG5cbm5ldyBSZWdFeHAoJy4nLCAndXUnKTsiLCJvcHRpb25zIjp7InJ1bGVzIjp7fSwibGFuZ3VhZ2VPcHRpb25zIjp7InNvdXJjZVR5cGUiOiJtb2R1bGUiLCJwYXJzZXJPcHRpb25zIjp7ImVjbWFGZWF0dXJlcyI6e319fX19)

mdjermanovic:
I changed the tag to `feat` because this produces more lint errors.

tests/lib/rules/no-invalid-regexp.js
Tanujkanti4441: Done! sorry for such late wasn't active due to exams.

fasttime: Thanks for the update. Hope your exam went well.

tests/lib/rules/no-invalid-regexp.js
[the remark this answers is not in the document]
Tanujkanti4441: You mean we should make the rule to report duplicate flags even in `{allowConstructorFlags: ["a", "a"] }` this case? Because right now the above case is valid according to this PR's logic.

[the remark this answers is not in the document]
fasttime: > You mean we should make the rule to report duplicate flags even in `{allowConstructorFlags: ["a", "a"] }` this case?

No, but we should report the duplicated flag in `"new RegExp('.', 'aa');"`. It shouldn't matter how many times it is specified in the `allowConstructorFlags` option.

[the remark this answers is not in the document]
Tanujkanti4441: Done!

fasttime (approved):
LGTM, thanks!

PR #18756, opened 2024-08-06
feat: add new rule no-useless-arrow-block

leonheess:
#### Prerequisites checklist

[ ] Documentation update
[ ] Add autofix to a rule
[ ] Add a CLI option
[ ] Add something to the core
[ ] Other, please explain:

**What should the new rule do?**
Arrow functions with a block body containing only a single return statement or a single expression statement can be simplified by removing the braces and, in the case of a return statement, the return keyword. This simplification can improve code readability.

Arrow function expressions

**What category of rule is this?**

[ ] Warns about a potential problem
[x] Suggests an alternate way of doing something

**Please provide some example JavaScript code that this rule will warn about:**

```js
/* eslint no-useless-arrow-block: "error" */

const foo = () => { return 5; };

const bar = (x) => { return x * 2; };

const baz = () => { console.log('Hello'); };

const qux = (a, b) => {
  return a + b;
};

```

It promotes code readability and simplicity by encouraging the use of concise arrow function syntax. Given that arrow functions are extremely widely used in modern JavaScript development, simplifying their syntax reduces visual clutter and makes the code easier to understand at a glance, aligning with ESLint's core goals.

### DIFF
--- a/docs/src/rules/no-useless-arrow-block.md
+++ b/docs/src/rules/no-useless-arrow-block.md
@@ -1,0 +1,72 @@
+---
+title: no-useless-arrow-block
+rule_type: suggestion
+---
+
+
+
+Arrow functions with a block body containing only a single return statement or a single expression statement can be simplified by removing the braces and, in the case of a return statement, the return keyword. This simplification can improve code readability.
+
+## Rule Details
+
+This rule aims to report and fix arrow functions that can have their block body simplified.
+
+Examples of **incorrect** code for this rule:
+
+::: incorrect
+
+```js
+/* eslint no-useless-arrow-block: "error" */
+
+const foo = () => { return 5; };
+
+const bar = (x) => { return x * 2; };
+
+const baz = () => { console.log('Hello'); };
+
+const qux = (a, b) => {
+  return a + b;
+};
+
+```
+
+:::
+
+Examples of **correct** code for this rule:
+
+::: correct
+
+```js
+/* eslint no-useless-arrow-block: "error" */
+
+const foo = () => 5;
+
+const bar = (x) => x * 2;
+
+const baz = () => console.log('Hello');
+
+const qux = (a, b) => a + b;
+
+const complex = () => {
+  doSomething();
+  return result;
+};
+
+const multiLine = () => (
+  veryLongExpression +
+  thatNeedsToBeWrapped
+);
+
+class MyClass {
+  static myMethod = () => {
+    this.doSomething();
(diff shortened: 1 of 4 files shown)


aladdin-add:
Thanks for your interest in improving ESLint. As of 2020, unfortunately, we have decided to [only accept rules related to new ECMAScript features](https://eslint.org/blog/2020/05/changes-to-rules-policies). We prefer that new rules be implemented in plugins, so I'm closing it.

Since ESLint is pluggable and can load custom rules at runtime, it doesn't need to be a blocker for you using this in your project, if you'd find it useful. It just means that you would need to [implement the rule](http://eslint.org/docs/developer-guide/working-with-rules) yourself, rather than using a bundled rule that is packaged with ESLint.

leonheess:
@aladdin-add I sure hope you did not once in your life complain about "devs these days not wanting to contribute to FOSS anymore" because you are the personification of the reason why. Even if you had no say in creating this ridiculous rule you are still the one not even considering the possibility of an exception and closing this PR without even reading past the headline

Tanujkanti4441 (commented):
There already a rule exists to disallow or allow blocks in arrow functions 
https://eslint.org/docs/latest/rules/arrow-body-style

leonheess:
@Tanujkanti4441 Unfortunately, with this rule you cannot disallow useless body braces. I wrote a plugin for it: https://www.npmjs.com/package/eslint-plugin-arrowsmith

aladdin-add:
I do feel its value, and putting it in a 3rd-party plugin is a more appropriate choice! 🎉

Tanujkanti4441:
> @Tanujkanti4441 Unfortunately, with this rule you cannot disallow useless body braces. I wrote a plugin for it: https://www.npmjs.com/package/eslint-plugin-arrowsmith

@leonheess, thanks for this plugin this will be helpful  for ESLint users.

though i feel most of the useless body braces are reported by the `arrow-body-style` rule when `as-needed` option is used but while checking the tests added in this PR with the rule i found this particular case i not reported by the rule.

```js
/*eslint arrow-body-style: ["error", "as-needed"]*/

() => {console.log('Hello');}
```

it is reported by other option and also

```js
/*eslint arrow-body-style: ["error", "always"]*/

() => console.log('Hello');

//fix this to 

() => {return console.log('Hello')};
```

which seems incorrect 

@leonheess would you like to help us by fixing this, i think opening an issue would be a great way to start.

Issue #17720, opened 2023-11-06
DeprecationWarning: Punycode


jasperkelder:
### ESLint version

8.53.0

### What problem do you want to solve?

One of eslint's dependencies is using a deprecated module, we're currently getting the following warning -> DeprecationWarning: The `punycode` module is deprecated. Please use a userland alternative instead.

└─┬ eslint@8.53.0
  └─┬ ajv@6.12.6
    └─┬ uri-js@4.4.1
      └── punycode@2.3.0

### What do you think is the correct solution?

Make sure the deprecated module gets replaced in the dependency tree.

Rec0iL99:
Hi @jasperkelder, thanks for the issue. I don't think this is something to be solved in ESLint. ESLint directly depends on `ajv` and there is an [open issue](https://github.com/ajv-validator/ajv/issues/2343) addressing this there.

domdomegg:
I've opened a PR to fix this in uri-js, the package ajv depends on here: https://github.com/garycourt/uri-js/pull/95

newinfinity12:
> I've opened a PR to fix this in uri-js, the package ajv depends on here: [garycourt/uri-js#95](https://github.com/garycourt/uri-js/pull/95)
@domdomegg 
still not fixed. would also like to note that eslint uses ajv v6.12.6 instead of the latest v8.12.0 from a year ago

Rec0iL99:
> would also like to note that eslint uses ajv v6.12.6 instead of the latest v8.12.0 from a year ago

@newinfinity12 perhaps this https://github.com/eslint/eslint/issues/13888#issuecomment-872591875 will help you understand why ESLint still uses `Ajv 6`.

newinfinity12:
> > would also like to note that eslint uses ajv v6.12.6 instead of the latest v8.12.0 from a year ago
> 
> @newinfinity12 perhaps this [#13888 (comment)](https://github.com/eslint/eslint/issues/13888#issuecomment-872591875) will help you understand why ESLint still uses `Ajv 6`.

@Rec0iL99
great resource however that does not change the fact that eslint is indirectly using a deprecated dependency.

Based on the threads created in Ajv and URI.js the blame/fix keeps getting pushed around to different repos. Ajv has not been updated in over a year and URI.js in over 3 years. So based on this info it looks like eslint should switch to a different repo to avoid the deprecation warning altogether

domdomegg:
[...]

PR #18762, opened 2024-08-08
fix: override uri-js with fast-uri

JudahZF:
By overriding uri-js with fasturi, the punycode deprecation can be fixed (#17720) without having to upgrade ajv to version 8 (#13888: comment-872591875)

#### Prerequisites checklist

[ ] Documentation update
[ ] Add autofix to a rule
[ ] Add a CLI option
[ ] Add something to the core
[ ] Other, please explain:

* **Node version:** v22.6.0
* **npm version:** v10.2.4
* **Local ESLint version:**
* **Global ESLint version:**
* **Operating System:** linux 6.10.3-arch1-2

[ ] `Default (Espree)`
[X] `@typescript-eslint/parser`
[ ] `@babel/eslint-parser`
[ ] `vue-eslint-parser`
[ ] `Other`

<details>

```js

```

</details>

NodeJS repeatedly logs punycode deprecation error

Added an override to replace uri-js (dependency of ajv@6) with fast-uri (version now used by ajv@8)


package.json
```diff
@@ -165,6 +165,11 @@
     "webpack-cli": "^4.5.0",
     "yorkie": "^2.0.0"
   },
+  "overrides": {
+  	"ajv": {
+      "uri-js": "npm:fast-uri@^2.3.0"
```
nzakas, line 170: Is this the correct syntax? Do we need "npm:" here?

JudahZF: Yes, the npm is required to tell the package.json that it should retrive a different package

JudahZF:
Looks like there are some other dependencies that include punycode, but only show up after overriding this one. I will look into over-riding them over the next week.

JudahZF:
So it looks like the other packages that use punycode use a different version (punycode.js) that still triggers the warning. I would still recommend that this goes through as it reduces the number of dependencies that use punycode and brings the uri dependency in-line with current versions of ajv.

nzakas:
If this is just to fix one deprecation warning, then I don't think it's very valuable. Deprecations aren't bad and I don't think deprecation hunting in our dependency tree is a good use of time.

aladdin-add:
Note that not all package managers support `overrides`. A better way is adding it in your own projects:

```json
  "overrides": {
    "eslint/ajv/uri-js": "npm:fast-uri@^2.3.0"
  },
```

HowieG:
@nzakas for what it's worth this user noted there may be a security vuln downstream of uri-js

> is is not only that there is deprecated punycode dependency, but also security scan shows issue with old version of braces library:
The NPM package braces, versions prior to 3.0.3, fails to limit the number of characters it can handle, which could lead to Memory Exhaustion. In lib/parse.js, if a malicious user sends "imbalanced braces" as input, the parsing will enter a loop, which will cause the program to start allocating heap memory without freeing it at any moment of the loop. Eventually, the JavaScript heap limit is reached, and the program will crash.

old ajv use old uri-js that use old babel-cli that use old chokidar that use old anymatch that use old micromatch that finally use old braces

https://github.com/eslint/eslint/issues/17733#issuecomment-2277264699

eric-net (approved):

nzakas:
@HowieG that use case doesn't happen in ESLint.

jupiTAr88 (commented):
!

PR #18793, opened 2024-08-20
Revert "chore: add tab index to powered by algolia text"

nzakas:
We should not remove tab accessibility from links. Let's think of a different way to solve this problem.

Reverts eslint/eslint#18792


amareshsm (approved):
LGTM

fasttime (approved):
Okay, let's revert the change and try another solution.

PR #18818, opened 2024-08-26
refactor: Extract processor logic into ProcessorService

nzakas:
#### Prerequisites checklist

[ ] Documentation update
[ ] Add autofix to a rule
[ ] Add a CLI option
[x] Add something to the core
[ ] Other, please explain:

This extracts processor logic into `ProcessService`:

- Created `ProcessorService`
- Refactored `Linter` to use `ProcessorService`
- Added `rawBody` to `VFile`
- Updated `VFile` tests

The biggest difference is that `ProcessorService` takes a `VFile` and returns an array of `VFile`s (or strings for legacy compat). This makes it easier to pass `VFile` instances around without worrying if it's from a processor or not.

- I'm unsure if the parameter order for `ProcessorService#postprocessSync()` makes sense. I like the idea of always having `config` as the last parameter, but would like some other opinions.
- I think we can avoid adding `rawBody` to `VFile` if we always pass BOM-less file contents to a processor. See inline comment.

### DIFF
--- a/lib/linter/linter.js
+++ b/lib/linter/linter.js
@@ -54,6 +54,7 @@ const { LATEST_ECMA_VERSION } = require("../../conf/ecma-version");
 const { VFile } = require("./vfile");
 const { ParserService } = require("../services/parser-service");
 const { FileContext } = require("./file-context");
+const { ProcessorService } = require("../services/processor-service");
 const STEP_KIND_VISIT = 1;
 const STEP_KIND_CALL = 2;
 
@@ -1292,27 +1293,18 @@ class Linter {
     }
 
     /**
-     * Same as linter.verify, except without support for processors.
-     * @param {string|SourceCode} textOrSourceCode The text to parse or a SourceCode object.
+     * Lint using eslintrc and without processors.
+     * @param {VFile} file The file to lint.
      * @param {ConfigData} providedConfig An ESLintConfig instance to configure everything.
      * @param {VerifyOptions} [providedOptions] The optional filename of the file being checked.
      * @throws {Error} If during rule execution.
      * @returns {(LintMessage|SuppressedLintMessage)[]} The results as an array of messages or an empty array if no messages.
      */
-    _verifyWithoutProcessors(textOrSourceCode, providedConfig, providedOptions) {
+    #eslintrcVerifyWithoutProcessors(file, providedConfig, providedOptions) {
+
         const slots = internalSlotsMap.get(this);
         const config = providedConfig || {};
         const options = normalizeVerifyOptions(providedOptions, config);
-        let text;
-
-        // evaluate arguments
-        if (typeof textOrSourceCode === "string") {
-            slots.lastSourceCode = null;
-            text = textOrSourceCode;
-        } else {
-            slots.lastSourceCode = textOrSourceCode;
-            text = textOrSourceCode.text;
-        }
 
         // Resolve parser.
         let parserName = DEFAULT_PARSER_NAME;
@@ -1339,7 +1331,7 @@ class Linter {
 
         // search and apply "eslint-env *".
         const envInFile = options.allowInlineConfig && !options.warnInlineConfig
-            ? findEslintEnv(text)
+            ? findEslintEnv(file.body)
             : {};
         const resolvedEnvConfig = Object.assign({ builtin: true }, config.env, envInFile);
         const enabledEnvs = Object.keys(resolvedEnvConfig)
@@ -1354,9 +1346,6 @@ class Linter {
             globals: config.globals,
             parser,
             parserOptions
-        });
-        const file = new VFile(options.filename, text, {
-            physicalPath: providedOptions.physicalFilename
         });
 
         if (!slots.lastSourceCode) {
@@ -1468,6 +1457,36 @@ class Linter {
                 .sort((problemA, problemB) => problemA.line - problemB.line || problemA.column - problemB.column),
             reportUnusedDisableDirectives: options.reportUnusedDisableDirectives
         });
+
+    }
+
+    /**
+     * Same as linter.verify, except without support for processors.
+     * @param {string|SourceCode} textOrSourceCode The text to parse or a SourceCode object.
+     * @param {ConfigData} providedConfig An ESLintConfig instance to configure everything.
+     * @param {VerifyOptions} [providedOptions] The optional filename of the file being checked.
+     * @throws {Error} If during rule execution.
+     * @returns {(LintMessage|SuppressedLintMessage)[]} The results as an array of messages or an empty array if no messages.
+     */
+    _verifyWithoutProcessors(textOrSourceCode, providedConfig, providedOptions) {
+        const slots = internalSlotsMap.get(this);
+        const filename = normalizeFilename(providedOptions.filename || "<input>");
+        let text;
+
+        // evaluate arguments
+        if (typeof textOrSourceCode === "string") {
+            slots.lastSourceCode = null;
+            text = textOrSourceCode;
+        } else {
+            slots.lastSourceCode = textOrSourceCode;
+            text = textOrSourceCode.text;
+        }
+
+        const file = new VFile(filename, text, {
+            physicalPath: providedOptions.physicalFilename
+        });
+
+        return this.#eslintrcVerifyWithoutProcessors(file, providedConfig, providedOptions);
     }
 
     /**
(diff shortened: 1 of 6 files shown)


lib/linter/linter.js
```diff
@@ -1297,27 +1298,18 @@ class Linter {
...
-     * @param {string|SourceCode} textOrSourceCode The text to parse or a SourceCode object.
+     * Lint using eslintrc and without processors.
+     * @param {VFile} file The file to lint.
      * @param {ConfigData} providedConfig An ESLintConfig instance to configure everything.
      * @param {VerifyOptions} [providedOptions] The optional filename of the file being checked.
      * @throws {Error} If during rule execution.
      * @returns {(LintMessage|SuppressedLintMessage)[]} The results as an array of messages or an empty array if no messages.
      */
-    _verifyWithoutProcessors(textOrSourceCode, providedConfig, providedOptions) {
+    #eslintrcVerifyWithoutProcessors(file, providedConfig, providedOptions) {
```
nzakas, line 1308: Created this to avoid creating a second `VFile` for each block.

lib/linter/linter.js
```diff
@@ -1542,102 +1561,87 @@ class Linter {
...
-     * @param {string|SourceCode} textOrSourceCode The text to parse or a SourceCode object.
+     * Verify using flat config and without any processors.
+     * @param {VFile} file The file to lint.
      * @param {FlatConfig} providedConfig An ESLintConfig instance to configure everything.
      * @param {VerifyOptions} [providedOptions] The optional filename of the file being checked.
      * @throws {Error} If during rule execution.
      * @returns {(LintMessage|SuppressedLintMessage)[]} The results as an array of messages or an empty array if no messages.
      */
-    _verifyWithFlatConfigArrayAndWithoutProcessors(textOrSourceCode, providedConfig, providedOptions) {
+    #flatVerifyWithoutProcessors(file, providedConfig, providedOptions) {
```
nzakas, line 1640: Same here. Wanted to avoid creating two `VFile`s for the same block.

tests/fixtures/subdir-only-config/foo
mdjermanovic, line 1: This file was probably accidentally left here after testing https://github.com/eslint/eslint/pull/18742#discussion_r1741781595

lib/linter/linter.js
```diff
@@ -1537,102 +1556,86 @@ class Linter {
...
                 return [];
             }
 
             // Resolve configuration again if the file content or extension was changed.
-            if (configForRecursive && (text !== blockText || path.extname(blockName) !== originalExtname)) {
+            if (configForRecursive && (text !== block.body || path.extname(block.path) !== originalExtname)) {
                 debug("Resolving configuration again because the file content or extension was changed.");
                 return this._verifyWithFlatConfigArray(
-                    blockText,
+                    block.body,
```
mdjermanovic, line 1607: Related to https://github.com/eslint/eslint/pull/18818#discussion_r1742470825

Before this change, with this config file:

```js
// eslint.config.js
module.exports = [{
    files: ["**/*.myjs"],
    processor: {
        preprocess(text, filename) {
            return [{ text, filename }];
        },
        postprocess(messages) {
            return [].concat(...messages);
        },
        supportsAutofix: true
    },
    rules: {
        "unicode-bom": ["error", "never"]
    }
}];
```

and this test file (has BOM):

```js
// test.myjs

var x = "This file has BOM";

```

Running `eslint test.myjs` would report that the file has BOM, and running `eslint test.myjs --fix` would remove it.

After this change, `eslint test.myjs` doesn't report that the file has BOM.

nzakas: Oops yeah, that would have to be `rawBody` to make this work.

nzakas:
It seems like processors not checking for BOMs is a likely source of bugs, so I'm more in favor of passing BOM-less text to processors now. What do you think?

nzakas:
I've updated the PR so that BOMs are not passed to processors any more. If we decide not to move forward with that, I can revert that commit, but I do think this is the right choice based on previous discussion.

mdjermanovic:
> I've updated the PR so that BOMs are not passed to processors any more.

Makes sense to me. It seems unlikely that a processor would particularly want to know if the original file has BOM.

However, I think we should ensure that, when the processor returns a block with the entire content that was passed to it (in order to lint the file as a whole + its code blocks), the original file with BOM is linted. The example in https://github.com/eslint/eslint/pull/18818#discussion_r1743608645 still doesn't work with the latest version of this branch.

nzakas:
> However, I think we should ensure that, when the processor returns a block with the entire content that was passed to it (in order to lint the file as a whole + its code blocks), the original file with BOM is linted. The example in [#18818 (comment)](https://github.com/eslint/eslint/pull/18818#discussion_r1743608645) still doesn't work with the latest version of this branch.

I think the only way to achieve this would be to add a third parameter to `preprocess` that indicates if there's a BOM. That way, processors could decide whether or not they wanted to pass the BOM along in their code blocks. That preserves the benefits of stripping the BOM from the text so processors don't forget to check while allowing processors to still opt-in to checking for it if they so desire.

(I think in an ideal world, processors would accept and return `VFile` objects, which would carry that additional information. But this would mean a v2 processors API just to support this use case, and I'm not sure it's worth that much effort.)

nzakas:
The failure is just the pull request labeler. No idea what's going on with that.

sam3k:
Per 2024-09-05 TSC meeting, we will preserve the current behavior and @nzakas will open an issue to discuss changing it in `v10`

mdjermanovic:
Sorry for the delay, I'll review this soon.

lib/linter/linter.js
```diff
@@ -1537,102 +1556,86 @@ class Linter {
...
             // Skip this block if filtered.
-            if (!filterCodeBlock(blockName, blockText)) {
+            if (!filterCodeBlock(block.path, block.body)) {
                 debug("This code block was skipped.");
                 return [];
             }
 
             // Resolve configuration again if the file content or extension was changed.
-            if (configForRecursive && (text !== blockText || path.extname(blockName) !== originalExtname)) {
+            if (configForRecursive && (text !== block.body || path.extname(block.path) !== originalExtname)) {
```
mdjermanovic, line 1604: The example from https://github.com/eslint/eslint/pull/18818#discussion_r1743608645 still doesn't work, because `text !== block.body` is `true` when file has BOM so it enters this code path and lints `block.body` which doesn't include BOM.

nzakas:
I went back to the original implementation that preserve the existing BOM behavior and added a test for `Linter` that validates the BOM is being passed.

lib/linter/linter.js
```diff
@@ -2057,77 +2088,78 @@ class Linter {
...
```
mdjermanovic, line 2137:
```suggestion
            if (configForRecursive && (text !== block.rawBody || path.extname(block.path) !== originalExtname)) {
```

Same as https://github.com/eslint/eslint/pull/18818#discussion_r1764868361

lib/linter/linter.js
```diff
@@ -2057,77 +2088,78 @@ class Linter {
...
-                    nodeType: null
-                }
-            ];
-        }
+        const { files } = preprocessResult;
 
-        const messageLists = blocks.map((block, i) => {
-            debug("A code block was found: %o", block.filename || "(unnamed)");
+        const messageLists = files.map(block => {
+            debug("A code block was found: %o", block.path ?? "(unnamed)");
```
mdjermanovic, line 2123:
```suggestion
            debug("A code block was found: %o", block.path);
```

Same as https://github.com/eslint/eslint/pull/18818#discussion_r1764787354, though I think it would be better to move both into `ProcessorService#preprocessSync()`.

lib/linter/linter.js
```diff
@@ -1537,102 +1556,86 @@ class Linter {
...
-                blockText,
+            return this.#flatVerifyWithoutProcessors(
+                block,
                 config,
-                { ...options, filename: blockName, physicalFilename }
+                { ...options, filename: block.path, physicalFilename: block.physicalPath }
             );
         });
 
         return postprocess(messageLists, filenameToExpose);
```
mdjermanovic, line 1623:
```suggestion
        return processorService.postprocessSync(file, messageLists, {
            processor: {
                preprocess,
                postprocess
            }
        });
```

I'm guessing we'd like to use ProcessorService here as well.

lib/services/processor-service.js
```diff
@@ -0,0 +1,113 @@
...
+     */
+    postprocessSync(file, messages, config) {
+
+        const { processor } = config;
+
+        if (processor.postprocess) {
+            return processor.postprocess(messages, file.path);
+        }
+
+        return messages;
```
mdjermanovic, line 108:
```suggestion
        return processor.postprocess(messages, file.path);
```

I think it would be a bug if there's no `postprocess` method here, because `messages` must be flattened.

lib/linter/linter.js
```diff
@@ -1537,102 +1556,86 @@ class Linter {
...
-                    column: ex.column,
-                    nodeType: null
-                }
-            ];
-        }
-
-        const messageLists = blocks.map((block, i) => {
-            debug("A code block was found: %o", block.filename || "(unnamed)");
+        const messageLists = files.map(block => {
+            debug("A code block was found: %o", block.path || "(unnamed)");
```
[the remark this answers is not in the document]
nzakas: `block` can be a string, in which case `block.path` doesn't exist.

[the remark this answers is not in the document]
nzakas: I think it's better to leave the debug comment here. There are other debug comments in the loop that we'd want to have in order.

mdjermanovic (approved):
LGTM, thanks!

PR #18902, opened 2024-09-14
fix: add missing types for `no-extra-boolean-cast` options

kripod:
#### Prerequisites checklist

[ ] Documentation update
[ ] Add autofix to a rule
[ ] Add a CLI option
[ ] Add something to the core
[ ] Other, please explain:

- **Node version:** v22.8.0
- **npm version:** v10.8.2
- **Local ESLint version:** v9.10.0 (Currently used)
- **Operating System:** darwin 23.6.0

N/A

N/A

Tried to create a custom config using TypeScript as follows:

```ts
import type { Linter } from "eslint";
import type { ESLintRules } from "eslint/rules";

export default {
  rules: {
    "no-extra-boolean-cast": [
      "warn",
      { enforceForInnerExpressions: true }, // does not exist in type
    ],
  },
} satisfies Linter.Config<ESLintRules>;

```

Options existing at runtime should be available on the type level.

TypeScript flagged a type-level issue as highlighted in the comment above.

Fixed the types to include `enforceForInnerExpressions` and `enforceForLogicalOperands` options for the `no-extra-boolean-cast` rule.

N/A

### DIFF
--- a/lib/types/rules/possible-errors.d.ts
+++ b/lib/types/rules/possible-errors.d.ts
@@ -259,12 +259,6 @@ export interface PossibleErrors extends Linter.RulesRecord {
                  * @default false
                  */
                 enforceForInnerExpressions: boolean;
-                /**
-                 * @deprecated
-                 * @since 7.0.0-alpha.2
-                 * @default false
-                 */
-                enforceForLogicalOperands: boolean;
             }>,
         ]
     >;
--- a/lib/types/rules/possible-errors.d.ts
+++ b/lib/types/rules/possible-errors.d.ts
@@ -253,13 +253,26 @@ export interface PossibleErrors extends Linter.RulesRecord {
      */
     "no-extra-boolean-cast": Linter.RuleEntry<
         [
-            Partial<{
-                /**
-                 * @since 9.3.0
-                 * @default false
-                 */
-                enforceForInnerExpressions: boolean;
-            }>,
+            | Partial<{
+                  /**
+                   * @since 9.3.0
+                   * @default false
+                   */
+                  enforceForInnerExpressions: boolean;
+                  /**
+                   * @deprecated
+                   */
+                  enforceForLogicalOperands: never;
+              }>
+            | Partial<{
+                  /**
+                   * @deprecated
+                   * @since 7.0.0-alpha.2
+                   * @default false
+                   */
+                  enforceForLogicalOperands: boolean;
+                  enforceForInnerExpressions: never;
+              }>,
         ]
     >;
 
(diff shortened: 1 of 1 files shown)


amareshsm (dismissed):
LGTM. 
`no-extra-boolean-cast` option - https://eslint.org/docs/latest/rules/no-extra-boolean-cast#options

Before marking this as an accepted pull request, I would like the ESLint team to review it once.

lib/types/rules/possible-errors.d.ts
```diff
@@ -251,7 +251,23 @@ export interface PossibleErrors extends Linter.RulesRecord {
...
+                 * @since 9.3.0
+                 * @default false
+                 */
+                enforceForInnerExpressions: boolean;
+                /**
+                 * @deprecated
+                 * @since 7.0.0-alpha.2
+                 * @default false
+                 */
+                enforceForLogicalOperands: boolean;
```
snitin315, line 267: Only one of the options can be present in the object.

https://github.com/kirkwaiblinger/eslint/blob/a7016a50d88011f279d52b9355a5662e561c414c/lib/rules/no-extra-boolean-cast.js#L33-L54

kripod: Indeed, I’ve just removed the deprecated option accordingly as the bundled types are a fairly new addition anyway.

Thanks for pointing this out 🙌

lib/types/rules/possible-errors.d.ts
```diff
@@ -251,7 +251,17 @@ export interface PossibleErrors extends Linter.RulesRecord {
...
-    "no-extra-boolean-cast": Linter.RuleEntry<[]>;
+    "no-extra-boolean-cast": Linter.RuleEntry<
+        [
+            Partial<{
+                /**
+                 * @since 9.3.0
+                 * @default false
+                 */
+                enforceForInnerExpressions: boolean;
+            }>,
```
snitin315, line 262: Let's include `enforceForLogicalOperands ` too because it does exist. Let's have something like below.
```js
           Partial<{
                enforceForInnerExpressions: boolean;
            }>
            
            |
        
            Partial<{
                /**
                 * @deprecated Use `enforceForInnerExpressions` instead
                 */
                enforceForLogicalOperands: boolean;
            }>
 ```

kripod: Thanks for the tip. Unfortunately that doesn’t make the two options mutually exclusive.

However, the `never` type can be leveraged for such functionality as seen in my most recent push.

snitin315 (approved):
LGTM, thank you!

PR #18914, opened 2024-09-17
fix: add missing types for `no-restricted-exports` rule

kripod:
#### Prerequisites checklist

[ ] Documentation update
[ ] Add autofix to a rule
[ ] Add a CLI option
[ ] Add something to the core
[ ] Other, please explain:

- **Node version:** v22.8.0
- **npm version:** v10.8.2
- **Local ESLint version:** v9.10.0 (Currently used)
- **Operating System:** darwin 23.6.0

N/A

N/A

```ts
import type { Linter } from "eslint";
import type { ESLintRules } from "eslint/rules";

export default {
  rules: {
    "no-restricted-exports": [
      "error",
      {
        restrictedNamedExports: [
          // Ambiguous with dynamic imports
          // https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/import#module_namespace_object
          "then",
        ],
        restrictDefaultExports: {
          // Prefer `export default` expressions
          named: true,

          // Avoid re-exporting entire modules as is
          namespaceFrom: true,
        },
      },
    ],
  },
} satisfies Linter.Config<ESLintRules>;
```

Autocompletion and type checking should be in place for the `no-restricted-exports` rule.

No assistance was given by TypeScript.

Added type declarations for the `no-restricted-exports` rule.

N/A

### DIFF
--- a/lib/types/rules/ecmascript-6.d.ts
+++ b/lib/types/rules/ecmascript-6.d.ts
@@ -232,6 +232,52 @@ export interface ECMAScript6 extends Linter.RulesRecord {
     "no-new-symbol": Linter.RuleEntry<[]>;
 
     /**
+     * Rule to disallow specified names in exports.
+     *
+     * @since 7.0.0-alpha.0
+     * @see https://eslint.org/docs/rules/no-restricted-exports
+     */
+    "no-restricted-exports": Linter.RuleEntry<
+        [
+            Partial<{
+                /**
+                 * @default []
+                 */
+                restrictedNamedExports: string[];
+                /**
+                 * @since 9.3.0
+                 */
+                restrictedNamedExportsPattern: string;
+                /**
+                 * @since 8.33.0
+                 */
+                restrictDefaultExports: Partial<{
+                    /**
+                     * @default false
+                     */
+                    direct: boolean;
+                    /**
+                     * @default false
+                     */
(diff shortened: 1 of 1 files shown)


lib/types/rules/ecmascript-6.d.ts
```diff
@@ -231,6 +231,52 @@ export interface ECMAScript6 extends Linter.RulesRecord {
...
+                 */
+                restrictedNamedExports: string[];
+                /**
+                 * @since 9.3.0
+                 */
+                restrictedNamedExportsPattern: string;
+                /**
+                 * @since 8.33.0
+                 */
+                restrictDefaultExports: Partial<{
```
snitin315, line 254: The option works only if the r`estrictedNamedExports` option does not contain the `"default"` value. Is it possible to have that check via typescript too?

JoshuaKGoldberg: Not really 😕. There are no "negated" types in TypeScript (https://github.com/microsoft/TypeScript/issues/4196), so you can't say _"this type is all strings **except** some literal"_.

snitin315: Hmm, it should be fine then.

snitin315 (approved):
LGTM, thank you.

Issue #18918, opened 2024-09-18
Docs: [no-await-in-loop] mention preventing unhandled rejections as an additional benefit of the rule


kirkwaiblinger:
### Docs page(s)

https://eslint.org/docs/latest/rules/no-await-in-loop

### What documentation issue do you want to solve?

The docs currently understate the value of the rule ❤ by not mentioning promise rejections

### What do you think is the correct solution?

Mention promise rejections

### Participation

- [x] I am willing to submit a pull request for this change.

### Additional comments

Specifically, it's usually considered bad to write code like this...
```js
async function asyncFunctionThatMightThrow() {
  const promise1 = Promise.reject(new Error('rejected 1'));
  const promise2 = Promise.reject(new Error('rejected 2'));
 
  // bad
  const value1 = await promise1;
  const value2 = await promise2;

  return [value1, value2];
}
```
...because even if you're aware of exceptions/rejections when calling it...
```js
async function foo() {
    try {
      return await asyncFunctionThatMightThrow();
    } catch (e) {
        // gotcha!
       return "don't worry, situation is under control!";
    }
}
```
...you still are punished with floating unhandled rejections:
```js

foo(); // returns "don't worry, situation is under control!";

// then, node process crashes, Uncaught Error: rejected 2
```

But, with the promise concurrency methods, all is well
```js
async function asyncFunctionThatMightThrow() {
  const promise1 = Promise.reject(new Error('rejected 1'));
  const promise2 = Promise.reject(new Error('rejected 2'));
 
  // good
  const [value1, value2] = await Promise.all([promise1, promise2])

  return [value1, value2];
}

// same foo
foo(); // returns "don't worry, situation is under control!"

// no crash!
```

Now, if I rewrite this in a way that handles arbitrarily many promises...
```js
async awaitInLoop(n) {
    const arrayOfRejections = new Array(n).fill(undefined).map((_, i) => Promise.reject(new Error("rejected " + i)));
    const returnArray = [];
    for (const promise of arrayOfRejections) {
          const promisedValue = await promise; // flagged by no-await-in-loop 😄 
          returnArray.push(promisedValue);
    }
    return returnArray;
}
```
...and repeat the exercise from before, I'll get floating unhandled rejections. Instead, using a promise concurrency method, ie the following....
```js
async awaitInLoop(n) {
    const arrayOfRejections = new Array(n).fill(undefined).map((_, i) => Promise.reject(new Error("rejected " + i)));
    const returnArray = [];
[...]

PR #18942, opened 2024-09-22
docs: add info about unhandled promise rejections.

Tanujkanti4441:
#### Prerequisites checklist

[x] Documentation update
[ ] Add autofix to a rule
[ ] Add a CLI option
[ ] Add something to the core
[ ] Other, please explain:

Added a section about unhandled promise rejections in `no-await-in-loops` rule docs.
Fixes: #18918


docs/src/rules/no-await-in-loop.md
```diff
@@ -37,6 +37,37 @@ async function foo(things) {
 }
 ```
 
+Also, if a loop iterates over `Promise` rejections with `await`, then after the first promise is rejected, it immediately throws an error, stopping the loop's execution before handling the rest of the promises. This leads to unhandled rejections for any promises that haven’t been awaited yet.
```
amareshsm, line 40:
```suggestion
Also, suppose a loop iterates over `Promise` rejections with `await` in that case, after the first promise is rejected, it immediately throws an error, stopping the loop's execution before handling the rest of the promises. This leads to unhandled rejections for promises that haven’t been awaited yet.
```

docs/src/rules/no-await-in-loop.md
```diff
@@ -37,6 +37,37 @@ async function foo(things) {
...
+}
+```
+
+This function should be refactored as shown:
+
+```js
+async function bar(things) {
+  const arrayOfRejections = new Array(things).fill(undefined).map((_, index) => Promise.reject(new Error("rejected " + index)));
+  const result = [];
+  // Good: all the promises are immediately rejected so there are no floating unhandled rejections.
```
amareshsm, line 62:
```suggestion
  // Good: if any of the promises fail, all others are rejected immediately, so there are no floating unhandled rejections.

```

kirkwaiblinger: > // Good: if any of the promises fail, all others are rejected immediately, so there are no floating unhandled rejections.

This suggestion isn't quite right. The other promises' resolutions aren't changed in any way. They just don't cause unhandled rejections https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/all

```suggestion
  // Good: if any of the promises rejects, the Promise.all() call will reject, and any rejections in the other promises will not be treated as unhandled rejections
```

docs/src/rules/no-await-in-loop.md
kirkwaiblinger, line 40: Note that this only applies to iterating a collection of _already existing_ promises. If you have a lazy, possibly-infinite (sync) generator of promises, then `await`ing in the loop may well be correct.

```ts
function * yieldPromises() {
   let n = 1;
   try {
    while (true) {
        if (n > 10) {
          yield Promise.reject(new Error(`Promise ${n}`));
        } else {
          yield Promise.resolve(`Promise ${n}`);
        }
        n++;
    }
   } finally {
    console.log('promise generator closed');
   }
}

async function awaitUntilRejection() {
   for (const promise of yieldPromises()) {
      try {
        const fulfillmentValue = await promise;
        console.log(`Fulfillment value: ${fulfillmentValue}`);
      } catch (e) {
        console.log(`first rejection message: ${e}` )
        break;
      }
   }
}

// no unhandled rejections, since no other promises were even created.
awaitUntilRejection();
```

docs/src/rules/no-await-in-loop.md
Tanujkanti4441: Updated!

docs/src/rules/no-await-in-loop.md
Tanujkanti4441: Even in this case, rule still reports `await`, so added this info under `when not to use section` with these examples.
Please share if it seems correct!

kirkwaiblinger: Hmm, so - the info is technically correct, but the example has been moved into a different context, where promise rejections aren't really germane. All that really matters there is: you obviously need to iterate using `await` inside loops sometimes for control flow reasons, since you can't eagerly `Promise.all()` an infinite collection.

I guess I would say, my examples in the original issue and in this PR have been intended to communicate with the ESlint members, not meant as polished user-facing communication, so no need to use any of them in the resulting docs if they don't fit the flow 🙂 

FYI - I don't want to overstep, but there are also quite a few grammar/sentence flow issues that I would request be addressed if I were reviewing a PR for a project of which I am a member.

nzakas: @kirkwaiblinger would you like to submit a PR that updates the docs in the way you think is appropriate?

kirkwaiblinger: I would be happy to.... But I am a little swamped right now, so I may not be able to get to it until next week. If I have time, I'll give it a shot, but no worries either if things move forward in the meantime.

nzakas: We're in no hurry. It seems like you'd be the best person to update the docs so we can wait.

Tanujkanti4441: I feel the same way, that would be even great.

Tanujkanti4441:
Based on conversations above, i think we can close the PR.

PR #19010, opened 2024-10-11
feat: add types for the `@eslint/js` package

snitin315:
#### Prerequisites checklist

[ ] Documentation update
[ ] Add autofix to a rule
[ ] Add a CLI option
[ ] Add something to the core
[x] Other, please explain: Add types

Fix https://github.com/eslint/eslint/issues/18998

I have defined `eslint` as a dependency because of it's usage in types similar to https://github.com/eslint/eslint/pull/18938

### DIFF
--- a/eslint.config.js
+++ b/eslint.config.js
@@ -292,11 +292,11 @@ module.exports = [
     })),
     {
         name: "eslint/ts-rules",
-        files: ["tests/lib/types/*.ts"],
+        files: ["tests/lib/types/*.ts", "packages/**/**/*.{ts,mts,cts,tsx}"],
         languageOptions: {
             parser: tsParser,
             parserOptions: {
-                project: "tests/lib/types/tsconfig.json"
+                project: ["tests/lib/types/tsconfig.json", "packages/js/tests/types/tsconfig.json"]
             }
         },
         plugins: {
--- a/eslint.config.js
+++ b/eslint.config.js
@@ -292,7 +292,7 @@ module.exports = [
     })),
     {
         name: "eslint/ts-rules",
-        files: ["tests/lib/types/*.ts", "packages/**/**/*.{ts,mts,cts,tsx}"],
+        files: ["tests/lib/types/*.ts", "packages/**/*.{ts,mts,cts,tsx}"],
         languageOptions: {
             parser: tsParser,
             parserOptions: {
--- a/packages/js/package.json
+++ b/packages/js/package.json
@@ -11,6 +11,9 @@
     "src",
     "types"
   ],
+  "dependencies": {
+    "eslint": "^9.12.0"
+  },
   "publishConfig": {
     "access": "public"
   },
--- a/packages/js/package.json
+++ b/packages/js/package.json
@@ -4,7 +4,9 @@
   "description": "ESLint JavaScript language implementation",
   "main": "./src/index.js",
   "types": "./types/index.d.ts",
-  "scripts": {},
+  "scripts": {
+    "test:types": "tsc -p tests/types/tsconfig.json"
+  },
   "files": [
     "LICENSE",
     "README.md",
--- a/packages/js/package.json
+++ b/packages/js/package.json
@@ -3,11 +3,13 @@
   "version": "9.13.0",
   "description": "ESLint JavaScript language implementation",
   "main": "./src/index.js",
+  "types": "./types/index.d.ts",
   "scripts": {},
   "files": [
     "LICENSE",
     "README.md",
-    "src"
+    "src",
+    "types"
   ],
   "publishConfig": {
     "access": "public"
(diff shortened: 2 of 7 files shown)


snitin315:
I'm not sure how to fix the trunk check. Looking into it.

packages/js/package.json
```diff
@@ -3,12 +3,19 @@
...
   "files": [
     "LICENSE",
     "README.md",
-    "src"
+    "src",
+    "types"
   ],
+  "dependencies": {
+    "eslint": "^9.12.0"
+  },
```
mdjermanovic, line 18: Is it necessary to declare `eslint` as a dependency? For example, `@eslint/markdown` uses ESLint types but doesn't have `eslint` in `dependencies`.

snitin315: What if someone uses the latest `@eslint/js` version but an older version of eslint that doesn't have the required types? If not a dependency, at least we should have it as a peer dependency (maybe we can mark it optional too).

mdjermanovic: I'd rather avoid introducing a circular dependency (`@eslint/js` is a dependency of `eslint`) especially since it's only needed for types, but I'm not sure what would the correct approach here. @eslint/eslint-team thoughts?

snitin315: I'm not sure about the correct approach here. Open to suggestions. Is there any case where @eslint/js can be used independently without eslint?

aryaemami59: I believe `eslint` should be an optional peer dependency of `@eslint/js` since `@eslint/js` is meant to be used with `eslint` but should not provide its own version of `eslint`. So in this case we should have:

```suggestion
  "devDependencies": {
    "eslint": "^9.13.0"
  },
  "peerDependencies": {
    "eslint": "^9.13.0"
  },
  "peerDependenciesMeta": {
    "eslint": {
      "optional": true
    }
  },
```

packages/js/tests/types/types.test.ts
```diff
@@ -0,0 +1,72 @@
...
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE
+ */
+
+import type { Linter } from "eslint";
+import js from "../../";
```
aryaemami59, line 29:
```suggestion
import js = require("@eslint/js");
```

packages/js/package.json
nzakas: i think it's fine to not have `eslint` specified as a dependency.

nzakas:
Please check the lint error.

packages/js/tests/types/types.test.ts
snitin315: This throws an error because it resolves to the published version inside `node_module` where types don't exist.

Screenshot 2024-10-26 at 6 12 57 PM

aryaemami59: @snitin315 That's weird, that should not be the case, what do you see when you hover over `"@eslint/js"` in `import js = require("@eslint/js")`?

snitin315: Same error on hovering too.

Screenshot 2024-10-26 at 6 25 50 PM
Even in DefinitelyTyped repo.


Screenshot 2024-10-26 at 6 25 00 PM

snitin315: It should be fine as long as we are pointing to the correct path.

aryaemami59: I guess then it that case we should have:

```ts
import js = require("../../types/index.js");
```

snitin315: >import js = require("../../types/index.js");

I don't think it should be types because we have `...js.configs.recommended` usage below.

>"@eslint/js": "file:."

This is the same as `import js from "../../";`.

nzakas (dismissed):

eslint.config.js
```diff
@@ -292,11 +292,11 @@ module.exports = [
     })),
     {
         name: "eslint/ts-rules",
-        files: ["tests/lib/types/*.ts"],
+        files: ["tests/lib/types/*.ts", "packages/**/**/*.{ts,mts,cts,tsx}"],
```
mdjermanovic, line 295:
```suggestion
        files: ["tests/lib/types/*.ts", "packages/**/*.{ts,mts,cts,tsx}"],
```

Consecutive `/**` seem unnecessary?

PR #19050, opened 2024-10-23
chore: add type definitions for the `eslint-config-eslint` package

### LINKED ISSUE #19043: Change Request: Add type definitions for `eslint-config-eslint`
### ESLint version

v9

Add type definitions for `eslint-config-eslint`.

Manually add type definitions for `eslint-config-eslint`.

### Participation

### Additional comments

_No response_

aryaemami59:
#### Prerequisites checklist

[ ] Documentation update
[ ] Add autofix to a rule
[ ] Add a CLI option
[ ] Add something to the core
[X] Other, please explain:

## **This PR**:

- [X] Adds type definitions for the `eslint-config-eslint` package.
- [X] Resolves #19043.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -114,8 +114,24 @@ jobs:
             wdio-logs/*.log
 
   test_types:
-    name: Types Test
+    name: Test Types of ${{ matrix.package.name }}
     runs-on: ubuntu-latest
+
+    strategy:
+      matrix:
+        package:
+          [
+            { name: eslint, directory: . },
+            {
+              name: eslint-config-eslint,
+              directory: packages/eslint-config-eslint,
+            },
+            {
+              name: '@eslint/js',
+              directory: packages/js,
+            },
+          ]
+
     steps:
       - uses: actions/checkout@v4
       - uses: actions/setup-node@v4
@@ -123,8 +139,11 @@ jobs:
           node-version: "lts/*"
       - name: Install Packages
         run: npm install
-      - name: Test eslint types
+
+      - name: Install Packages for ${{ matrix.package.name }}
+        working-directory: ${{ matrix.package.directory }}
+        run: npm install
+
+      - name: Test types for ${{ matrix.package.name }}
+        working-directory: ${{ matrix.package.directory }}
         run: npm run test:types
-      - name: Test @eslint/js types
-        run: npm run test:types
-        working-directory: packages/js
--- a/eslint.config.js
+++ b/eslint.config.js
@@ -70,6 +70,9 @@ function createInternalFilesPatterns(pattern = null) {
         }));
 }
 
+/**
+ * @type {import("./lib/types/index.js").Linter.Config[]}
+ */
 module.exports = [
     ...eslintConfigESLintCJS.map(config => ({
         ...config,
@@ -296,7 +299,7 @@ module.exports = [
         languageOptions: {
             parser: tsParser,
             parserOptions: {
-                project: ["tests/lib/types/tsconfig.json", "packages/js/tests/types/tsconfig.json"]
+                project: ["tests/lib/types/tsconfig.json", "packages/js/tests/types/tsconfig.json", "packages/eslint-config-eslint/tsconfig.json"]
             }
         },
         plugins: {
--- a/packages/eslint-config-eslint/formatting.js
+++ b/packages/eslint-config-eslint/formatting.js
@@ -1,123 +1,138 @@
 "use strict";
 
+/**
+ * @type {import("eslint").Linter.RulesRecord}
+ */
+const rules = {
+    "array-bracket-spacing": "error",
+    "arrow-parens": ["error", "as-needed"],
+    "arrow-spacing": "error",
+    "block-spacing": "error",
+    "brace-style": ["error", "1tbs"],
+    "comma-dangle": "error",
+    "comma-spacing": "error",
+    "comma-style": ["error", "last"],
+    "computed-property-spacing": "error",
+    "dot-location": ["error", "property"],
+    "eol-last": "error",
+    "func-call-spacing": "error",
+    "function-call-argument-newline": ["error", "consistent"],
+    "function-paren-newline": ["error", "consistent"],
+    "generator-star-spacing": "error",
+    indent: ["error", 4, { SwitchCase: 1 }],
+    "key-spacing": ["error", { beforeColon: false, afterColon: true }],
+    "keyword-spacing": "error",
+    "lines-around-comment": [
+        "error",
+        {
+            beforeBlockComment: true,
+            afterBlockComment: false,
+            beforeLineComment: true,
+            afterLineComment: false
+        }
+    ],
+    "max-len": [
+        "error",
+        160,
+        {
+            ignoreComments: true,
+            ignoreUrls: true,
+            ignoreStrings: true,
+            ignoreTemplateLiterals: true,
+            ignoreRegExpLiterals: true
+        }
+    ],
+    "max-statements-per-line": "error",
+    "new-parens": "error",
+    "no-confusing-arrow": "error",
+    "no-extra-semi": "error",
+    "no-floating-decimal": "error",
+    "no-mixed-spaces-and-tabs": ["error", false],
+    "no-multi-spaces": "error",
+    "no-multiple-empty-lines": [
+        "error",
+        {
+            max: 2,
+            maxBOF: 0,
+            maxEOF: 0
+        }
+    ],
+    "no-tabs": "error",
+    "no-trailing-spaces": "error",
+    "no-whitespace-before-property": "error",
+    "object-curly-newline": [
+        "error",
+        {
+            consistent: true,
+            multiline: true
+        }
+    ],
+    "object-curly-spacing": ["error", "always"],
+    "object-property-newline": [
+        "error",
+        {
+            allowAllPropertiesOnSameLine: true
+        }
+    ],
+    "one-var-declaration-per-line": "error",
+    "operator-linebreak": "error",
+    "padding-line-between-statements": [
+        "error",
+        {
+            blankLine: "always",
+            prev: ["const", "let", "var"],
+            next: "*"
+        },
+        {
+            blankLine: "any",
+            prev: ["const", "let", "var"],
+            next: ["const", "let", "var"]
+        }
+    ],
+    quotes: ["error", "double", { avoidEscape: true }],
+    "quote-props": ["error", "as-needed"],
+    "rest-spread-spacing": "error",
+    semi: "error",
+    "semi-spacing": [
+        "error",
+        {
+            before: false,
+            after: true
+        }
+    ],
+    "semi-style": "error",
+    "space-before-blocks": "error",
+    "space-before-function-paren": [
+        "error",
+        {
+            anonymous: "never",
+            named: "never",
+            asyncArrow: "always"
+        }
+    ],
+    "space-in-parens": "error",
+    "space-infix-ops": "error",
+    "space-unary-ops": [
+        "error",
+        {
+            words: true,
+            nonwords: false
+        }
+    ],
+    "spaced-comment": [
+        "error",
+        "always",
+        {
+            exceptions: ["-"]
+        }
+    ],
+    "switch-colon-spacing": "error",
+    "template-curly-spacing": ["error", "never"],
+    "template-tag-spacing": "error",
+    "wrap-iife": "error",
+    "yield-star-spacing": "error"
+};
+
 module.exports = {
-    rules: {
-        "array-bracket-spacing": "error",
-        "arrow-parens": ["error", "as-needed"],
-        "arrow-spacing": "error",
-        "block-spacing": "error",
-        "brace-style": ["error", "1tbs"],
-        "comma-dangle": "error",
-        "comma-spacing": "error",
-        "comma-style": ["error", "last"],
-        "computed-property-spacing": "error",
-        "dot-location": ["error", "property"],
-        "eol-last": "error",
-        "func-call-spacing": "error",
-        "function-call-argument-newline": ["error", "consistent"],
-        "function-paren-newline": ["error", "consistent"],
-        "generator-star-spacing": "error",
-        indent: ["error", 4, { SwitchCase: 1 }],
-        "key-spacing": ["error", { beforeColon: false, afterColon: true }],
-        "keyword-spacing": "error",
-        "lines-around-comment": ["error",
-            {
-                beforeBlockComment: true,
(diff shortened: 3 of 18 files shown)


.github/workflows/ci.yml
```diff
@@ -116,12 +116,24 @@ jobs:
   test_types:
     name: Types Test
     runs-on: ubuntu-latest
+
+    strategy:
+      matrix:
+        package:
+          [{ name: eslint-config-eslint, directory: packages/eslint-config-eslint }]
```
nzakas, line 123: I might be missing something, but with this change, won't we end up skipping the types test in the project root? Don't we need another entry in the array?

eslint.config.js
```diff
@@ -70,6 +70,9 @@ function createInternalFilesPatterns(pattern = null) {
         }));
 }
 
+/**
+ * @type {import("eslint").Linter.Config[]}
```
nzakas, line 74: Should import from `lib/types` instead?

packages/eslint-config-eslint/formatting.js
```diff
@@ -1,123 +1,135 @@
 "use strict";
 
+const rules = {
```
nzakas, line 3: What's the purpose of this change?

nzakas (commented):
Overall LGTM. Just a few questions.

eslint.config.js
aryaemami59: It resolves to `lib/types/index` so they're the same thing but I'll change it to be more explicit.

packages/eslint-config-eslint/formatting.js
aryaemami59: It's so that we can use named imports in ESM files:

```ts
import { rules } from "eslint-config-eslint/formatting";

export default [...eslintConfigESLint, rules] satisfies Linter.Config[];
```

.github/workflows/ci.yml
aryaemami59: You're right, I fixed it.

packages/eslint-config-eslint/nodejs.js
```diff
@@ -1,14 +1,24 @@
 "use strict";
 
-const recommendedScriptConfig = require("eslint-plugin-n/configs/recommended-script");
-const recommendedModuleConfig = require("eslint-plugin-n/configs/recommended-module");
+const {
+    configs: {
+        "flat/recommended-script": recommendedScriptConfig,
+        "flat/recommended-module": recommendedModuleConfig
+    }
+} = require("eslint-plugin-n");
```
aryaemami59, line 8: [**"eslint-plugin-n/configs/recommended-script"**](https://github.com/eslint-community/eslint-plugin-n/blob/master/configs/recommended-script.js#L4) and [**"eslint-plugin-n/configs/recommended-module"**](https://github.com/eslint-community/eslint-plugin-n/blob/master/configs/recommended-script.js#L4) are deprecated.

packages/eslint-config-eslint/package.json
```diff
@@ -21,7 +50,8 @@
     "cjs.js",
     "formatting.js",
     "index.js",
-    "nodejs.js"
+    "nodejs.js",
+    "types"
```
snitin315, line 54: We don't need to include `ambient-module-declarations`?

snitin315 (commented):
Overall looks good. Few comments. Can you look into the failing CI as well?

packages/eslint-config-eslint/ambient-module-declarations/@eslint-community-eslint-plugin-eslint-comments.d.ts
```diff
@@ -0,0 +1,23 @@
+declare module "@eslint-community/eslint-plugin-eslint-comments" {
```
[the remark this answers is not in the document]
aryaemami59: No because We don't want them to be included in the final package tarball.

packages/eslint-config-eslint/ambient-module-declarations/eslint-plugin-jsdoc.d.ts
```diff
@@ -0,0 +1,23 @@
+declare module "eslint-plugin-jsdoc" {
+  import type { ESLint, Linter } from "eslint";
+
+  const eslintPluginJsdoc: ESLint.Plugin & {
+      configs: Record<
+          | "recommended"
```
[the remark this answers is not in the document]
aryaemami59: No we don't have to worry about these as they are just added for intellisense for us, but I will open issues in `eslint-plugin-jsdoc` and `eslint-plugin-unicorn` regardless.

packages/eslint-config-eslint/package.json
aryaemami59: No. The `ambient-module-declarations` is just for us.

packages/eslint-config-eslint/formatting.js
aryaemami59: In an ESM file we will be able to do:

```ts
import { rules } from "eslint-config-eslint/formatting";

export default [...eslintConfigESLint, rules] satisfies Linter.Config[];
```

Without the provided changes, this would fail.

.github/workflows/ci.yml
```diff
@@ -57,6 +57,8 @@ jobs:
         uses: trunk-io/trunk-action@v1
         with:
 
+          save-annotations: false
+
```
aryaemami59, line 61: Not sure why but I had to disable `save-annotations` to make CI happy.

snitin315: What was the error with  `save-annotations` enabled?

aryaemami59: https://github.com/eslint/eslint/actions/runs/11533071521/job/32105566286

nzakas: That looks like an error in Trunk that we should get checked out.

@det

fasttime: This is the same problem we had in the main branch before pinning `trunk-action` to v1.1.17: https://github.com/trunk-io/trunk-action/pull/266.

JoshuaKGoldberg (approved):
LGTM! Just one suggestion about redundant properties, but nothing blocking IMO.

packages/eslint-config-eslint/tsconfig.json
```diff
@@ -0,0 +1,20 @@
+{
+    "compilerOptions": {
+        "allowSyntheticDefaultImports": false,
+        "esModuleInterop": false,
```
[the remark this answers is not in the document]
aryaemami59: Actually [**`esModuleInterop`**](https://www.typescriptlang.org/tsconfig/#esModuleInterop) is automatically set to `true` if [**`module`**](https://www.typescriptlang.org/tsconfig/#module) is set to **`Node16`** or **`NodeNext`**. But this makes me wonder if explicitly disable [**`esModuleInterop`**](https://www.typescriptlang.org/tsconfig/#esModuleInterop), would it still techinically be set to `true` since we have [**`module`**](https://www.typescriptlang.org/tsconfig/#module) set to **`NodeNext`**?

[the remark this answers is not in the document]
JoshuaKGoldberg: Since it's just being used for type-checking and not transpilation, I'd say as long as there are no type errors it's 👍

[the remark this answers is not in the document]
aryaemami59: Awesome, so you still think I should remove them?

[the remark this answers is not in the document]
JoshuaKGoldberg: I do, yup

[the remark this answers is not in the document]
aryaemami59: I tried it, removing them makes the type tests fail.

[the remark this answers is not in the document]
JoshuaKGoldberg: In that case, I suppose I'm wrong to suggest removing it 😄

packages/eslint-config-eslint/formatting.js
aladdin-add: is it supported and documented somewhere? As far as I can remember, cjs should be treated as a default export.

aladdin-add: it's very helpful, thanks for the explanation!

I agreed with @mdjermanovic said in https://github.com/eslint/eslint/issues/19110#issuecomment-2460723401.
it's explicitly stated the default export usage, and it's the officially recommended; we do not guarantee that non-recommended usage will work.

aladdin-add (approved):

packages/eslint-config-eslint/package.json
```diff
@@ -36,6 +66,20 @@
     "eslint-plugin-n": "^17.11.1",
     "eslint-plugin-unicorn": "^52.0.0"
   },
+  "devDependencies": {
+    "@arethetypeswrong/cli": "^0.16.4",
+    "eslint": "^9.14.0",
+    "eslint-config-eslint": "file:",
```
fasttime, line 72: As explained in https://nodejs.org/api/packages.html#self-referencing-a-package-using-its-name, this dependency is unnecessary: `eslint-config-eslint` will be automatically resolved to the current package because it matches the `"name"` property in `package.json` and an `"exports"` property is present.

packages/eslint-config-eslint/ambient-module-declarations/@eslint-community-eslint-plugin-eslint-comments.d.ts
fasttime, line 1: What is the purpose of these ambient declarations? VSCode seems to ignore them completely. If I delete the `ambient-module-declarations` directory I can see the types inferred correctly in the IDE and all scripts in `package.json` are still passing:

image

fasttime (approved):
LGTM, thanks! Since this is a `feat:` pull request we'll have to wait until #19096 is closed before merging.

packages/eslint-config-eslint/formatting.js
mdjermanovic, line 1: If the changes in this file are made only to allow for `import { rules } from "eslint-config-eslint/formatting";`, then I think we should revert them, because `import { rules } from "eslint-config-eslint/formatting";` is not supposed to work. This module exports a config object with a `rules` property, not named export `rules`.

packages/eslint-config-eslint/types/formatting.d.ts
```diff
@@ -0,0 +1,3 @@
+import type { Linter } from "eslint";
+
+export declare const rules: NonNullable<Linter.Config["rules"]>;
```
mdjermanovic, line 3: I think this doesn't reflect the intended usage, which is `import eslintConfigESLintFormatting from "eslint-config-eslint/formatting";`:

https://github.com/eslint/eslint/tree/main/packages/eslint-config-eslint#formatting-config

 This module exports a `Linter.Config` object, not named export `rules`.

packages/eslint-config-eslint/tests/types/types.test-d.mts
```diff
@@ -0,0 +1,20 @@
+import type { Linter } from "eslint";
+
+import * as eslintConfigESLint from "eslint-config-eslint";
+import * as eslintConfigESLintBase from "eslint-config-eslint/base";
+import * as eslintConfigESLintCJS from "eslint-config-eslint/cjs";
+import * as eslintConfigESLintFormatting from "eslint-config-eslint/formatting";
```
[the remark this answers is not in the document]
fasttime: > I'm not sure how these tests are passing because, for example, `eslintConfigESLint` evaluates to `{ default: [...] }`.

I think that's because the [`allowSyntheticDefaultImports`](https://www.typescriptlang.org/tsconfig/#allowSyntheticDefaultImports) is set to `false` in the TypeScript config. We could change that option to `true` to disallow the `import *` syntax in the types tests if it's confusing.

mdjermanovic:
I changed the tag to `chore` because all commits in this repo appear in changelogs for the `eslint` package but this isn't a new feature in `eslint`.

Regardless, when releasing `eslint-config-eslint` with this change, I think we should choose __major__ as RELEASE_TYPE, to bump the version to 12.0.0, because adding `peerDependencies` is a breaking change.

fasttime:
If we are going to release this change in a new major version I think we could as well set the `eslint` version in `peerDependencies` to `"^9.10.0"`, since v9.10.0 is the first ESLint version that includes types.

packages/eslint-config-eslint/tests/types/types.test-d.mts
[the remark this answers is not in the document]
mdjermanovic: > I think that's because the [`allowSyntheticDefaultImports`](https://www.typescriptlang.org/tsconfig/#allowSyntheticDefaultImports) is set to `false` in the TypeScript config. We could change that option to `true` to disallow the `import *` syntax in the types tests if it's confusing.

Sounds good to me 👍

nzakas:
Is this ready for another review?

fasttime:
It looks like https://github.com/eslint/eslint/pull/19050#discussion_r1845564637 hasn't been addressed yet.

aryaemami59:
@fasttime Done.

@nzakas Yes.

mdjermanovic (dismissed):
LGTM, thanks! Leaving open for @fasttime to re-review.

packages/eslint-config-eslint/tsconfig.json
```diff
@@ -0,0 +1,19 @@
+{
+    "compilerOptions": {
+        "allowSyntheticDefaultImports": false,
```
fasttime, line 3:
```suggestion
        "allowSyntheticDefaultImports": true,
```

I think the idea was to set `allowSyntheticDefaultImports` to `true` to enforce the syntax `import eslintConfigESLint from "eslint-config-eslint";` over `import * as eslintConfigESLint from "eslint-config-eslint";` in our tests.

`allowSyntheticDefaultImports` makes TypeScript behave more similarly to Node.js when it imports a CommonJS file, and that's what the team is more familiar with.

packages/eslint-config-eslint/tests/types/types.test-d.mts
fasttime, line 1: Just asking: is this file named with `-d`  because it tests declaration files (`.d.ts`)? It took me a second look to realize that the file extension is just `.mts`.

packages/eslint-config-eslint/tests/types/types.test-d.ts
fasttime, line 1: Same question as above about `-d` in the file name.

aryaemami59: See [**this comment**](https://github.com/eslint/eslint/pull/19050#discussion_r1873812918)

fasttime (approved):
LGTM, thanks!

PR #19060, opened 2024-10-27
fix: update types for `no-restricted-imports` rule

snitin315:
#### Prerequisites checklist

[x] Documentation update
[ ] Add autofix to a rule
[ ] Add a CLI option
[ ] Add something to the core
[ ] Other, please explain:

Update types for the `no-restricted-imports` rule.

NA


lib/types/rules/ecmascript-6.d.ts
```diff
@@ -291,23 +313,39 @@ export interface ECMAScript6 extends Linter.RulesRecord {
...
+                                      allowImportNames?: string[];
+                                      importNames?: never;
+                                  }
+                            ))
+                      >;
+                      patterns: Array<
+                          string | ValidNoRestrictedImportPatternOptions
+                      >;
+                  }>
+            >
```
JoshuaKGoldberg, line 348: [Style] This is getting to be pretty deep & complex 😅. Suggestion: extract out the reused portions into named interfaces, such as the two `{ importNames?: ...; allowImportNames?: ... }` shapes?

For example, the `ValidNoRestrictedImportPatternOptions` type could look something like:

```ts
type ValidNoRestrictedImportPatternOptions =
  NoRestrictedImportPatternCommonOptions &
    EitherGroupOrRegEx &
    EitherNameSpecifiers;
```

It might also make sense to write a utility type to generate the _"union of objects, where each only allows one property"_ types. [`type-fest`](https://github.com/sindresorhus/type-fest)'s [`MergeExclusive`](https://github.com/sindresorhus/type-fest/blob/3d473d1f8b0ec5820e1f3aaf7965b68784423778/source/merge-exclusive.d.ts) does that for just two. [Here's a playground for a type that allows arbitrary objects](https://www.typescriptlang.org/play/?exactOptionalPropertyTypes=true#code/C4TwDgpgBA6glsAFgewK7ADwBUA0UDSEIAggDanIDuEAJlBAB7AQB2NAzlANZHIBmULAD4oAXigBvAL5QAZJIBQUKAG1CIAJIsWEAE5Q4LKAFEGAY1KoaEDDxD9BedWQrUaQgLoB+AFxQdAG56ANwKUqEKoJBQAApwZlzEbNhO9EysHNy8AsJikjLyEkqqMQZG+B5+WCoxHqHhCpHg0HEJSTQAInDsAIbkVADySHrsKQRELlS0acxsnHYOuaLF8q2JybjjJP1uQiuwCCjoY847tEIRUdAAqixwyCwAYsi6AyykIG8Q6tgi4tJyRTKNRlLL2HKVWLxdadbp9VxDRAjE4XMIqBYQy7NKAANT6cBoADlkAAlCDsYC6eLMGgaAC2YBewBiPWAzF0LAGYGA9xYnH+BSgt15z1e70+Oh+RWU8Ko9MZumAhJ6dIgLLZehYfgpVJYAHMVHVirLKPKmcrVextZTDAajco4AzzSq1az2VqoDrbYbQg6nYqLeTrbq7fVUQprBYerpoGYHhSoMhubyrbj8UTSeSbWYaWbFer3VyefGInABAAKJPFvkAOkdCqVLvYAEogYnkyXigB6LtQAB6XjCQA), with `{} &`s added to get TypeScript to print the type nicely after narrowing. 🤷 that might be a bit heavyweight for just this one rule's types though.

lib/types/rules/ecmascript-6.d.ts
```diff
@@ -291,23 +313,39 @@ export interface ECMAScript6 extends Linter.RulesRecord {
...
-                }
+                | ({
+                      name: string;
+                      message?: string;
+                      importNames?: string[];
+                      allowImportNames?: string[];
+                  } & (
+                      | { importNames?: string[]; allowImportNames?: never }
+                      | { allowImportNames?: string[]; importNames?: never }
+                  ))
```
JoshuaKGoldberg, line 324: [Bug] `importNames` and `allowImportNames` will become `?: undefined` in this union, not `?: never`, even with [`exactOptionalPropertyTypes`](https://aka.ms/tsconfig#exactOptionalPropertyTypes) [[Playground showing `exactOptionalPropertyTypes`](https://www.typescriptlang.org/play/?exactOptionalPropertyTypes=true#code/C4TwDgpgBAKhDOwoF4oAoDeAoKUB2AhgLYQBcUiATgJZ4DmA3DlCfPAXRAPzlW2PNqRMAHtKwAHLEEPCsBr0A2gF0muAgBsNIgO4BJYWMnT4svktVYAvlABk6ZrgA+UDFCGjxU1mfn8VDFCa2vqGXiayeBAAbhCUUFaOUC5uwboGnsY+vH4WgR5G3jLkUbHxibgAlJVYWAAmEADGGgSU0I0ieIhQ0ZoArmSwCMBMWNQAZui9GgMAdAXhrJWuzNMDalAA9JtQAHpctbhrELNpoZlF8BvbuLj7hz39JwtZCNc7t-dWQA)].

This is because `never | undefined` is `undefined`, not `never`. So `importNames?: string[]` & `importNames?: never` becomes `importNames?: never`.

I'd rearrange this to something like the `ValidNoRestrictedImportPatternOptions` earlier in the file. As in, have a type for the common options (`name`, `message`) and then `&` it with other types.

JoshuaKGoldberg (changes requested):
Looks pretty close to complete to! I verified the types match what's on https://eslint.org/docs/latest/rules/no-restricted-imports.

The only blocker is that there's an inconsistency in the types: the _`pattern`_ options properly set up `?: never` with `exactOptionalPropertyTypes`, but  the _`path`_ options reduce down to `?: undefined`.

Left a couple of stylistic suggestions for the TypeScript syntax too.

lib/types/rules/ecmascript-6.d.ts
snitin315: Good catch! Updated.

lib/types/rules/ecmascript-6.d.ts
snitin315: >that might be a bit heavyweight for just this one rule's types though.

Yeah, will look into it if there are more rule types like this.

snitin315:
@JoshuaKGoldberg Thanks for the feedback, I've resolved your comments. PTAL

lib/types/rules/ecmascript-6.d.ts
```diff
@@ -27,6 +27,37 @@
 
 import { Linter } from "../index";
 
+interface NoRestrictedImportPathCommonOptions {
```
JoshuaKGoldberg, line 30: Sorry, I think I missed this before - I think it'd be good to `export` all the types in this file. That way they're easier for consumers to `import` ad hoc.

```suggestion
export interface NoRestrictedImportPathCommonOptions {
```

https://github.com/typescript-eslint/typescript-eslint/issues/7670 has points on more.

JoshuaKGoldberg (dismissed):
Looks great, yay!

Just one suggestion on `exports`, I trust whatever you do here 🙂

lib/types/rules/ecmascript-6.d.ts
snitin315: Exported all types 👍🏻

nzakas (approved):
LGTM. Would like another approval before merging.

aladdin-add (approved):
LGTM, thanks!

JoshuaKGoldberg (approved):
🔥

Issue #18709, opened 2024-07-22
Performance: long print time in stylish formatter's text-table for long report strings


JoshuaKGoldberg:
When running `npx eslint` on a repository with many errors containing long descriptions and rule names, the stylish formatter can take a surprisingly long time to print results. @johnsoncodehk and I discovered this on https://github.com/johnsoncodehk/core/tree/d47d660758485c04dc26948b6ee9a42ac51a50df: more than half the command's time is spent in `printResults`!

The problem is a very slow `/\s+$/` regex inside [`text-table`](https://unpkg.com/browse/text-table@0.2.0/index.js) on line 53. That regex is meant to remove all trailing whitespace in the string... which is now what [`String.prototype.trimEnd`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/trimEnd) does.

The following fix inside `text-table` reduces the stylish formatter's runtime to near-zero:

```diff
        return map(row, function (c, ix) {
            // ...
-        }).join(hsep).replace(/\s+$/, '');
+        }).join(hsep).trimEnd();
    }).join('\n');
```

Single example `time node node_modules/eslint/bin/eslint.js` measurements on my M1 Max Mac Studio, with screenshots of the equivalent attached `node --cpu-prof` traces:

<table>
<thead>
<tr>
<th>Measurement</th>
<th><code>time</code></th>
<th>Trace</th>
</tr>
</thead>
<tbody>
<tr>
<th>Baseline</th>
<td><code>17.35s</code></td>
<td><img alt="Performance trace showing over half the time in a printResults call" src="https://github.com/user-attachments/assets/79aaa185-319c-4309-a19a-55281b4d7ee1" /></td>
</tr>
<tr>
<th>Fixed</th>
<td><code>8.70s</code></td>
<td><img alt="Performance trace showing almost no time in a printResults call" src="https://github.com/user-attachments/assets/b6eb7086-acff-4e10-9842-95f17ac05391" /></td>
</tr>
</tbody>
</table>

* [Baseline with replace regex.cpuprofile](https://github.com/user-attachments/files/16337429/Baseline.with.replace.regex.cpuprofile)
* [Fixed with string trimEnd.cpuprofile](https://github.com/user-attachments/files/16337428/Fixed.with.string.trimEnd.cpuprofile)

<details>
<summary>Here is one of the strings being trimmed...</summary>

```js
[...]

PR #19061, opened 2024-10-27
perf: optimize `text-table` by replacing regex with `trimEnd`

snitin315:
#### Prerequisites checklist

[ ] Documentation update
[ ] Add autofix to a rule
[ ] Add a CLI option
[ ] Add something to the core
[x] Other, please explain: Performance improvements

Fix https://github.com/eslint/eslint/issues/18709

I disabled some lint rules to avoid unwanted refactors in 3rd party code.

### DIFF
--- a/lib/cli-engine/formatters/text-table.js
+++ b/lib/cli-engine/formatters/text-table.js
@@ -1,0 +1,114 @@
+/**
+ * @fileoverview Optimized version of the `text-table` npm module to improve performance by replacing inefficient regex-based
+ * whitespace trimming with a modern built-in method.
+ *
+ * @author Nitin Kumar
+ *
+ * This modification addresses a performance issue reported in https://github.com/eslint/eslint/issues/18709
+ *
+ * The `text-table` module is published under the MIT License. For the original source, refer to:
+ * https://www.npmjs.com/package/text-table.
+ */
+
+"use strict";
+
+function dotindex(c) {
+    const m = /\.[^.]*$/u.exec(c);
+
+    return m ? m.index + 1 : c.length;
+}
+
+function reduce(xs, f, init) {
+    if (xs.reduce) {
+        return xs.reduce(f, init);
+    }
+    let i = 0;
+    const acc = arguments.length >= 3 ? init : xs[i++];
+
+    for (; i < xs.length; i++) {
+        f(acc, xs[i], i);
+    }
+    return acc;
+}
+
+function forEach(xs, f) {
+    if (xs.forEach) {
+        return xs.forEach(f);
+    }
+    for (let i = 0; i < xs.length; i++) {
+        f.call(xs, xs[i], i);
+    }
+}
+
+function map(xs, f) {
+    if (xs.map) {
+        return xs.map(f);
+    }
+    const res = [];
+
+    for (let i = 0; i < xs.length; i++) {
+        res.push(f.call(xs, xs[i], i));
+    }
+    return res;
+}
+
+module.exports = function(rows_, opts) {
+    if (!opts) {
+        opts = {};
+    }
+    const hsep = typeof opts.hsep === "undefined" ? "  " : opts.hsep;
+    const align = opts.align || [];
+    const stringLength = opts.stringLength ||
+        function(s) {
+            return String(s).length;
+        };
+    const dotsizes = reduce(rows_, (acc, row) => {
+        forEach(row, (c, ix) => {
+            const n = dotindex(c);
+
+            if (!acc[ix] || n > acc[ix]) {
+                acc[ix] = n;
+            }
+        });
+        return acc;
+    }, []);
+
+    const rows = map(rows_, row => map(row, (c_, ix) => {
+        const c = String(c_);
+
+        if (align[ix] === ".") {
+            const index = dotindex(c);
+            const size = dotsizes[ix] + (/\./u.test(c) ? 1 : 2) -
+                    (stringLength(c) - index);
+
+            return c + Array(size).join(" ");
+        }
+        return c;
+    }));
+
+    const sizes = reduce(rows, (acc, row) => {
+        forEach(row, (c, ix) => {
+            const n = stringLength(c);
+
+            if (!acc[ix] || n > acc[ix]) {
+                acc[ix] = n;
+            }
+        });
+        return acc;
+    }, []);
+
+    return map(rows, row => map(row, (c, ix) => {
+        const n = (sizes[ix] - stringLength(c)) || 0;
+        const s = Array(Math.max(n + 1, 1)).join(" ");
+
+        if (align[ix] === "r" || align[ix] === ".") {
+            return s + c;
+        }
+        if (align[ix] === "c") {
+            return Array(Math.ceil(n / 2 + 1)).join(" ") +
+                    c + Array(Math.floor(n / 2 + 1)).join(" ");
+        }
+
+        return c + s;
+    }).join(hsep).trimEnd()).join("\n");
+};
--- a/eslint.config.js
+++ b/eslint.config.js
@@ -317,5 +317,20 @@ module.exports = [
              */
             "n/no-unsupported-features/node-builtins": [2, { ignores: ["module.enableCompileCache"] }]
         }
+    },
+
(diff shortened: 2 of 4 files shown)


lib/cli-engine/formatters/text-table.js
```diff
@@ -0,0 +1,114 @@
+/**
+ * @fileoverview Optimized version of the `text-table` npm module to improve performance by replacing inefficient regex-based
+ * whitespace trimming with a modern built-in method.
+ *
+ * @author Nitin Kumar
```
nzakas, line 5: Because you only modified this code and didn't create it, we should skip this.
```suggestion
```

lib/cli-engine/formatters/text-table.js
```diff
@@ -0,0 +1,114 @@
...
+ *
+ * This modification addresses a performance issue reported in https://github.com/eslint/eslint/issues/18709
+ *
+ * The `text-table` module is published under the MIT License. For the original source, refer to:
```
nzakas, line 9: We need to include the full MIT License per the terms of the license:

> The above copyright notice and this permission notice shall be included in all copies or substantial portions of the Software.

lib/cli-engine/formatters/text-table.js
mdjermanovic, line 1: Can we move this file somewhere else, e.g. to `lib/shared`? If it's in `lib/cli-engine/formatters`, then it can be used as a formatter (`-f text-table`). Also, I believe our release process would treat it as a formatter and eventually crash.

https://github.com/eslint/eslint/blob/425202ed49a1372c1719d4e7b48d0fbdda8af9fa/Makefile.js#L475

(it already happened before: https://github.com/eslint/eslint/pull/14612).

snitin315: Oh! I'll update the PR.

lib/cli-engine/formatters/text-table.js
```diff
@@ -0,0 +1,114 @@
...
+    if (xs.map) {
+        return xs.map(f);
+    }
+    const res = [];
+
+    for (let i = 0; i < xs.length; i++) {
+        res.push(f.call(xs, xs[i], i));
+    }
+    return res;
+}
```
JoshuaKGoldberg, line 53: ESLint doesn't support runtimes that don't have `Array`'s `forEach`, `map`, or `reduce`. These polyfills aren't needed.

```suggestion
```

For reference, my reimplementation that doesn't include any polyfills for old runtimes: 
https://github.com/JoshuaKGoldberg/text-table-fast/blob/a835b26d31ac4645900107ce7640fb950f5b0ac9/src/index.ts

lib/cli-engine/formatters/text-table.js
```diff
@@ -0,0 +1,114 @@
...
+
+module.exports = function(rows_, opts) {
+    if (!opts) {
+        opts = {};
+    }
+    const hsep = typeof opts.hsep === "undefined" ? "  " : opts.hsep;
```
JoshuaKGoldberg, line 59: The `stylish` formatter only ever calls to `table()` once. I think we can remove dead code in here by hardcoding in only what's actually used. For example, `opts` is always defined, and `hsep` can just be `"  "`.

JoshuaKGoldberg (changes requested):
Functional changes LGTM - it's easier to review when they're so close to the original! 🙂 

But, something like half the file is dead code from over a decade ago. It feels odd to me to include so much of that in ESLint. Proposal: how about treating this as a fork of `text-table` and removing the portions that are never used?

lib/cli-engine/formatters/text-table.js
nzakas: :+1:

snitin315: Yeah, makes sense. Removed.

lib/cli-engine/formatters/text-table.js
snitin315: Updated 👍🏻

lib/shared/text-table.js
```diff
@@ -0,0 +1,110 @@
...
+ *
+ * This modification addresses a performance issue reported in https://github.com/eslint/eslint/issues/18709
+ *
+ * The `text-table` module is published under the MIT License. For the original source, refer to:
+ * https://www.npmjs.com/package/text-table.
+ */
+
+/*
+ * MIT License
+ * Copyright (c) Microsoft Corporation.
```
mdjermanovic, line 13: Why Microsoft?

snitin315: Yeah, this is a copy paste error. I'll update

lib/shared/text-table.js
```diff
@@ -0,0 +1,110 @@
...
+                const index = dotindex(c);
+                const size =
+                    dotsizes[ix] +
+                    (/\./u.test(c) ? 1 : 2) -
+                    (stringLength(c) - index);
+
+                return c + Array(size).join(" ");
+            }
+            return c;
+        }));
```
JoshuaKGoldberg, line 74: `align` is always `['', 'r', 'l']` and `rows` are always `string[]`, so there's no need for the `String(c_)` or the `if (align[ix] === ".")`:

```suggestion
```

lib/shared/text-table.js
```diff
@@ -0,0 +1,110 @@
...
+                    if (align[ix] === "r" || align[ix] === ".") {
+                        return s + c;
+                    }
+                    if (align[ix] === "c") {
+                        return (
+                            Array(Math.ceil(n / 2 + 1)).join(" ") +
+                            c +
+                            Array(Math.floor(n / 2 + 1)).join(" ")
+                        );
+                    }
```
JoshuaKGoldberg, line 103: Since `align` is never `'c'`, and `s + c` is the same as `c + s`:

```suggestion
```

lib/shared/text-table.js
```diff
@@ -0,0 +1,110 @@
...
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE
+ */
+
+"use strict";
+
+/**
+ *
+ * @param c
+ */
```
JoshuaKGoldberg, line 38:
```suggestion
```

JoshuaKGoldberg (changes requested):
There's some more code I think we can directly remove. But consider this a ✅ otherwise!

lib/shared/text-table.js
snitin315: Updated.

lib/shared/text-table.js
snitin315: Removed, and we no longer need to pass `align` as an option too.

lib/shared/text-table.js
snitin315: `String(c_)` is required because the line & column numbers are actually of type number.

https://github.com/eslint/eslint/blob/db0b844a66ee25483f9619d04346de1a2a0d79fa/lib/cli-engine/formatters/stylish.js#L65-L66

So we can either keep String(c_) here or change the code in stylish to `String(message.line)`. I just updated the code to `String(message.line || 0)` to avoid unnecessary looping later.

JoshuaKGoldberg (dismissed):
💯 awesome!

mdjermanovic (changes requested):
line:column was right-aligned:

image

But after this change, it would be left-aligned, and also seem to be breaking the table formatting:

image

lib/shared/text-table.js
snitin315: It seems we need the `s+c` check, otherwise regression as reported in https://github.com/eslint/eslint/pull/19061#pullrequestreview-2411621760

snitin315:
@mdjermanovic Good catch! Fixed in https://github.com/eslint/eslint/pull/19061/commits/89efc43efd17e12d505ce62e3a49de3a8860f7b0
Screenshot 2024-11-03 at 8 15 35 AM

mdjermanovic (approved):
LGTM, thanks! Leaving open for @nzakas and @JoshuaKGoldberg to verify.

nzakas (approved):
LGTM. Waiting for @JoshuaKGoldberg to verify before merging.

JoshuaKGoldberg (approved):
Looks great to me, thanks!

PR #19092, opened 2024-11-03
fix: throw on `ENFILE`

LiviaMedeiros:
#### Prerequisites checklist

[ ] Documentation update
[ ] Add autofix to a rule
[ ] Add a CLI option
[ ] Add something to the core
[ ] Other, please explain:

Retrying on `EMFILE` (when `eslint` hits its own per-process limit) makes sense here.

However, hitting `ENFILE` (when global limit is exhausted) indicates an issue in system configuration/condition. Constantly retrying and keeping number of opened files at maximum would render whole system unusable: it wouldn't be able to spawn new processes, to start ssh sessions, to open log files; it might cause any service to crash or misbehave, etc. In this case, we should just throw and exit immediately.

It would make sense to search for alternative approach here, since retrier would constantly hit `EMFILE`, causing unnecessary overhead.

IMHO the best approach would be:
- Start with high-concurrency queue
- If we catch `ENFILE` or any other non-`EMFILE` error, throw immediately
- If we catch `EMFILE`, lower the concurrency limit (for example, start with `128` and do `concurrency /= 2` each time we hit it; rethrow `EMFILE` if it happened on `concurrency === 1`) and retry
- (optional) Print a warning each time we got error and had to lower concurrency
- (optional) Make default concurrency limit configurable and/or set it as relative to `os.cpus().length` by default in runtime

### DIFF
--- a/lib/eslint/eslint.js
+++ b/lib/eslint/eslint.js
@@ -745,8 +745,7 @@ class ESLint {
             errorOnUnmatchedPattern
         });
         const controller = new AbortController();
-        const retryCodes = new Set(["ENFILE", "EMFILE"]);
-        const retrier = new Retrier(error => retryCodes.has(error.code), { concurrency: 100 });
+        const retrier = new Retrier(error => error.code === "EMFILE", { concurrency: 100 });
 
         debug(`${filePaths.length} files found in: ${Date.now() - startTime}ms`);
 


nzakas:
It's fairly common in Node.js utilities to treat EMFILE and ENFILE the same ([example](https://github.com/isaacs/node-graceful-fs/blob/234379906b7d2f4c9cfeb412d2516f42b0fb4953/graceful-fs.js#L119)). If you'd like to propose a different behavior, please open an issue to discuss first.

PR #19231, opened 2024-12-08
docs: mark rules that are frozen with ❄️

### LINKED ISSUE #18694: Docs: Clearly mark rules that are frozen

amareshsm:
#### Prerequisites checklist

[x] Documentation update
[ ] Add autofix to a rule
[ ] Add a CLI option
[ ] Add something to the core
[ ] Other, please explain:

fixes #18694


docs/src/_includes/components/rule.macro.html
```diff
@@ -31,6 +31,8 @@
         <span class="visually-hidden">Categories:</span>
         {%- if params.deprecated -%}
             <p class="rule__categories__type">❌</p>
+            {%- elif params.categories and params.categories.frozen %}
+                <p class="rule__categories__type"> ❄️ <span class="visually-hidden">Frozen</span> </p>
```
fasttime, line 35: `no-extra-boolean-cast` is both recommended and frozen, but after this change, the recommended symbol won't be shown any more on the overview:

image

Maybe it would be better to show both symbols, especially since we may decide to freeze more rules later?

fasttime (commented):
`no-undefined` should be marked as frozen.

docs/src/_includes/components/rule.macro.html
nzakas: The layout is really only designed to support three icons. So I'd say if something is both recommended and frozen then we should favor showing the recommended checkmark. We can still show the frozen icon on the rule page.

amareshsm: image

Would it look good to display the frozen icon next to the rule name?

fasttime: > Would it look good to display the frozen icon next to the rule name?

Works for me.

nzakas: Ooh yes, good idea @amareshsm :+1:

amareshsm: sure I will make the changes accordingly.

docs/src/assets/scss/components/rules.scss
```diff
@@ -90,6 +90,16 @@
...
 }
 
+.rule__name_wrapper {
+    display: flex;
+    align-items: center;
+    gap: 0.5rem;
+
+    .frozen {
+        font-size: 12px;
+    }
```
Tanujkanti4441, line 100: I feel we can increase it's size a bit to make it more noticeable?
right now
Screenshot 2024-12-17 191833
we can increase it to `.875rem` as other icons.
Screenshot 2024-12-17 191947
or to `1rem`
Screenshot 2024-12-17 192013

amareshsm: ya, we can set  `.875rem` it looks better.

Tanujkanti4441 (approved):
LGTM, thanks!

Would like another team member to review this.

fasttime (approved):
LGTM, thanks!

PR #19351, opened 2025-01-20
docs: clarify frozen rule description

### LINKED ISSUE #19346: Docs: improve description of frozen rules for clarity
### Docs page(s)

https://eslint.org/docs/latest/rules/no-magic-numbers#:~:text=This%20rule%20is%20currently%20frozen%20and%20is%20not%20accepting%20changes. and others, see the full list of rules here: https://github.com/eslint/eslint/issues/18694#issuecomment-2355493603

`This rule is currently frozen and is not accepting changes.` For me, this means you won't accept any changes, even bug fixes or documentation changes. Only after reading https://github.com/eslint/eslint/issues/18439#issuecomment-2105500273, did I get the difference between frozen and deprecated rules.

You could describe with more detail what frozen rules like `typescript-eslint` do: `This rule is feature frozen: it will no longer receive new features such as new options. It still will accept bug and documentation fixes for its existing area of features.`

See https://typescript-eslint.io/rules/member-ordering/#:~:text=This%20rule%20is%20feature%20frozen%3A%20it%20will%20no%20longer%20receive%20new%20features%20such%20as%20new%20options.%20It%20still%20will%20accept%20bug%20and%20documentation%20fixes%20for%20its%20existing%20area%20of%20features.

### Participation

### Additional comments

_No response_

rakleed:
#### Prerequisites checklist

[x] Documentation update
[ ] Add autofix to a rule
[ ] Add a CLI option
[ ] Add something to the core
[ ] Other, please explain:

Related #19346

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,7 +36,7 @@ ESLint is a tool for identifying and reporting on patterns found in ECMAScript/J
 1. [Releases](#releases)
 1. [Security Policy](#security-policy)
 1. [Semantic Versioning Policy](#semantic-versioning-policy)
-1. [Frozen Rule Updates](#frozen-rule-updates)
+1. [Stylistic Rule Updates](#stylistic-rule-updates)
 1. [License](#license)
 1. [Team](#team)
 1. [Sponsors](#sponsors)
@@ -192,7 +192,7 @@ ESLint follows [semantic versioning](https://semver.org). However, due to the na
 
 According to our policy, any minor update may report more linting errors than the previous release (ex: from a bug fix). As such, we recommend using the tilde (`~`) in `package.json` e.g. `"eslint": "~3.1.0"` to guarantee the results of your builds.
 
-## Frozen Rule Updates
+## Stylistic Rule Updates
 
 Stylistic rules are frozen according to [our policy](https://eslint.org/blog/2020/05/changes-to-rules-policies) on how we evaluate new rules and rule changes.
 This means:
--- a/README.md
+++ b/README.md
@@ -36,7 +36,7 @@ ESLint is a tool for identifying and reporting on patterns found in ECMAScript/J
 1. [Releases](#releases)
 1. [Security Policy](#security-policy)
 1. [Semantic Versioning Policy](#semantic-versioning-policy)
-1. [Stylistic Rule Updates](#stylistic-rule-updates)
+1. [Frozen Rule Updates](#frozen-rule-updates)
 1. [License](#license)
 1. [Team](#team)
 1. [Sponsors](#sponsors)
@@ -192,7 +192,7 @@ ESLint follows [semantic versioning](https://semver.org). However, due to the na
 
 According to our policy, any minor update may report more linting errors than the previous release (ex: from a bug fix). As such, we recommend using the tilde (`~`) in `package.json` e.g. `"eslint": "~3.1.0"` to guarantee the results of your builds.
 
-## Stylistic Rule Updates
+## Frozen Rule Updates
 
 Stylistic rules are frozen according to [our policy](https://eslint.org/blog/2020/05/changes-to-rules-policies) on how we evaluate new rules and rule changes.
 This means:
--- a/docs/src/_includes/components/rule-categories.macro.html
+++ b/docs/src/_includes/components/rule-categories.macro.html
@@ -32,7 +32,8 @@
     <div class="rule-category">
         <span class="rule-category__icon">❄️ <span class="visually-hidden">Frozen</span></span>
         <p class="rule-category__description">
-            This rule is currently frozen and is not accepting changes.
+            This rule is currently frozen and is not accepting <a
+                href="https://github.com/eslint/eslint/#stylistic-rule-updates">feature requests</a>
         </p>
     </div>
     {%- endif -%}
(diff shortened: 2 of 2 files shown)


docs/src/_includes/components/rule-categories.macro.html
```diff
@@ -32,7 +32,8 @@
     <div class="rule-category">
         <span class="rule-category__icon">❄️ <span class="visually-hidden">Frozen</span></span>
         <p class="rule-category__description">
-            This rule is currently frozen and is not accepting changes.
+            This rule is currently frozen and is not accepting <a
+                href="https://github.com/eslint/eslint/#stylistic-rule-updates">feature requests</a>
```
rakleed, line 36: What do you think about this solution? I think this will allow people who want to learn more to find information about it faster. Since it is not at all obvious that you need to look for it in the README

amareshsm: Great! We can include a link. Let's get other people's opinions too.

Tanujkanti4441: The link seems useful to me, but how about also updating the description of frozen rules in `Readme` as it only mentions stylistic rules as frozen?

rakleed: I have changed the title in README per https://github.com/eslint/eslint/issues/19346#issuecomment-2598878884.

What is the best way to change the description? Should I completely remove the mention of stylistic rules and just leave general information about frozen rules?

README.md
```diff
@@ -192,7 +192,7 @@ ESLint follows [semantic versioning](https://semver.org). However, due to the na
 
 According to our policy, any minor update may report more linting errors than the previous release (ex: from a bug fix). As such, we recommend using the tilde (`~`) in `package.json` e.g. `"eslint": "~3.1.0"` to guarantee the results of your builds.
 
-## Stylistic Rule Updates
+## Frozen Rule Updates
```
JoshuaKGoldberg, line 195: I don't think this change is for the better. My assumption is explicitly making the section about _"stylistic"_ rules better surfaces what they're for. Someone skimming through the ToC and seeing a _"frozen rule updates"_ point won't be informed about why the rules are frozen.

docs/src/_includes/components/rule-categories.macro.html
```diff
@@ -32,7 +32,8 @@
     <div class="rule-category">
         <span class="rule-category__icon">❄️ <span class="visually-hidden">Frozen</span></span>
         <p class="rule-category__description">
-            This rule is currently frozen and is not accepting changes.
+            This rule is currently frozen and is not accepting <a
+                href="https://github.com/eslint/eslint/#frozen-rule-updates">feature requests</a>
```
JoshuaKGoldberg, line 36: > feature requests

[Accessibility] From [Understanding SC 2.4.4 Link Purpose (In Context)](https://www.w3.org/WAI/WCAG21/Understanding/link-purpose-in-context.html): links need to have text that makes sense on their own. This one reads as the text _"feature requests"_ which isn't a very helpful description or title for the page it links to.

Two common ways of fixing this are to either:

* Add a `title` attribute to the link: `<a href="..." title="ESLint Frozen rule updates">feature requests</a>
* Include more text in the link: `<a href="...">not accepting feature requests</a>`

I'd personally use the `title` approach so as to not change what the link visually looks like.

JoshuaKGoldberg, line 36: > https://github.com/eslint/eslint/#frozen-rule-updates

[Suggestion] I do find it a little odd to have the _eslint.org_ website link to the GitHub repository. That's not common on rule pages, and certainly not this high up in the page. How about instead linking to a good resource on the website, maybe <https://eslint.org/docs/latest/use/core-concepts/glossary#stylistic-rule>?

JoshuaKGoldberg (changes requested):
Thanks for pushing this forward! Left a couple of suggestions, as well as a note on an accessibility bug.

rakleed:
> Per [this comment](https://github.com/eslint/eslint/issues/19346#issuecomment-2601032617), let's just update the text in the banner and not the README in this PR.

@nzakas I actually like this suggestion: https://github.com/eslint/eslint/pull/19351#discussion_r1923668993. However, I guess it would be nice to add information from README to the glossary's description of `Stylistic (Rule)` first.

Should I then completely remove the link to README and create a separate issue for updating the glossary description? Then, I can add this link to the description of the frozen rules.

nzakas:
> > Per [this comment](https://github.com/eslint/eslint/issues/19346#issuecomment-2601032617), let's just update the text in the banner and not the README in this PR.
> 
> @nzakas I actually like this suggestion: [#19351 (comment)](https://github.com/eslint/eslint/pull/19351#discussion_r1923668993). However, I guess it would be nice to add information from README to the glossary's description of `Stylistic (Rule)` first.

I actually don't think that this is the correct solution. Important information like this doesn't belong in the glossary, which is list of words and phrases to know. We do need a place on the website that describes frozen rules, but the glossary isn't the right place for it.

> Should I then completely remove the link to README and create a separate issue for updating the glossary description? Then, I can add this link to the description of the frozen rules.

Yes. Let's just update the text in the banner for now while we figure out the next steps.

rakleed:
> Let's just update the text in the banner for now while we figure out the next steps.

Done.

nzakas:
We still need to remove the link from the banner.

JoshuaKGoldberg (approved):
Thanks!

mdjermanovic (approved):
LGTM, thanks! Leaving it open for @nzakas to verify.

nzakas (approved):
LGTM. Thanks.

PR #19431, opened 2025-02-17
feat: support TypeScript syntax in `default-param-last`

JoshuaKGoldberg:
#### Prerequisites checklist

[ ] Documentation update
[ ] Add autofix to a rule
[ ] Add a CLI option
[ ] Add something to the core
[ ] Other, please explain:

Starts on #19173.

Updates `default-param-last` to account for TypeScript syntax features as handled in the [`@typescript-eslint/default-param-last`](https://typescript-eslint.io/rules/default-param-last) extension rule:

* Class parameter properties
* Type annotations

Includes tests equivalent to what the extension rule currently tests that's exclusive to TypeScript. They're in a separate `ruleTesterTypeScript` instance that uses `@typescript-eslint/parser`.

Also updates `tools/check-rule-examples.js` to allow TypeScript code block. Previously it only allowed JavaScript.

Should we also update documentation?

### DIFF
--- a/docs/src/rules/default-param-last.md
+++ b/docs/src/rules/default-param-last.md
@@ -41,6 +41,34 @@ Examples of **correct** code for this rule:
 /* eslint default-param-last: ["error"] */
 
 function f(a, b = 0) {}
+
+function g(a, b = 0, c = 0) {}
+```
+
+:::
+
+This rule additionally supports TypeScript type syntax.
+
+Examples of **incorrect** TypeScript code for this rule:
+
+::: incorrect
+
+```ts
+/* eslint default-param-last: ["error"] */
+
+function h(a = 0, b: number) {}
+```
+
+:::
+
+Examples of **correct** TypeScript code for this rule:
+
+::: correct
+
+```ts
+/* eslint default-param-last: ["error"] */
+
+function h(a = 0, b?: number) {}
 ```
 
 :::
--- a/lib/rules/default-param-last.js
+++ b/lib/rules/default-param-last.js
@@ -5,6 +5,35 @@
 
 "use strict";
 
+/**
+ * checks if node is optional parameter
+ * @param {ASTNode} node the node to be evaluated
+ * @returns {boolean} whether the node is optional
+ */
+function isOptionalParam(node) {
+    return (
+        (node.type === "ArrayPattern" ||
+            node.type === "AssignmentPattern" ||
+            node.type === "Identifier" ||
+            node.type === "ObjectPattern" ||
+            node.type === "RestElement") &&
+        node.optional
+    );
+}
+
+/**
+ * checks if node is plain parameter
+ * @param {ASTNode} node the node to be evaluated
+ * @returns {boolean} whether the node is plain
+ */
+function isPlainParam(node) {
+    return !(
+        node.type === "AssignmentPattern" ||
+        node.type === "RestElement" ||
+        isOptionalParam(node)
+    );
+}
+
 /** @type {import('../shared/types').Rule} */
 module.exports = {
     meta: {
@@ -35,19 +64,24 @@ module.exports = {
             let hasSeenPlainParam = false;
 
             for (let i = node.params.length - 1; i >= 0; i -= 1) {
-                const param = node.params[i];
+                const current = node.params[i];
+                const param =
+                    current.type === "TSParameterProperty"
+                        ? current.parameter
+                        : current;
 
-                if (
-                    param.type !== "AssignmentPattern" &&
-                    param.type !== "RestElement"
-                ) {
+                if (isPlainParam(param)) {
                     hasSeenPlainParam = true;
                     continue;
                 }
 
-                if (hasSeenPlainParam && param.type === "AssignmentPattern") {
+                if (
+                    hasSeenPlainParam &&
+                    (isOptionalParam(param) ||
+                    param.type === "AssignmentPattern")
+                ) {
                     context.report({
-                        node: param,
+                        node: current,
                         messageId: "shouldBeLast"
                     });
                 }
--- a/lib/rules/default-param-last.js
+++ b/lib/rules/default-param-last.js
@@ -6,9 +6,9 @@
 "use strict";
 
 /**
- * checks if node is optional parameter
+ * Checks if node is a TypeScript-optional parameter.
  * @param {ASTNode} node the node to be evaluated
- * @returns {boolean} whether the node is optional
+ * @returns {boolean} true if the node is optional, false if not.
  */
 function isOptionalParam(node) {
     return (
--- a/lib/rules/default-param-last.js
+++ b/lib/rules/default-param-last.js
@@ -27,7 +27,8 @@ module.exports = {
             description: "Enforce default parameters to be last",
             recommended: false,
             frozen: true,
-            url: "https://eslint.org/docs/latest/rules/default-param-last"
+            url: "https://eslint.org/docs/latest/rules/default-param-last",
+            typescript: "syntax"
         },
 
         schema: [],
--- a/lib/rules/default-param-last.js
+++ b/lib/rules/default-param-last.js
@@ -6,31 +6,15 @@
 "use strict";
 
 /**
- * Checks if node is a TypeScript-optional parameter.
+ * Checks if node is required: i.e. does not have a default value or ? optional indicator.
  * @param {ASTNode} node the node to be evaluated
- * @returns {boolean} true if the node is optional, false if not.
+ * @returns {boolean} true if the node is required, false if not.
  */
-function isOptionalParam(node) {
-    return (
-        (node.type === "ArrayPattern" ||
-            node.type === "AssignmentPattern" ||
-            node.type === "Identifier" ||
-            node.type === "ObjectPattern" ||
-            node.type === "RestElement") &&
-        node.optional
-    );
-}
-
-/**
- * checks if node is plain parameter
- * @param {ASTNode} node the node to be evaluated
- * @returns {boolean} whether the node is plain
- */
-function isPlainParam(node) {
+function isRequiredParameter(node) {
     return !(
         node.type === "AssignmentPattern" ||
         node.type === "RestElement" ||
-        isOptionalParam(node)
+        node.optional
     );
 }
 
@@ -61,7 +45,7 @@ module.exports = {
          * @returns {void}
          */
         function handleFunction(node) {
-            let hasSeenPlainParam = false;
+            let hasSeenRequiredParameter = false;
 
             for (let i = node.params.length - 1; i >= 0; i -= 1) {
                 const current = node.params[i];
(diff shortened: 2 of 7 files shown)


lib/rules/default-param-last.js
```diff
@@ -5,6 +5,35 @@
...
+/**
+ * checks if node is plain parameter
+ * @param {ASTNode} node the node to be evaluated
+ * @returns {boolean} whether the node is plain
+ */
+function isPlainParam(node) {
+    return !(
+        node.type === "AssignmentPattern" ||
+        node.type === "RestElement" ||
+        isOptionalParam(node)
```
nzakas, line 33: `isOptionalParam` is also checking for `"RestElement"` and `"AssignmentPattern"`. It seems like there should be a way to avoid checking this twice when `isPlainParam()` is called.

lib/rules/default-param-last.js
```diff
@@ -5,6 +5,35 @@
...
+            node.type === "AssignmentPattern" ||
+            node.type === "Identifier" ||
+            node.type === "ObjectPattern" ||
+            node.type === "RestElement") &&
+        node.optional
+    );
+}
+
+/**
+ * checks if node is plain parameter
```
nzakas, line 25: For clarity: Can you explain what "plain" means here? I'm guessing you mean it's just an identifier but I'm not 100% sure.

lib/rules/default-param-last.js
```diff
@@ -5,6 +5,35 @@
 
 "use strict";
 
+/**
+ * checks if node is optional parameter
```
nzakas, line 9: For clarity.
```suggestion
 * Checks if node is a TypeScript-optional parameter.
```

lib/rules/default-param-last.js
```diff
@@ -5,6 +5,35 @@
...
+ * @param {ASTNode} node the node to be evaluated
+ * @returns {boolean} whether the node is optional
```
nzakas, line 11: For clarity.

```suggestion
 * @returns {boolean} true if the node is optional, false if not.
```

lib/rules/default-param-last.js
```diff
@@ -5,6 +5,35 @@
...
+ */
+function isOptionalParam(node) {
+    return (
+        (node.type === "ArrayPattern" ||
+            node.type === "AssignmentPattern" ||
+            node.type === "Identifier" ||
+            node.type === "ObjectPattern" ||
+            node.type === "RestElement") &&
+        node.optional
```
nzakas, line 20: I'm curious about this logic. Why isn't it enough to just check `node.optional`?

If we do need to check for all of these node types, then it's more efficient to place them all in a `Set` and use `has()`.

nzakas (changes requested):
Thanks for getting this started. I left some points of clarification.

With regards to documentation, I think it would be good to add some code examples showing TypeScript.

I'm also wondering if maybe we should add something into `meta` indicating that this rule is TypeScript-aware?

lib/rules/default-param-last.js
JoshuaKGoldberg: Ha, yeah, this is a code style thing. In the typescript-eslint repo, since `TSESTree.Node` is strictly typed as a union of nodes, you can't access arbitrary properties that only exist on some nodes. We're very cautious with arbitrary property accesses and require type narrowing: e.g. https://github.com/typescript-eslint/typescript-eslint/issues/8226.

Since we don't do that here, I simplified to just check `node.optional` inline.

lib/rules/default-param-last.js
JoshuaKGoldberg: Existing code that I also found confusing. "Plain" meant non-optional, or in other words, required. I switched the naming around.

lib/rules/default-param-last.js
JoshuaKGoldberg, line 1: > With regards to documentation, I think it would be good to add some code examples showing TypeScript.

Oh and, this is currently not supported that I can see in the docs infrastructure. Trying to add a `ts` code block:

```plaintext

/Users/josh/repos/eslint/docs/src/rules/default-param-last.md
  33:20  error  Unexpected lint error found: Parsing error: Unexpected token :
  49:20  error  Unexpected lint error found: Parsing error: Unexpected token ?
  60:4   error  Nonstandard language tag 'ts': use one of 'javascript', 'js' or 'jsx'
  63:20  error  Unexpected lint error found: Parsing error: Unexpected token :
  72:4   error  Nonstandard language tag 'ts': use one of 'javascript', 'js' or 'jsx'
  75:20  error  Unexpected lint error found: Parsing error: Unexpected token ?
```

JoshuaKGoldberg: b74684e90

lib/rules/default-param-last.js
```diff
@@ -14,7 +27,8 @@ module.exports = {
             description: "Enforce default parameters to be last",
             recommended: false,
             frozen: true,
-            url: "https://eslint.org/docs/latest/rules/default-param-last"
+            url: "https://eslint.org/docs/latest/rules/default-param-last",
+            typescript: "syntax"
```
JoshuaKGoldberg, line 31: > I'm also wondering if maybe we should add something into `meta` indicating that this rule is TypeScript-aware?

I found this to be a surprisingly difficult design question. The property needs to indicate:

* It _does_ support TypeScript _syntax_
* It _doesn't_ support TypeScript _types_

...and I couldn't figure out any nice succinct single-word property names for this. Using `typescript: "syntax"` here explicitly tells folks it's _only_ syntax, not types. It leaves an option for custom parsers like tseslint to later put in `"types"` too, maybe. But I'm not confident in this.

JoshuaKGoldberg:
No, that's out of scope. See https://github.com/eslint/eslint/issues/19173 -> https://github.com/eslint/eslint/issues/19173#issuecomment-2619544962.

JoshuaKGoldberg:
Depends what you mean by _"TypeScript linting"_.

For _syntax-only_ linting with just ESLint core rules, technically yes. See https://github.com/eslint/eslint/issues/19173#issuecomment-2619649206 -> https://github.com/eslint/eslint/issues/19173#issuecomment-2619975111 for a list of what will switch from needing a typescript-eslint extension rule to working out-of-the-box in ESLint core.

_Typed linting_, however, will not be supported in ESLint core. You'll still need typescript-eslint or an equivalent for that. See https://github.com/eslint/eslint/issues/19173 -> https://github.com/eslint/eslint/issues/19173#issuecomment-2619544962.

ESLint core also presently doesn't have any plans that I know of to support TypeScript-specific rules that don't need typed linting. Example: [`@typescript-eslint/no-empty-object-type`](https://typescript-eslint.io/rules/no-empty-object-type).

tl;dr: no, the strong recommendation from typescript-eslint going forward will still be to use typescript-eslint when linting TypeScript code.

nzakas:
We discussed today and we're at least going to start with this:

* `meta.language = "javascript"` - to indicate that this rule is meant for JavaScript or a derivative of it
* `meta.dialects = ["javascript", "typescript"]` - to indicate that this rule is known to work with both plain JavaScript and TypeScript.

For now, I think this is enough and we don't need to get to deep into the difference between being syntax-aware and type-aware. We're not planning on displaying this info anywhere so in the short-term it will be mostly for our own edification but we are also looking at how we can use this info to, for example, automatically disable JavaScript-specific rules when linting JSON.

lib/rules/default-param-last.js
```diff
@@ -5,16 +5,32 @@
...
+        node.type === "RestElement" ||
+        node.optional
+    );
+}
+
 /** @type {import('../shared/types').Rule} */
 module.exports = {
     meta: {
+        dialects: ["javascript", "typescript"],
+        language: "javascript",
```
JoshuaKGoldberg, line 25: Coming over from https://github.com/eslint/eslint/pull/19431#issuecomment-2672813876:

> * `meta.language = "javascript"` - to indicate that this rule is meant for JavaScript or a derivative of it
> * `meta.dialects = ["javascript", "typescript"]` - to indicate that this rule is known to work with both plain JavaScript and TypeScript.

Can I suggest a `meta.language: { dialects: string[]; family: string }` shape or something similar to that? If `dialects` is only ever meant to be used with `name`, I personally find it easier to reason about pushing them into an object together.

nzakas: The idea is that rules might only ever have `language` without `dialects` if they're narrowly focused.

Not opposed to revisiting elsewhere, I'd just like to stick with this to get started.

nzakas: https://github.com/eslint/eslint/issues/19462

Tanujkanti4441:
Not stale! waiting for a review.

nzakas:
@JoshuaKGoldberg did you intended to include updates to `check-rule-examples.js` in this PR?

JoshuaKGoldberg:
Ah, yes, and never mentioned this in the OP - editing it in now. Without the changes:

```plaintext
docs/src/rules/default-param-last.md
  56:4   error  Nonstandard language tag 'ts': use one of 'javascript', 'js' or 'jsx'
  59:20  error  Unexpected lint error found: Parsing error: Unexpected token :
  68:4   error  Nonstandard language tag 'ts': use one of 'javascript', 'js' or 'jsx'
  71:20  error  Unexpected lint error found: Parsing error: Unexpected token ?
```

nzakas (approved):
LGTM. Would like another review to double check the `check-rule-examples` changes.

docs/src/rules/default-param-last.md
```diff
@@ -41,6 +41,34 @@ Examples of **correct** code for this rule:
...
+This rule additionally supports TypeScript type syntax.
+
+Examples of **incorrect** TypeScript code for this rule:
+
+::: incorrect
+
+```ts
+/* eslint default-param-last: ["error"] */
+
+function h(a = 0, b: number) {}
```
mdjermanovic, line 59: Looks like the error location isn't underlined?

image

mdjermanovic: And perhaps we should remove the "Open in Playground" button because ts examples don't work in ESLint Playground?

JoshuaKGoldberg: 👍 filed https://github.com/eslint/eslint.org/issues/709 and added a check for `ts` language to skip adding the button. I'll add a root-level comment for the error location issue.

JoshuaKGoldberg:
Coming over from https://github.com/eslint/eslint/pull/19431/files#r1991932318: this PR is accumulating more and more TypeScript support changes. Would it be preferable to split them out once everything is working into their own PR?

I see #19529 was also sent with some of the same changes.

mdjermanovic:
> I think if we hide "Open in Playground" for now, that's enough to merge this. The playground changes may take a bit to get working and we can always iterate on the error location in a separate PR.

Agreed.

mdjermanovic (approved):
LGTM, thanks!

PR #19465, opened 2025-02-26
test: replace WebdriverIO with Cypress

### LINKED ISSUE #19234: Change Request: Switch browser test runner

Pixel998:
#### Prerequisites checklist

[ ] Documentation update
[ ] Add autofix to a rule
[ ] Add a CLI option
[ ] Add something to the core
[ ] Other, please explain:

Fix #19234 

Replace WebdriverIO with Cypress


fasttime:
Thanks for the new pull request @Pixel998. I can see that the browser test is still running quite slowly in CI. The [current run](https://github.com/eslint/eslint/actions/runs/13553247195/job/37881703412?pr=19465) took over 8 minutes. I was trying to use [cypress-io/github-action](https://github.com/cypress-io/github-action) to speed it up, but it didn't help. So I'm not sure. Either Cypress is just this lame in GitHub or we don't know how to set it up properly.

fasttime:
I think this is quite acceptable now. It's not as fast as WebdriverIO but still faster then Mocha unit tests and it never timed out in my test runs. Probably it isn't even necessary to use Firefox as a browser.

fasttime:
Note that [the output](https://github.com/eslint/eslint/actions/runs/13601439974/job/38027797382) of the Test step is showing two strange warnings:

> ```console
> Couldn't find tsconfig.json. tsconfig-paths will be skipped
> 
> tput: No value for $TERM and no -T specified
> ```

The first warning is filed as a bug in Cypress (cypress-io/cypress#21805 and cypress-io/cypress#22273), so maybe we could leave it as it is until they fix it. If we want to get rid of the warning, we could create a directory `cypress` with a TS config file `cypress/tsconfig.json` in it, move `cypress.config.js` to `cypress/config.js` and launch Cypress with the CLI option `--config-file=cypress/config.js`. That would remove the warning without adding an unnecessary `tsconfig.json` file at the repo root.

The second warning seems to only occur when `--no-runner-ui` is specified. It can be fixed by defining a `TERM` environment variable in `.github/workflows/ci.yml`:

```diff
  test_on_browser:
    name: Browser Test
    runs-on: ubuntu-latest
+   env:
+     TERM: xterm-256color
```

Clearly there are also other options.

package.json
```diff
@@ -138,20 +138,18 @@
...
-    "@wdio/cli": "^9.2.4",
-    "@wdio/concise-reporter": "^9.2.2",
-    "@wdio/mocha-framework": "^9.2.2",
     "babel-loader": "^8.0.5",
     "c8": "^7.12.0",
     "chai": "^4.0.1",
     "cheerio": "^0.22.0",
     "common-tags": "^1.8.0",
     "core-js": "^3.1.3",
+    "cypress": "^14.0.2",
```
fasttime, line 152:
```suggestion
    "cypress": "^14.1.0",
```

fasttime:
I think the output of Cypress is also easier to read. For comparison, this is what a test failure in GitHub Actions currently looks like with WebdriverIO — notice that there's no stack trace:

image

And this is the same failure with Cypress and the `progress` reporter, the same one we use for Mocha unit tests:

image

fasttime (approved):
LGTM, thanks! I would like another review from @nzakas.

mdjermanovic (approved):
LGTM, thanks!

nzakas (approved):
LGTM. I tested this locally and it has roughly the same speed as the WebDriver approach. Well done!

Just need to wait for the patch release window to close.

PR #19527, opened 2025-03-16
feat: add new options to class-methods-use-this

### LINKED ISSUE #19506: Rule Change: [class-methods-use-this] Add ignoreOverrideMethods option

class-methods-use-this

Generate fewer warnings

A new option

### Example code

```js
abstract class Base {
  abstract method(): void;
  abstract property: () => void;
}

class Derived extends Base {
  override method() {}
  override property = () => {};
}
```

Reports on the `method` and `property` for not using `this`.

This is sub-optimal because some projects prefer organizing code in class extensions even when derived methods don't always need to use `this`.

If the user has enabled a new `ignoreOverrideMethods` option, ignore methods and method-likes that are declared with the `override` keyword. No reports would be generated for that code.

### Participation

### Additional comments

Following #19498, this feature request corresponds to the existing typescript-eslint rule option: https://typescript-eslint.io/rules/class-methods-use-this/#ignoreoverridemethods

### LINKED ISSUE #19507: Rule Change: [class-methods-use-this] Add ignoreClassesThatImplementAnInterface option

class-methods-use-this

Generate fewer warnings

A new option

### Example code

```js
interface Base {
  method(): void;
}

class Derived implements Base {
  method() {}
  property = () => {};
}
```

Reports on the `method` and `property` for not using `this`.

This is sub-optimal because some projects prefer organizing code with classes implementing interfaces even when implementing methods don't always need to use `this`.

If the user has enabled a new `ignoreClassesThatImplementAnInterface` option to...

* `'all'`: ignore all classes that implement an interface
  * This is `true` in the typescript-eslint rule for legacy reasons, but IMO it's odd having a `boolean | string` option
* `'public-fields'`: ignore only the public fields of classes that implement an interface

Either way, the provided snippet would not receive any lint reports.

### Participation

### Additional comments

It would be nice to know which methods are actually defined on the base interface. But doing so would require typed linting, and not even the typescript-eslint extension rule is typed.

Following https://github.com/eslint/eslint/pull/19498, this feature request corresponds to the existing typescript-eslint rule option: https://typescript-eslint.io/rules/class-methods-use-this/#ignoreclassesthatimplementaninterface

sethamus:
#### Prerequisites checklist

[ ] Documentation update
[ ] Add autofix to a rule
[ ] Add a CLI option
[ ] Add something to the core
[ ] Other, please explain:

Added two new options to the `class-methods-use-this` rule:
* `ignoreOverrideMethods` ignores members that are marked with the `override` modifier.
* `ignoreClassesWithImplements` ignores class members that are defined within a class that `implements` an interface.

Closes #19506
Closes #19507


docs/src/rules/class-methods-use-this.md
```diff
@@ -106,10 +106,12 @@ class C {
 
 ## Options
 
-This rule has two options:
+This rule has four options:
 
 * `"exceptMethods"` allows specified method names to be ignored with this rule.
 * `"enforceForClassFields"` enforces that functions used as instance field initializers utilize `this`. (default: `true`)
+* `"ignoreOverrideMethods"` ignores members that are marked with the `override` modifier. (default: `false`)
+* `"ignoreClassesWithImplements"` ignores class members that are defined within a class that `implements` an interface.
```
nzakas, line 114: For clarity:
```suggestion
* `"ignoreOverrideMethods"` ignores members that are marked with the `override` modifier. (TypeScript only, default: `false`)
* `"ignoreClassesWithImplements"` ignores class members that are defined within a class that `implements` an interface. (TypeScript only)
```

lib/rules/class-methods-use-this.js
```diff
@@ -103,6 +114,20 @@ module.exports = {
...
+         * @param {ASTNode} node node to check
+         * @returns {boolean} True if parent class implements interfaces
+         * @private
+         */
+        function hasImplements(node) {
+            const classNode = node.parent.parent;
+            return classNode && 
+                   classNode.type === "ClassDeclaration" && 
+                   classNode.implements && 
+                   classNode.implements.length > 0;
```
nzakas, line 128: We can shorten this a bit
```suggestion
            return classNode?.type === "ClassDeclaration" &&
                   classNode.implements?.length > 0;
```

nzakas (commented):
This looks really good. Just left a couple notes to clean things up a bit.

nzakas:
Please double-check the lint errors.

docs/src/rules/class-methods-use-this.md
```diff
@@ -200,3 +202,126 @@ class A {
 ```
 
 :::
+
+### ignoreOverrideMethods
+
+```js
+"class-methods-use-this": [<enabled>, { "ignoreOverrideMethods": true | false }]
+```
```
snitin315, line 210: Do we need this block?

docs/src/rules/class-methods-use-this.md
```diff
@@ -200,3 +202,126 @@ class A {
...
+class Derived extends Base {
+    override foo() {};
+}
+```
+
+:::
+
+### ignoreClassesWithImplements
+
+```js
```
snitin315, line 260:
```suggestion
```ts
```

docs/src/rules/class-methods-use-this.md
```diff
@@ -200,3 +202,126 @@ class A {
 ```
 
 :::
+
+### ignoreOverrideMethods
+
+```js
```
snitin315, line 208:
```suggestion
```ts
```

lib/rules/class-methods-use-this.js
```diff
@@ -115,6 +138,24 @@ module.exports = {
...
+                    const implementsInterfaces = hasImplements(node);
+                    if (implementsInterfaces) {
+                        if (
+                            ignoreClassesWithImplements === "all" ||
+                            (ignoreClassesWithImplements === "public-fields" &&
+                                (!node.accessibility ||
+                                    node.accessibility === "public"))
+                        ) {
+                            return false;
+                        }
```
snitin315, line 155: private methods & properties can also start with `#`.
```js
class MyClass extends Base {
  #privateMethod() {
  }
}
```

docs/src/rules/class-methods-use-this.md
```diff
@@ -200,3 +202,126 @@ class A {
...
+Examples of **correct** TypeScript code for this rule with the `{ "ignoreOverrideMethods": true }` option:
+
+::: correct
+
+```ts
+/*eslint class-methods-use-this: ["error", { "ignoreOverrideMethods": true }] */
+
+class Derived extends Base {
+    override foo() {};
+}
```
snitin315, line 253: As mentioned in the issue description, it should work similarly to the [corresponding ts rule](https://typescript-eslint.io/rules/class-methods-use-this/#ignoreoverridemethods)

```suggestion
abstract class Base {
  abstract method(): void;
  abstract property: () => void;
}

class Derived extends Base {
  override method() {}
  override property = () => {};
}
```

docs/src/rules/class-methods-use-this.md
```diff
@@ -200,3 +202,126 @@ class A {
...
+Examples of **incorrect** TypeScript code for this rule with the `{ "ignoreClassesWithImplements": "all" }`:
+
+::: incorrect
+
+```ts
+/*eslint class-methods-use-this: ["error", { "ignoreClassesWithImplements": "all" }] */
+
+class A {
+    foo() {}
+}
```
snitin315, line 278:
```suggestion
class A {
    foo() {}
    property = () => {};
}
```

docs/src/rules/class-methods-use-this.md
```diff
@@ -200,3 +202,126 @@ class A {
...
+Examples of **correct** TypeScript code for this rule with the `{ "ignoreClassesWithImplements": "all" }` option:
+
+::: correct
+
+```ts
+/*eslint class-methods-use-this: ["error", { "ignoreClassesWithImplements": "all" }] */
+
+class Derived implements Base {
+    foo() {};
+}
```
snitin315, line 292:
```suggestion
interface Base {
  method(): void;
}

class Derived implements Base {
  method() {}
  property = () => {};
}
```

docs/src/rules/class-methods-use-this.md
```diff
@@ -200,3 +202,126 @@ class A {
...
+::: incorrect
+
+```ts
+/*eslint class-methods-use-this: ["error", { "ignoreClassesWithImplements": "public-fields" }] */
+
+class Derived implements Base {
+    foo() {}
+    private bar() {}
+    protected baz() {}
+}
```
snitin315, line 308:
```suggestion
interface Base {
  method(): void;
}

class Derived implements Base {
  method() {}
  property = () => {};

  private privateMethod() {}
  private privateProperty = () => {};

  protected protectedMethod() {}
  protected protectedProperty = () => {};
}
```

docs/src/rules/class-methods-use-this.md
```diff
@@ -200,3 +202,126 @@ class A {
...
+::: correct
+
+```ts
+/*eslint class-methods-use-this: ["error", { "ignoreClassesWithImplements": "public-fields" }] */
+
+class Derived implements Base {
+    foo() {}
+    bar() {}
+    baz() {}
+}
```
snitin315, line 324:
```suggestion
interface Base {
  method(): void;
}

class Derived implements Base {
  method() {}
  property = () => {};
}
```

tests/lib/rules/class-methods-use-this.js
```diff
@@ -244,8 +244,40 @@ ruleTesterTypeScript.run("class-methods-use-this", rule, {
         { code: "class A { #bar() {} }", options: [{ exceptMethods: ["#bar"] }] },
         { code: "class A { foo = function () {} }", options: [{ enforceForClassFields: false }] },
         { code: "class A { foo = () => {} }", options: [{ enforceForClassFields: false }] },
+        { code: "class A { override foo = () => {} }", options: [{ enforceForClassFields: false }] },
```
snitin315, line 247: Can we also add test cases for [privateIdentifiers using `#`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_properties)

snitin315 (changes requested):

JoshuaKGoldberg:

```shell
git merge -X ours main
```

docs/src/rules/class-methods-use-this.md
nzakas: I think @snitin315's suggestion is a good one. :+1:

docs/src/rules/class-methods-use-this.md
snitin315: This is not a javascript syntax 😄

docs/src/rules/class-methods-use-this.md
snitin315: Earlier we didn't supported the `ts` syntax, but now we do.

snitin315 (approved):
LGTM, thanks!

Leaving it open for 2nd review.

JoshuaKGoldberg (approved):
🚀

snitin315:
Will merge once we determine if a patch release is needed or not.

Issue #4192, opened 2015-10-19
no-magic-numbers: UnaryExpression + number is considered as error


ipanasenko:
Currently following code is considered as error: ```var foo = -42;```
I don't think it should do so

eslintbot:
Thanks for the issue! If you're reporting a bug, please be sure to include:

1. The version of ESLint you are using (run `eslint -v`)
2. What you did (the source code and ESLint configuration)
3. The actual ESLint output complete with numbers
4. What you expected to happen instead

Requesting a new rule? Please see [Proposing a New Rule](http://eslint.org/docs/developer-guide/contributing/new-rules) for instructions.

michaelficarra:
How is `-42` any less of a magic number than `42`?

ipanasenko:
It's magic, right.
But it should be allowed to declare it as a ```var/let/const```

```var foo = 42;``` is not error, why -42 is?

michaelficarra:
Oh, okay. Sounds good to me.

ilyavolodin:
Agree. Variable declaration should be excluded from this rule (that's the whole point of the rule).

alberto:
Working on this.

hzoo:
@alberto looks like @ipanasenko has a PR https://github.com/eslint/eslint/pull/4191 but isn't linked

alberto:
Oh, thanks @hzoo

PR #4191, opened 2015-10-19
Fix: no-magic-numbers: variable declarations (fixes #4192)

ipanasenko:
Currently following code is considered as error: ```var foo = -42;```
I don't think it should

PR #19561, opened 2025-03-22
feat: support TS syntax in `no-magic-numbers`

snitin315:
#### Prerequisites checklist

[x] Documentation update
[ ] Add autofix to a rule
[ ] Add a CLI option
[ ] Add something to the core
[ ] Other, please explain:

Refs #19173 

It adds support for:

- numeric literal types (type T = 1),
- enum members (enum Foo { bar = 1 }),
- readonly class properties (class Foo { readonly bar = 1 }).

Added 4 new options:

```ts
ignoreEnums?: boolean;
  ignoreNumericLiteralTypes?: boolean;
  ignoreReadonlyClassProperties?: boolean;
  ignoreTypeIndexes?: boolean;
```

### DIFF
--- a/lib/rules/no-magic-numbers.js
+++ b/lib/rules/no-magic-numbers.js
@@ -26,10 +26,139 @@ function normalizeIgnoreValue(x) {
 	return x;
 }
 
+/**
+ * Gets the true parent of the literal, handling prefixed numbers (-1 / +1)
+ * @param {ASTNode} node The literal node
+ * @returns {ASTNode|undefined} The parent node
+ */
+function getLiteralParent(node) {
+	if (
+		node.parent.type === "UnaryExpression" &&
+		["-", "+"].includes(node.parent.operator)
+	) {
+		return node.parent.parent;
+	}
+
+	return node.parent;
+}
+
+/**
+ * Checks if the node parent is a TypeScript enum member
+ * @param {ASTNode} node The node to be validated
+ * @returns {boolean} True if the node parent is a TypeScript enum member
+ */
+function isParentTSEnumDeclaration(node) {
+	const parent = getLiteralParent(node);
+	return parent?.type === "TSEnumMember";
+}
+
+/**
+ * Checks if the node parent is a TypeScript literal type
+ * @param {ASTNode} node The node to be validated
+ * @returns {boolean} True if the node parent is a TypeScript literal type
+ */
+function isParentTSLiteralType(node) {
+	return node.parent?.type === "TSLiteralType";
+}
+
+/**
+ * Checks if the node grandparent is a TypeScript type alias declaration
+ * @param {ASTNode} node The node to be validated
+ * @returns {boolean} True if the node grandparent is a TypeScript type alias declaration
+ */
+function isGrandparentTSTypeAliasDeclaration(node) {
+	return node.parent?.parent?.type === "TSTypeAliasDeclaration";
+}
+
+/**
+ * Checks if the node grandparent is a TypeScript union type and its parent is a type alias declaration
+ * @param {ASTNode} node The node to be validated
+ * @returns {boolean} True if conditions are met
+ */
+function isGrandparentTSUnionType(node) {
+	if (node.parent?.parent?.type === "TSUnionType") {
+		return isGrandparentTSTypeAliasDeclaration(node.parent);
+	}
+
+	return false;
+}
+
+/**
+ * Checks if the node is a valid TypeScript numeric literal type.
+ * @param {ASTNode} node The node to be validated
+ * @returns {boolean} True if the node is a TypeScript numeric literal type
+ */
+function isTSNumericLiteralType(node) {
+	// For negative numbers, use the parent node
+	if (
+		node.parent?.type === "UnaryExpression" &&
+		node.parent.operator === "-"
+	) {
+		node = node.parent;
+	}
+
+	// If the parent node is not a TSLiteralType, early return
+	if (!isParentTSLiteralType(node)) {
+		return false;
+	}
+
+	// If the grandparent is a TSTypeAliasDeclaration, ignore
+	if (isGrandparentTSTypeAliasDeclaration(node)) {
+		return true;
+	}
+
+	// If the grandparent is a TSUnionType and it's parent is a TSTypeAliasDeclaration, ignore
+	if (isGrandparentTSUnionType(node)) {
+		return true;
+	}
+
+	return false;
+}
+
+/**
+ * Checks if the node parent is a readonly class property
+ * @param {ASTNode} node The node to be validated
+ * @returns {boolean} True if the node parent is a readonly class property
+ */
+function isParentTSReadonlyPropertyDefinition(node) {
+	const parent = getLiteralParent(node);
+
+	if (parent?.type === "PropertyDefinition" && parent.readonly) {
+		return true;
+	}
+
+	return false;
+}
+
+/**
+ * Checks if the node is part of a type indexed access (eg. Foo[4])
+ * @param {ASTNode} node The node to be validated
+ * @returns {boolean} True if the node is part of an indexed access
+ */
+function isAncestorTSIndexedAccessType(node) {
+	// Handle unary expressions (eg. -4)
+	let ancestor = getLiteralParent(node);
+
+	/*
+	 * Go up another level while we're part of a type union (eg. 1 | 2) or
+	 * intersection (eg. 1 & 2)
+	 */
+	while (
+		ancestor?.parent?.type === "TSUnionType" ||
+		ancestor?.parent?.type === "TSIntersectionType"
+	) {
+		ancestor = ancestor.parent;
+	}
+
+	return ancestor?.parent?.type === "TSIndexedAccessType";
+}
+
 /** @type {import('../shared/types').Rule} */
 module.exports = {
 	meta: {
 		type: "suggestion",
+		dialects: ["typescript", "javascript"],
+		language: "javascript",
 
 		docs: {
 			description: "Disallow magic numbers",
@@ -75,6 +204,22 @@ module.exports = {
 						type: "boolean",
 						default: false,
 					},
+					ignoreEnums: {
+						type: "boolean",
+						default: false,
+					},
+					ignoreNumericLiteralTypes: {
+						type: "boolean",
+						default: false,
+					},
+					ignoreReadonlyClassProperties: {
+						type: "boolean",
+						default: false,
+					},
+					ignoreTypeIndexes: {
+						type: "boolean",
+						default: false,
+					},
 				},
 				additionalProperties: false,
 			},
@@ -93,8 +238,11 @@ module.exports = {
 			ignore = new Set((config.ignore || []).map(normalizeIgnoreValue)),
 			ignoreArrayIndexes = !!config.ignoreArrayIndexes,
 			ignoreDefaultValues = !!config.ignoreDefaultValues,
-			ignoreClassFieldInitialValues =
-				!!config.ignoreClassFieldInitialValues;
+			ignoreClassFieldInitialValues = !!config.ignoreClassFieldInitialValues,
+			ignoreEnums = !!config.ignoreEnums,
+			ignoreNumericLiteralTypes = !!config.ignoreNumericLiteralTypes,
+			ignoreReadonlyClassProperties = !!config.ignoreReadonlyClassProperties,
+			ignoreTypeIndexes = !!config.ignoreTypeIndexes;
 
 		const okTypes = detectObjects
 			? []
@@ -239,6 +387,10 @@ module.exports = {
 					(ignoreDefaultValues && isDefaultValue(fullNumberNode)) ||
 					(ignoreClassFieldInitialValues &&
 						isClassFieldInitialValue(fullNumberNode)) ||
+					(ignoreEnums && isParentTSEnumDeclaration(node)) ||
+					(ignoreNumericLiteralTypes && isTSNumericLiteralType(node)) ||
+					(ignoreTypeIndexes && isAncestorTSIndexedAccessType(node)) ||
+					(ignoreReadonlyClassProperties && isParentTSReadonlyPropertyDefinition(node)) ||
 					isParseIntRadix(fullNumberNode) ||
 					isJSXNumber(fullNumberNode) ||
 					(ignoreArrayIndexes && isArrayIndex(fullNumberNode, value))
--- a/docs/src/rules/no-magic-numbers.md
+++ b/docs/src/rules/no-magic-numbers.md
@@ -290,3 +290,131 @@ const dutyFreePrice = 100,
 ```
 
 :::
+
+### ignoreEnums (TypeScript only)
+
+Whether enums used in TypeScript are considered okay. `false` by default.
+
+Examples of **incorrect** code for the `{ "ignoreEnums": false }` option:
+
+::: incorrect
+
+```ts
+/*eslint no-magic-numbers: ["error", { "ignoreEnums": false }]*/
+
+enum foo {
+  SECOND = 1000,
+}
+```
+
+:::
+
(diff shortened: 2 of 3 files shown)


JoshuaKGoldberg (changes requested):
🔢

lib/rules/no-magic-numbers.js
```diff
@@ -26,10 +26,134 @@ function normalizeIgnoreValue(x) {
...
+		nodeToCheck = node.parent;
+	}
+
+	// If the parent node is not a TSLiteralType, early return
+	if (!isParentTSLiteralType(nodeToCheck)) {
+		return false;
+	}
+
+	// If the grandparent is a TSTypeAliasDeclaration, ignore
+	if (isGrandparentTSTypeAliasDeclaration(nodeToCheck)) {
```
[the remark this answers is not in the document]
snitin315: @JoshuaKGoldberg Fixed in https://github.com/eslint/eslint/pull/19561/commits/96c9ddf4540fae6e9a9db825765a14314d352174

Looks like a bug in [existing typescript-eslint rule as well](https://typescript-eslint.io/play/#ts=5.8.2&fileType=.ts&code=C4TwDgpgBAchDOwIBMoF4oAoDk2oB8tcCsBGASkoG4g&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQHYHsBaAWwEMBzASwGNDdZiAjRaVDSfAM08gBpxskAAIAXAJ4AHFNWiUJIwiniVcIgPQESFGnQbNW6MAG0B2KC2j5ofU9ixnBlcgWiIAcgxY0AMpREtSeAAVSRRDEThEWwgAX1sAXQE4mKA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false)

lib/rules/no-magic-numbers.js
```diff
@@ -75,6 +194,22 @@ module.exports = {
...
+						default: false,
+					},
+					ignoreReadonlyClassProperties: {
+						type: "boolean",
+						default: false,
+					},
+					ignoreTypeIndexes: {
+						type: "boolean",
+						default: false,
+					},
```
mdjermanovic, line 212: Can you also update types for this rule?

https://github.com/eslint/eslint/blob/865dbfed6cbade8a22756965be256da317801937/lib/types/rules.d.ts#L2821-L2842

lib/rules/no-magic-numbers.js
```diff
@@ -26,10 +26,129 @@ function normalizeIgnoreValue(x) {
...
+
+	// Use the more thorough check for deeply nested types
+	if (isInsideTSTypeAliasDeclaration(nodeToCheck)) {
+		return true;
+	}
+
+	// If the grandparent is a TSUnionType and it's parent is a TSTypeAliasDeclaration, ignore
+	if (isGrandparentTSUnionType(nodeToCheck)) {
+		return true;
+	}
```
mdjermanovic, line 103: This doesn't seem to have any effect, i.e., when I remove it, all the tests are still passing. Because `isInsideTSTypeAliasDeclaration(nodeToCheck)` above already checks if there's `TSTypeAliasDeclaration` anywhere in the ancestry.

lib/rules/no-magic-numbers.js
```diff
@@ -26,10 +26,129 @@ function normalizeIgnoreValue(x) {
...
+
+	while (current && current.parent) {
+		if (current.parent.type === "TSTypeAliasDeclaration") {
+			return true;
+		}
+		current = current.parent;
+	}
+
+	return false;
+}
```
mdjermanovic, line 68: If we allow numbers anywhere inside type alias declarations, that would differ from how `@typescript-eslint/no-magic-numbers` works. For example, it doesn't allow numbers in object types:

```ts
/* eslint @typescript-eslint/no-magic-numbers: ["error", { "ignoreNumericLiteralTypes": true }] */

type Foo = { bar: 42 }; // error
```

snitin315: Good catch, fixed.

lib/rules/no-magic-numbers.js
snitin315: Fixed.

lib/rules/no-magic-numbers.js
snitin315: Done.

lib/rules/no-magic-numbers.js
```diff
@@ -26,10 +26,129 @@ function normalizeIgnoreValue(x) {
...
+ * @param {ASTNode} node The literal node
+ * @returns {ASTNode} The parent node
+ */
+function getLiteralParent(node) {
+	if (["-", "+"].includes(node.parent.operator)) {
+		return node.parent.parent;
+	}
+
+	return node.parent;
+}
```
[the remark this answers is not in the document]
snitin315: So you mean if both the following are valid for TS:

```
class Foo {
  readonly D = 1;
  readonly F = +1;
}
```

The following should also be valid:

```js
/*eslint no-magic-numbers: ["error", { "ignoreArrayIndexes": true }]*/
a = data[+1]; // Currently reports error
a = data[1]; 
```

lib/rules/no-magic-numbers.js
mdjermanovic: I don't see changes in `eslint/lib/types/rules.d.ts` in this PR, maybe you forgot to push the commit?

lib/rules/no-magic-numbers.js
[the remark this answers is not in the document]
mdjermanovic: I think the rule should treat unary `+` consistently, one way or the other. Currently, the implementation in this PR works like this:

```ts
/* eslint no-magic-numbers: ["error", {
	ignoreReadonlyClassProperties: true,
	ignoreClassFieldInitialValues: true
}] */

class C {
    readonly foo = +42; // ok, because it's allowed by `ignoreReadonlyClassProperties: true`
    bar = +42; // not ok, even though `ignoreClassFieldInitialValues: true` is set
}

const MY_NUMBER = +42; // not ok, although it's a constant
```

So we could either:

1. Never treat unary `+` as part of the number, which would be consistent with how the core `no-magic-numbers` rule already works, but inconsistent with how the extra options in the `@typescript-eslint/no-magic-numbers` rule work.
2. Always treat unary `+` as part of the number. This would mean that some code currently reported by this rule, like `const MY_NUMBER = +42;`, would become valid.

I'd vote for 2., although the rule is frozen, because unary `+` is sometimes used to emphasize that the number is positive, and I think it doesn't make it more magical.

@eslint/eslint-tsc thoughts?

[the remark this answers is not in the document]
snitin315: I'd vote for 2 as well but would it be a breaking change for this rule?

lib/rules/no-magic-numbers.js
snitin315: Maybe so. Added in https://github.com/eslint/eslint/pull/19561/commits/d0b9b64bb8587ece2134308b63e1544cba2c1bee

lib/rules/no-magic-numbers.js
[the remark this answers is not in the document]
fasttime: It looks like the rule is inconsistent in handling unary plus vs. unary minus with certain options:

```js
/* eslint no-magic-numbers: ["error", { "ignoreDefaultValues": true }] */

const {
    foo = +42, // Only this number is reported
    bar = -42,
    baz = 42,
} = fooBarBaz;
```

[**Playground  Link**](https://eslint.org/play/#eyJ0ZXh0IjoiLyogZXNsaW50IG5vLW1hZ2ljLW51bWJlcnM6IFtcImVycm9yXCIsIHsgXCJpZ25vcmVEZWZhdWx0VmFsdWVzXCI6IHRydWUgfV0gKi9cblxuY29uc3Qge1xuICAgIGZvbyA9ICs0MixcbiAgICBiYXIgPSAtNDIsXG4gICAgYmF6ID0gNDIsXG59ID0gZm9vQmFyQmF6OyIsIm9wdGlvbnMiOnsicnVsZXMiOnt9LCJsYW5ndWFnZU9wdGlvbnMiOnsicGFyc2VyT3B0aW9ucyI6eyJlY21hRmVhdHVyZXMiOnt9fX19fQ==)

or

```js
/* eslint no-magic-numbers: ["error", { "ignoreClassFieldInitialValues": true }] */

class Test {
    foo = +42; // Only this number is reported
    bar = -42;
    baz = 42;
}
```

[**Playground  Link**](https://eslint.org/play/#eyJ0ZXh0IjoiLyogZXNsaW50IG5vLW1hZ2ljLW51bWJlcnM6IFtcImVycm9yXCIsIHsgXCJpZ25vcmVDbGFzc0ZpZWxkSW5pdGlhbFZhbHVlc1wiOiB0cnVlIH1dICovXG5cbmNsYXNzIFRlc3Qge1xuICAgIGZvbyA9ICs0MjtcbiAgICBiYXIgPSAtNDI7XG4gICAgYmF6ID0gNDI7XG59Iiwib3B0aW9ucyI6eyJydWxlcyI6e30sImxhbmd1YWdlT3B0aW9ucyI6eyJwYXJzZXJPcHRpb25zIjp7ImVjbWFGZWF0dXJlcyI6e319fX19)

With other options like `ignoreArrayIndexes`, a unary plus or minus isn't even considered:

```js
/*eslint no-magic-numbers: ["error", { "ignoreArrayIndexes": true }]*/

foo = a[42];  // no error
foo = a[+42]; // error
foo = a[-42]; // error
```

[**Playground  Link**](https://eslint.org/play/#eyJ0ZXh0IjoiLyplc2xpbnQgbm8tbWFnaWMtbnVtYmVyczogW1wiZXJyb3JcIiwgeyBcImlnbm9yZUFycmF5SW5kZXhlc1wiOiB0cnVlIH1dKi9cblxuZm9vID0gYVs0Ml07XG5mb28gPSBhWys0Ml07XG5mb28gPSBhWy00Ml07Iiwib3B0aW9ucyI6eyJydWxlcyI6e30sImxhbmd1YWdlT3B0aW9ucyI6eyJwYXJzZXJPcHRpb25zIjp7ImVjbWFGZWF0dXJlcyI6e319fX19)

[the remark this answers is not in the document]
mdjermanovic: Yes, the core rule didn't treat unary `+` as part of the number. On the other hand, unary `-` is, I believe, consistently treated as part of the number; negative numbers are just not considered valid array indexes.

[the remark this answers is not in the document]
fasttime: > On the other hand, unary `-` is, I believe, consistently treated as part of the number; negative numbers are just not considered valid array indexes.

I see now, thanks. `foo = a[-0]` is correctly ignored with `"ignoreArrayIndexes": true`.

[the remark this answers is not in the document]
fasttime: It looks like the unary `-` operator wasn't regarded as part of a number until #4191. This was changed following the discussion in #4192, but I don't see a mention of numbers with a unary `+` in the PR or in the issue. Maybe they were not considered because they are less common.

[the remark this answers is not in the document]
nzakas: I think it makes sense to treat unary + and - the same, which is to say, I vote for 2.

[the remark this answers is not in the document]
snitin315: Cool, will implement the 2nd option.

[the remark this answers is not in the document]
snitin315: Added in https://github.com/eslint/eslint/pull/19561/commits/1e855aa6493b1faa3762b2e63a807befdac51c4c

JoshuaKGoldberg (dismissed):
Super! 🚀

lib/rules/no-magic-numbers.js
```diff
@@ -26,10 +26,152 @@ function normalizeIgnoreValue(x) {
...
```
mdjermanovic, line 40: This looks like a duplicate logic now that `fullNumberNode` includes unary `+`. I think we can remove this function and pass `fullNumberNode` instead of `node` to `isParentTSEnumDeclaration()` and others.

snitin315: Yes, fixed in https://github.com/eslint/eslint/pull/19561/commits/b95eab9c3e16d50ff8f89625d6f6b3cf534fbfd6

lib/rules/no-magic-numbers.js
```diff
@@ -26,10 +26,136 @@ function normalizeIgnoreValue(x) {
...
+	return false;
+}
+
+/**
+ * Checks if the node is part of a type indexed access (eg. Foo[4])
+ * @param {ASTNode} node The node to be validated
+ * @returns {boolean} True if the node is part of an indexed access
+ */
+function isAncestorTSIndexedAccessType(node) {
+	// Handle unary expressions (eg. -4)
```
mdjermanovic, line 136:
```suggestion
```

I think this comment was referring to a now-removed code.

lib/rules/no-magic-numbers.js
```diff
@@ -26,10 +26,136 @@ function normalizeIgnoreValue(x) {
...
+ * @param {ASTNode} node The node to be validated
+ * @returns {boolean} True if conditions are met
+ */
+function isGrandparentTSUnionType(node) {
+	if (node.parent.parent.type === "TSUnionType") {
+		return isInsideTSTypeAliasDeclaration(node.parent.parent);
+	}
+
+	return false;
+}
```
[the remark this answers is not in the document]
snitin315: Yes, that's correct. Fixed.

lib/types/rules.d.ts
```diff
@@ -2850,6 +2850,22 @@ export interface ESLintRules extends Linter.RulesRecord {
...
+				 */
+				ignoreNumericLiteralTypes?: boolean;
+				/**
+				 * @default false
+				 */
+				ignoreReadonlyClassProperties?: boolean;
+				/**
+				 * @default false
+				 */
+				ignoreTypeIndexes?: boolean;
```
mdjermanovic, line 2868:
```suggestion
				ignoreEnums: boolean;
				/**
				 * @default false
				 */
				ignoreNumericLiteralTypes: boolean;
				/**
				 * @default false
				 */
				ignoreReadonlyClassProperties: boolean;
				/**
				 * @default false
				 */
				ignoreTypeIndexes: boolean;
```

For consistency with other options, since we're already using Partial<>.

lib/rules/no-magic-numbers.js
```diff
@@ -26,10 +26,73 @@ function normalizeIgnoreValue(x) {
...
+/**
+ * Checks if the node is a valid TypeScript numeric literal type.
+ * @param {ASTNode} node The node to be validated
+ * @returns {boolean} True if the node is a TypeScript numeric literal type
+ */
+function isTSNumericLiteralType(node) {
+	let ancestor = node.parent;
+
+	// Go up while we're part of a type union
+	while (ancestor.parent && ancestor.parent.type === "TSUnionType") {
```
mdjermanovic, line 47:
```suggestion
	while (ancestor.parent.type === "TSUnionType") {
```

It doesn't seem necessary to check if `ancestor.parent` exists here.

mdjermanovic (commented):
Just two small suggestions, then LGTM.

lib/types/rules.d.ts
```diff
@@ -2850,6 +2850,22 @@ export interface ESLintRules extends Linter.RulesRecord {
 				 * @default false
 				 */
 				detectObjects: boolean;
+				/**
+				 * @default false
+				 */
+				ignoreEnum: boolean;
```
mdjermanovic, line 2856:
```suggestion
				ignoreEnums: boolean;
```

snitin315:
Merging this since Josh had already approved it earlier and is unavailable until next week. Will create a follow-up PR if any further changes are needed.

PR #19562, opened 2025-03-22
feat: support TS syntax in `no-restricted-imports`

snitin315:
#### Prerequisites checklist

[ ] Documentation update
[ ] Add autofix to a rule
[ ] Add a CLI option
[ ] Add something to the core
[ ] Other, please explain:

Refers #19173 

It adds support for type import syntaxes:

- `import type X from "..."`
- `import { type X } from "..."`
- `import x = require("...")`

https://typescript-eslint.io/rules/no-restricted-imports/


lib/rules/no-restricted-imports.js
```diff
@@ -324,6 +339,56 @@ module.exports = {
...
+		}
+
+		/**
+		 * Check if a type import is allowed for a given path
+		 * @param {string} importSource The path of the import
+		 * @returns {boolean} Whether the type import is allowed
+		 */
+		function isAllowedTypeImportPath(importSource) {
+			return allowedTypeImportPathNameSet.has(importSource);
+		}
```
JoshuaKGoldberg, line 360: [Bug] I replaced the body of `isAllowedTypeImportPath` with `return true;` and all tests still passed. Which means the logic that fills in `allowedTypeImportPathNameSet` is effectively unused right now.

JoshuaKGoldberg (changes requested):
There seems to be either extra functionality or missing tests with `allowedTypeImportPathNameSet` / `isAllowedTypeImportPath`.

lib/rules/no-restricted-imports.js
snitin315: It is used, added a test case in https://github.com/eslint/eslint/pull/19562/commits/f9559eb7ce07895c289c1b7cb5946a3922efca82

We should add it in typescrip-eslint as well 😄

JoshuaKGoldberg: Haha, I'm even more excited about deleting the extension rules now...

lib/rules/no-restricted-imports.js
```diff
@@ -678,6 +779,35 @@ module.exports = {
 				}
 			},
 			ExportAllDeclaration: checkNode,
+			// Add support for TypeScript import equals declarations
+			TSImportEqualsDeclaration(node) {
+				if (node.moduleReference.type === "TSExternalModuleReference") {
+					const importSource = node.moduleReference.expression.value;
+
+					// Create a proper synthesized import declaration
+					const synthesizedImport = {
```
JoshuaKGoldberg, line 788: [Refactor] 🤔 Synthesizing nodes is a rather unusual practice. It's brittle to changes in AST and AST handling, as well as verbose. It was done in the extension rule because the core rule wasn't extensible.

For this PR I think it'd be cleaner to have `checkNode` handle these TS nodes natively. There's not much logic to change:

* `checkRestrictedPathAndReport` will need to also check this node type along with `ImportDeclaration`

...and that's about it that I can see? No unit tests fail locally when I make that change and pass `node` to `checkRestrictedPathAndReport` instead of `synthesizedImport`. The `node.moduleReference.type === "TSExternalModuleReference"` check is also able to be removed.

JoshuaKGoldberg (changes requested):
I think we can actually get rid of a bunch of this extension code! 🔪

lib/rules/no-restricted-imports.js
snitin315: Makes sense, done.

snitin315:
@JoshuaKGoldberg This needs a re-review.

lib/rules/no-restricted-imports.js
```diff
@@ -333,6 +398,25 @@ module.exports = {
...
+				isAllowedTypeImportPath(importSource)
+			) {
+				return;
+			}
+
+			// Skip if this is a type-only export and it's allowed
+			if (
+				node.type === "ExportNamedDeclaration" &&
+				isTypeOnlyExport(node) &&
+				isAllowedTypeImportPath(importSource)
```
JoshuaKGoldberg, line 415: [Testing] Removing this `&& isAllowedTypeImportPath(importSource)` didn't fail any unit tests. I added one locally that was a copy of the last test, `// Both regular and type imports should still be restricted`, but switched to exports instead of imports.

lib/rules/no-restricted-imports.js
```diff
@@ -452,6 +536,24 @@ module.exports = {
...
+				group.allowTypeImports
+			) {
+				return;
+			}
+
+			// Skip if this is a type-only export and it's allowed
+			if (
+				node.type === "ExportNamedDeclaration" &&
+				isTypeOnlyExport(node) &&
+				group.allowTypeImports
```
JoshuaKGoldberg, line 552: [Testing] Removing this `&& group.allowTypeImports` didn't fail any unit tests. I added one locally that was a copy of the test that fails if the earlier `&& group.allowTypeImports` is removed, `patterns: ["@typescript-eslint/utils/dist/*"],`, but switched to an export instead of an import.

lib/rules/no-restricted-imports.js
```diff
@@ -678,6 +780,20 @@ module.exports = {
 				}
 			},
 			ExportAllDeclaration: checkNode,
+			// Add support for TypeScript import equals declarations
+			TSImportEqualsDeclaration(node) {
+				if (node.moduleReference.type === "TSExternalModuleReference") {
```
JoshuaKGoldberg, line 785: [Testing] Removing this `if` doesn't fail any unit tests. This can be hit for namespace imports: `import B = A.B;`.

docs/src/rules/no-restricted-imports.md
```diff
@@ -282,6 +282,60 @@ import { AllowedObject } from "foo";
 
 :::
 
+#### allowTypeImports (TypeScript only)
+
+Whether to allow [Type-Only Imports](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-8.html#type-only-imports-and-export) for a path. Default: `false`.
```
JoshuaKGoldberg, line 287: [Docs] I'm only now realizing this, but I think there's an ambiguity in the existing typescript-eslint docs: does "type-only imports" also include "type-only exports"? Intuitively I don't think so, but all the docs and implementations say so. IIRC it's because an `export ... from` by nature includes an `import`... but that's not something every user would know.

Proposal: can we mention this in the docs here, to fix this long-standing potential confusion?

Maybe:

> This includes type-only `export`s, as they are equivalent to re-exporting an `import`.

JoshuaKGoldberg (changes requested):
The core functionality looks perfect to me. Just requesting changes on docs (sorry for not till now) and a bit more tests. 🚀

lib/rules/no-restricted-imports.js
snitin315: Added

docs/src/rules/no-restricted-imports.md
snitin315: Updated in https://github.com/eslint/eslint/pull/19562/commits/921496ade00f69b09dcc52130391da397350905d

mdjermanovic (commented):
Can you also update the types for this rule in `lib/types/rules.d.ts`?

lib/rules/no-restricted-imports.js
```diff
@@ -333,6 +398,25 @@ module.exports = {
...
 		function checkRestrictedPathAndReport(importSource, importNames, node) {
+			// Skip if this is a type-only import and it's allowed
+			if (
+				(node.type === "ImportDeclaration" ||
+					node.type === "TSImportEqualsDeclaration") &&
+				isTypeOnlyImport(node) &&
+				isAllowedTypeImportPath(importSource)
+			) {
+				return;
+			}
```
mdjermanovic, line 409: This logic might not be correct when there are multiple entries for the same import source. For example:

```ts
/* eslint no-restricted-imports: ["error", {
    paths: [
        {
            name: "mod",
            importNames: ["foo"],
            allowTypeImports: true,
            message: "import 'foo' only as type"
        },
        {
            name: "mod",
            importNames: ["bar"],
            message: "don't import 'bar' at all"
        }
    ]
}]*/

import type { bar } from "mod"; // ok, but should be error?

```

snitin315: Hmm, seems to be an existing bug in[ @typescript-eslint/no-restricted-imports rule](https://typescript-eslint.io/play/#ts=5.8.2&fileType=.ts&code=PQKgBApgzgNglgOwC5gAJIJ4AdoGMBOcWSAtNPMsAgPYn7RKG5IQAmJcAtltfklAC4wAbQBEEfPl6iANGADeAKDAqwWAIZIAFoJHLVBpQeOqE6zhCGjO1VrP0mDXHnwBy56ELEAzatVEAujIOjirqMDDUAO4AKtgQAJLcvPxCjACuEMGhBhZQUOoA5pZgos4pYADkvtSVYNQIMBhg6lBgmDiiIQYAvtmORjkqZhZWNnb9Q%2BVuHrpiAEbq%2BIGTOXkFxVasDZUo0yiVi-h1mi0RXaE9IQGKPQEgwIqK%2B%2B3xCmBHYD1g3lKcpeNRABuIA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQHYHsBaaFAF2gEsBjUxAE0IoFsAHfaU1DSfAM18gAacNkgABUgE8WKKpRalCKeBVykA9AWJlKNeo1btO6MAG1IiaNHZDMAXwC6IuyDtA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false), will fix.

snitin315: Fixed in https://github.com/eslint/eslint/pull/19562/commits/9521fb99949641da52c204a0c880980d657c2c7b

JoshuaKGoldberg (dismissed):
Looks great, nicely done! 👏

lib/rules/no-restricted-imports.js
```diff
@@ -344,6 +389,27 @@ module.exports = {
...
+							node.type === "TSImportEqualsDeclaration") &&
+						isTypeOnlyImport(node) &&
+						allowTypeImports
+					) {
+						return;
+					}
+
+					// Skip if this is a type-only export and it's allowed for this specific entry
+					if (
+						node.type === "ExportNamedDeclaration" &&
```
mdjermanovic, line 407: I'm not sure about the following case, since both rules report it although I think it should pass:

```js

/* eslint no-restricted-imports: ["error", {  
  "paths": [{
    "name": "foo",
    "allowTypeImports": true
  }]
}]
*/

/* eslint @typescript-eslint/no-restricted-imports: ["error", {  
  "paths": [{
    "name": "foo",
    "allowTypeImports": true
  }]
}]
*/

export type * from "foo"; // reported by both rules
``` 

It might be a false positive in `@typescript-eslint/no-restricted-imports`?

lib/rules/no-restricted-imports.js
```diff
@@ -344,6 +389,27 @@ module.exports = {
...
 					const allowedImportNames =
 						restrictedPathEntry.allowImportNames;
+					const allowTypeImports =
+						restrictedPathEntry.allowTypeImports;
+
+					// Skip if this is a type-only import and it's allowed for this specific entry
+					if (
+						(node.type === "ImportDeclaration" ||
+							node.type === "TSImportEqualsDeclaration") &&
+						isTypeOnlyImport(node) &&
```
mdjermanovic, line 399: Also not sure about this one, since both rules report it. Adding an allowed value import makes an otherwise allowed type import not allowed:

```js
/* eslint no-restricted-imports: ["error", {  
  "paths": [{
    "name": "foo",
    "importNames": ["bar"],
    "allowTypeImports": true
  }]
}]
*/

/* eslint @typescript-eslint/no-restricted-imports: ["error", {  
  "paths": [{
    "name": "foo",
    "importNames": ["bar"],
    "allowTypeImports": true
  }]
}]
*/

import { type bar } from "foo"; // this is valid

import { type bar, baz } from "foo"; // but this reports (both rules): 'bar' import from 'foo' is restricted

```

snitin315: Fixed in https://github.com/eslint/eslint/pull/19562/commits/a5adab3cd128abb48b1d892da5bb21a721e0b942

lib/rules/no-restricted-imports.js
```diff
@@ -446,11 +531,32 @@ module.exports = {
...
 		 * @param {Map<string,Object[]>} importNames Map of import names that are being imported
+		 * @param {string} importSource the import source string
 		 * @returns {void}
 		 * @private
 		 */
-		function reportPathForPatterns(node, group, importNames) {
-			const importSource = node.source.value.trim();
+		function reportPathForPatterns(node, group, importNames, importSource) {
+			// Use provided importSource or fallback to node.source.value for backward compatibility
+			const actualImportSource = importSource || node.source.value.trim();
```
mdjermanovic: I think this fallback isn't necessary as we're always passing `importSource`.

lib/rules/no-restricted-imports.js
```diff
@@ -678,6 +816,30 @@ module.exports = {
 				}
 			},
 			ExportAllDeclaration: checkNode,
+			// Add support for TypeScript import equals declarations
+			TSImportEqualsDeclaration(node) {
```
mdjermanovic, line 825: Can you add examples with this kind of import to the docs? I think it's important to note that they can only be restricted entirely, and that config objects that have either of `importNames`, `allowImportNames`, `importNamePattern`, or `allowImportNamePattern` won't restrict these imports.

lib/rules/no-restricted-imports.js
```diff
@@ -446,11 +531,32 @@ module.exports = {
...
+			const actualImportSource = importSource || node.source.value.trim();
+
+			// Skip if this is a type-only import and it's allowed
+			if (
+				node.type === "ImportDeclaration" &&
+				isTypeOnlyImport(node) &&
+				group.allowTypeImports
+			) {
+				return;
+			}
```
mdjermanovic, line 557: I think this would be a false positive:

```ts
/* eslint no-restricted-imports: ["error", {
  "patterns": [{
    "group": ["f*"],
	"allowTypeImports": true
  }]
}]
*/

import type fs = require("fs");

```

lib/rules/no-restricted-imports.js
```diff
@@ -400,6 +467,15 @@ module.exports = {
...
 						) {
 							specifiers.forEach(specifier => {
+								// Skip if this is a type-only import specifier and type imports are allowed
+								if (
+									allowTypeImports &&
+									specifier.specifier &&
+									specifier.specifier.importKind === "type"
+								) {
+									return;
+								}
```
mdjermanovic, line 488: This looks like a false positive:

```ts
/* eslint no-restricted-imports: ["error", {
  "paths": [{
    "name": "foo",
	"importNames": ["bar"]
	"allowTypeImports": true
  }]
}]
*/

export { type bar, baz } from "foo";

```

(the same for the other four places where only `specifier.specifier.importKind` is checked).

lumirlumir:
ping @snitin315

snitin315:
Apologies for the delay, will push an update & resolve the comments.

lib/rules/no-restricted-imports.js
snitin315: Added 👍🏻

lib/rules/no-restricted-imports.js
snitin315: Fixed in https://github.com/eslint/eslint/pull/19562/commits/59711edf410e6ebc6b4d3d729b6735b890addec1

lib/rules/no-restricted-imports.js
snitin315: Fixed in https://github.com/eslint/eslint/pull/19562/commits/59711edf410e6ebc6b4d3d729b6735b890addec1

docs/src/rules/no-restricted-imports.md
```diff
@@ -336,6 +336,53 @@ import type foo = require("import-foo");
 
 :::
 
+TypeScript [`import = require()` syntax](https://www.typescriptlang.org/docs/handbook/2/modules.html#es-module-syntax-with-commonjs-behavior) is supported, but it has limitations.
+
+You can only **fully restrict** these imports, you **cannot restrict** them based on specific import names like `importNames`, `allowImportNames`, `importNamePattern`, or `allowImportNamePattern` options.
```
nzakas: We don't use this much emphasis in the docs.
```suggestion
You can only fully restrict these imports, you cannot restrict them based on specific import names like `importNames`, `allowImportNames`, `importNamePattern`, or `allowImportNamePattern` options.
```

docs/src/rules/no-restricted-imports.md
```diff
@@ -336,6 +336,53 @@ import type foo = require("import-foo");
 
 :::
 
+TypeScript [`import = require()` syntax](https://www.typescriptlang.org/docs/handbook/2/modules.html#es-module-syntax-with-commonjs-behavior) is supported, but it has limitations.
```
nzakas: Please explain was "is supported" means here.

docs/src/rules/no-restricted-imports.md
```diff
@@ -336,6 +336,53 @@ import type foo = require("import-foo");
...
+/*eslint no-restricted-imports: ["error", { 
+  "paths": [{ "name": "disallowed-import" }] 
+}]*/
+
+import foo = require("disallowed-import");
+```
+
+:::
+
+**Note:** Import name restrictions do not apply to TypeScript import equals declarations. The following configuration will **not** restrict the import equals declaration:
```
nzakas:
```suggestion
**Note:** Import name restrictions do not apply to TypeScript import equals declarations. The following configuration will not restrict the import equals declaration:
```

nzakas (commented):
Added some notes on the docs.

docs/src/rules/no-restricted-imports.md
snitin315: Updated.

lib/rules/no-restricted-imports.js
```diff
@@ -446,11 +542,32 @@ module.exports = {
...
 		 * and any restricted import names that have been specified in the config
 		 * @param {Map<string,Object[]>} importNames Map of import names that are being imported
+		 * @param {string} importSource the import source string
 		 * @returns {void}
 		 * @private
 		 */
-		function reportPathForPatterns(node, group, importNames) {
-			const importSource = node.source.value.trim();
+		function reportPathForPatterns(node, group, importNames, importSource) {
+			const actualImportSource = importSource;
```
mdjermanovic: I think there's no need for another variable, we could just use `importSource` directly in the rest of the code of this function.

docs/src/rules/no-restricted-imports.md
```diff
@@ -282,6 +282,107 @@ import { AllowedObject } from "foo";
...
+    name: "import-foo",
+    allowTypeImports: true,
+    message: "Please use only type-only imports from 'import-foo'."
+}]}]*/
+
+import foo from 'import-foo';
+export { Foo } from 'import-foo';
+
+import baz from 'import-baz';
+export { Baz } from 'import-baz';
```
mdjermanovic:
```suggestion
```

This is correct code so it might be confusing why it's in incorrect examples?

docs/src/rules/no-restricted-imports.md
```diff
@@ -282,6 +282,107 @@ import { AllowedObject } from "foo";
...
+
+:::
+
+Examples of **correct** code for `allowTypeImports` in `paths`:
+
+::: correct { "sourceType": "module" }
+
+```ts
+/*eslint no-restricted-imports: ["error", { paths: [{
+    name: "foo",
```
mdjermanovic:
```suggestion
    name: "import-foo",
```

Otherwise, this would be valid even without `allowTypeImports: true`.

docs/src/rules/no-restricted-imports.md
```diff
@@ -282,6 +282,107 @@ import { AllowedObject } from "foo";
...
+    allowTypeImports: true,
+    message: "Please use only type-only imports from 'import-foo'."
+}]}]*/
+
+import { foo } from 'other-module';
```
mdjermanovic:
```suggestion
```

I think this isn't necessary for this example, and it might be better for understanding to just match the incorrect code example above (just with the `type` keyword added).

docs/src/rules/no-restricted-imports.md
```diff
@@ -282,6 +282,107 @@ import { AllowedObject } from "foo";
...
+
+import type foo from 'import-foo';
+export type { Foo } from 'import-foo';
+
+import type foo = require("import-foo");
+```
+
+:::
```
mdjermanovic, line 323: Can you also add a "correct" example where `type` is used with an imported name? For example:

```js
/*eslint no-restricted-imports: ["error", { paths: [{
    name: "import-foo",
	importNames: ["Baz"],
    allowTypeImports: true,
    message: "Please use 'Baz' from 'import-foo' as a type only."
}]}]*/

import { Bar, type Baz } from "import-foo";

```

docs/src/rules/no-restricted-imports.md
```diff
@@ -282,6 +282,107 @@ import { AllowedObject } from "foo";
...
+    "importNames": ["foo"]
+  }] 
+}]*/
+
+// This import equals declaration will NOT be restricted
+// even though it imports the entire module
+import foo = require("foo");
+```
+
+:::
```
mdjermanovic: This is in the `allowTypeImports` section, while it isn't directly related to it. Perhaps we could move it to Known Limitations?

docs/src/rules/no-restricted-imports.md
```diff
@@ -282,6 +282,107 @@ import { AllowedObject } from "foo";
 
 :::
 
+#### allowTypeImports (TypeScript only)
```
mdjermanovic, line 285: Should we also document this option for `patterns`?

docs/src/rules/no-restricted-imports.md
snitin315: Yes, added. Also found a false positive as well for same with patterns, figit xed that too.

```js
/*eslint no-restricted-imports: ["error", { patterns: [{
    group: ["import/private/*"],
    importNames: ["Baz"],
    allowTypeImports: true,
    message: "Please use 'Baz' from 'import/private/*' as a type only."
}]}]*/

import { Bar, type Baz } from "import/private/bar"; 
```

docs/src/rules/no-restricted-imports.md
```diff
@@ -282,6 +282,61 @@ import { AllowedObject } from "foo";
...
+```
+
+:::
+
+::: correct { "sourceType": "module" }
+
+```ts
+/*eslint no-restricted-imports: ["error", { paths: [{
+    name: "import-foo",
+	  importNames: ["Baz"],
```
mdjermanovic:
```suggestion
    importNames: ["Baz"],
```

To align indentation (there were mixed spaces and tabs on this line).

lib/types/rules.d.ts
```diff
@@ -41,8 +41,16 @@ interface NoRestrictedImportPathCommonOptions {
...
+	| {
+			importNames?: string[];
+			allowImportNames?: never;
+			allowTypeImports?: boolean;
+	  }
+	| {
+			allowImportNames?: string[];
+			importNames?: never;
+			allowTypeImports?: boolean;
+	  };
```
mdjermanovic: I think it would be simpler to add `allowTypeImports` to `NoRestrictedImportPathCommonOptions` instead of here. Also, I think we should update types for `patterns` (`NoRestrictedImportPatternCommonOptions`) as well.

lib/rules/no-restricted-imports.js
mdjermanovic: Just a reminder that this has not been addressed yet.

snitin315: Resolved now

lib/types/rules.d.ts
snitin315: Updated 👍🏻

nzakas:
@snitin315 can you check the CI failures?

lib/rules/no-restricted-imports.js
```diff
@@ -324,6 +339,48 @@ module.exports = {
...
+
+		/**
+		 * Check if the node is a type-only export
+		 * @param {ASTNode} node The node to check
+		 * @returns {boolean} Whether the node is a type-only export
+		 */
+		function isTypeOnlyExport(node) {
+			return (
+				node.exportKind === "type" ||
+				(node.specifiers?.length > 0 &&
```
mdjermanovic, line 377: I think there's no test that targets this case, so perhaps we should add one (e.g., `export {} from "mod"` should not pass with `allowTypeImports: true`).

lib/rules/no-restricted-imports.js
```diff
@@ -344,6 +401,28 @@ module.exports = {
...
 						restrictedPathEntry.allowImportNames;
+					const allowTypeImports =
+						restrictedPathEntry.allowTypeImports;
+
+					// Skip if this is a type-only import and it's allowed for this specific entry
+					if (
+						(node.type === "ImportDeclaration" ||
+							node.type === "TSImportEqualsDeclaration") &&
+						isTypeOnlyImport(node) &&
+						allowTypeImports
```
mdjermanovic:
```suggestion
						allowTypeImports &&
						(node.type === "ImportDeclaration" ||
							node.type === "TSImportEqualsDeclaration") &&
						isTypeOnlyImport(node)
```

It would be more efficient to check `allowTypeImports` first, as there's no need for calculations if it isn't enabled.

lib/rules/no-restricted-imports.js
```diff
@@ -344,6 +401,28 @@ module.exports = {
...
+					) {
+						return;
+					}
+
+					// Skip if this is a type-only export and it's allowed for this specific entry
+					if (
+						(node.type === "ExportNamedDeclaration" ||
+							node.type === "ExportAllDeclaration") &&
+						isTypeOnlyExport(node) &&
+						allowTypeImports
```
mdjermanovic:
```suggestion
						allowTypeImports &&
						(node.type === "ExportNamedDeclaration" ||
							node.type === "ExportAllDeclaration") &&
						isTypeOnlyExport(node)
```

Same as the previous comment.

lib/rules/no-restricted-imports.js
```diff
@@ -420,6 +508,14 @@ module.exports = {
 							!allowedImportNames.includes(importName)
 						) {
 							specifiers.forEach(specifier => {
+								// Skip if this is a type-only import specifier and type imports are allowed
+								if (
+									allowTypeImports &&
+									isTypeOnlySpecifier(specifier)
```
mdjermanovic:
```suggestion
									isTypeOnlySpecifier(specifier.specifier)
```

Seems that tests for this are missing as this wasn't caught.

lib/rules/no-restricted-imports.js
```diff
@@ -400,6 +479,15 @@ module.exports = {
 							restrictedImportNames.includes(importName)
 						) {
 							specifiers.forEach(specifier => {
+								// Skip if this is a type-only import specifier and type imports are allowed
+								if (
+									allowTypeImports &&
+									specifier.specifier &&
```
mdjermanovic:
```suggestion
```

This doesn't seem necessary?

lib/rules/no-restricted-imports.js
```diff
@@ -446,11 +542,30 @@ module.exports = {
...
 		 */
-		function reportPathForPatterns(node, group, importNames) {
-			const importSource = node.source.value.trim();
+		function reportPathForPatterns(node, group, importNames, importSource) {
+			// Skip if this is a type-only import and it's allowed
+			if (
+				(node.type === "ImportDeclaration" ||
+					node.type === "TSImportEqualsDeclaration") &&
+				isTypeOnlyImport(node) &&
+				group.allowTypeImports
```
mdjermanovic:
```suggestion
				group.allowTypeImports &&
				(node.type === "ImportDeclaration" ||
					node.type === "TSImportEqualsDeclaration") &&
				isTypeOnlyImport(node)
```

It would be more efficient to check `group.allowTypeImports` first, as there's no need for calculations if it isn't enabled.

lib/rules/no-restricted-imports.js
```diff
@@ -446,11 +542,30 @@ module.exports = {
...
+			) {
+				return;
+			}
+
+			// Skip if this is a type-only export and it's allowed
+			if (
+				(node.type === "ExportNamedDeclaration" ||
+					node.type === "ExportAllDeclaration") &&
+				isTypeOnlyExport(node) &&
+				group.allowTypeImports
```
mdjermanovic:
```suggestion
				group.allowTypeImports &&
				(node.type === "ExportNamedDeclaration" ||
					node.type === "ExportAllDeclaration") &&
				isTypeOnlyExport(node)
```

Same as the previous comment.

lib/rules/no-restricted-imports.js
```diff
@@ -573,6 +696,14 @@ module.exports = {
 					!allowedImportNames.includes(importName)
 				) {
 					specifiers.forEach(specifier => {
+						// Skip if this is a type-only import specifier and type imports are allowed
+						if (
+							group.allowTypeImports &&
+							isTypeOnlySpecifier(specifier.specifier)
+						) {
+							return;
+						}
```
mdjermanovic, line 704: Looks like tests for this case are missing as none fail if this is removed.

lib/rules/no-restricted-imports.js
```diff
@@ -592,6 +723,14 @@ module.exports = {
 					!allowedImportNamePattern.test(importName)
 				) {
 					specifiers.forEach(specifier => {
+						// Skip if this is a type-only import specifier and type imports are allowed
+						if (
+							group.allowTypeImports &&
+							isTypeOnlySpecifier(specifier.specifier)
+						) {
+							return;
+						}
```
mdjermanovic, line 731: Same as the previous comment, it looks like tests for this case are missing as none fail if this is removed.

lib/rules/no-restricted-imports.js
snitin315: Added this case in https://github.com/eslint/eslint/pull/19562/commits/460ebda4d689224c45dc43fb0a0e60dafbe00ef1

lib/rules/no-restricted-imports.js
snitin315: Added in https://github.com/eslint/eslint/pull/19562/commits/2709278147d762bb2fb80c009e535533255b3a56

lib/rules/no-restricted-imports.js
snitin315: Added in https://github.com/eslint/eslint/pull/19562/commits/2709278147d762bb2fb80c009e535533255b3a56

lib/rules/no-restricted-imports.js
snitin315: Added

lib/rules/no-restricted-imports.js
```diff
@@ -420,6 +507,11 @@ module.exports = {
 							!allowedImportNames.includes(importName)
 						) {
 							specifiers.forEach(specifier => {
+								// Skip if this is a type-only import specifier and type imports are allowed
+								if (allowTypeImports) {
+									return;
+								}
```
mdjermanovic:
```suggestion
								if (
									allowTypeImports &&
									isTypeOnlySpecifier(specifier.specifier)
								) {
									return;
								}
```

Looks like the `isTypeOnlySpecifier()` check has been removed instead of fixed (https://github.com/eslint/eslint/pull/19562#discussion_r2244807927).

For example, this should not pass:

```js
/* eslint no-restricted-imports: ["error", { paths: [{
    name: "foo",
	allowImportNames: ["bar"],
	allowTypeImports: true
}]}] */

import { baz } from "foo";

```

mdjermanovic:
Can you fix lint errors?

mdjermanovic (approved):
LGTM, thanks! Leaving open for a second review.

JoshuaKGoldberg (approved):
🔥 wonderful!

@mdjermanovic's thorough review makes me worried about edge cases existing in the typescript-eslint extension rule. I am relieved to see this PR land. 😄

PR #19563, opened 2025-03-22
feat: support TS syntax in `no-redeclare`

snitin315:
#### Prerequisites checklist

[ ] Documentation update
[ ] Add autofix to a rule
[ ] Add a CLI option
[ ] Add something to the core
[ ] Other, please explain:

Refs https://github.com/eslint/eslint/issues/19173

Adds TS support to `no-redeclare`. Added a new typescript only option similar to [typescript-eslint/no-redeclare](https://typescript-eslint.io/rules/no-redeclare/#ignoredeclarationmerge):

Examples of correct code with `{ ignoreDeclarationMerge: true }`:
```ts
interface A {
  prop1: 1;
}
interface A {
  prop2: 2;
}

namespace Foo {
  export const a = 1;
}
namespace Foo {
  export const b = 2;
}

class Bar {}
namespace Bar {}

function Baz() {}
namespace Baz {}
```
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved redeclaration checks across JavaScript and TypeScript code.
  * Reduced false positives for ambient declarations, valid namespace and declaration merges, classes, imports, and type-only declarations.
  * Improved detection of conflicting variables, duplicate functions or enums, incompatible value/type combinations, and unsupported overload combinations.
  * Added support for additional patterns, including namespaces, enums, globals, and destructuring.

* **Tests**
  * Expanded coverage with comprehensive TypeScript redeclaration scenarios and clearer validation of valid and invalid cases.


snitin315:
Tests are passing locally but failing on CI. I can reproduce the same behavior locally if I remove `languageOptions.sourceType: "script` option from `ruleTesterTypeScript` locally. I'm not sure why it's failing on CI.

tests/lib/rules/no-redeclare.js
```diff
@@ -716,3 +716,681 @@ ruleTester.run("no-redeclare", rule, {
 		},
 	],
 });
+
+const ruleTesterTypeScript = new RuleTester({
+	languageOptions: {
+		parser: require("@typescript-eslint/parser"),
+		sourceType: "script",
```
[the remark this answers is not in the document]
snitin315: Adding `parserOptions.sourceType` did fix the issue (https://github.com/eslint/eslint/pull/19563/commits/491d4036d127e5de76e0cd0a167848010eaf8178). @JoshuaKGoldberg I think the documentation needs to be updated, I can see it [defined in the code ](https://github.com/abrahamguo/typescript-eslint/blob/2224f12ff05b59c5a459291fe93fc8099221089d/packages/types/src/parser-options.ts#L113)

lib/rules/no-redeclare.js
```diff
@@ -65,8 +91,112 @@ module.exports = {
...
+			if (ignoreDeclarationMerge && identifiers.length > 1) {
+				// interfaces merging
+				if (
+					identifiers.every(
+						({ parent }) =>
+							parent.type === "TSInterfaceDeclaration",
+					)
+				) {
+					return;
+				}
```
JoshuaKGoldberg, line 112: [Refactor] This is handled by the `class + interface/namespace merging` logic since `CLASS_DECLARATION_MERGE_NODES` includes interfaces.

```suggestion
```

tests/lib/rules/no-redeclare.js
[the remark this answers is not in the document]
JoshuaKGoldberg: Heh, it's possible the usage in typescript-eslint code is an old holdover that's no longer necessary. Or not. https://github.com/typescript-eslint/typescript-eslint/issues/11016

JoshuaKGoldberg (changes requested):
I don't think I have enough context to weigh in on the `sourceType` points, but otherwise this looks very reasonable! Just one area of code I think can be deduplicated. 🚀

lib/rules/no-redeclare.js
snitin315: Done.

snitin315:
@JoshuaKGoldberg This needs a re-review.

JoshuaKGoldberg (changes requested):
I have to admit my confidence in declaration merging logic has never been super high - but I think there are some points of simplification we could go for?

lib/rules/no-redeclare.js
```diff
@@ -65,8 +91,103 @@ module.exports = {
...
+						CLASS_DECLARATION_MERGE_NODES.has(parent.type),
+					)
+				) {
+					const classDecls = identifiers.filter(
+						({ parent }) => parent.type === "ClassDeclaration",
+					);
+					if (classDecls.length === 1) {
+						// safe declaration merging
+						return;
+					}
```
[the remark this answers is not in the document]
snitin315: Hmm, looks like unnecessary code only. Removed.

JoshuaKGoldberg (dismissed):
🙌

lib/rules/no-redeclare.js
```diff
@@ -65,8 +91,91 @@ module.exports = {
...
+			// Handle TypeScript declaration merging if enabled
+			if (ignoreDeclarationMerge && identifiers.length > 1) {
+				// namespace/module merging
+				if (
+					identifiers.every(
+						({ parent }) => parent.type === "TSModuleDeclaration",
+					)
+				) {
+					return;
+				}
```
mdjermanovic, line 111: Doesn't https://github.com/eslint/eslint/pull/19563#discussion_r2021370869 apply here, too?

snitin315: It does, missed this.

lib/rules/no-redeclare.js
```diff
@@ -65,8 +91,91 @@ module.exports = {
...
+			const identifiers = variable.identifiers
+				.map(id => ({
+					identifier: id,
+					parent: id.parent,
+				}))
+				// ignore function declarations because TS will treat them as an overload
+				.filter(({ parent }) => parent.type !== "TSDeclareFunction");
+
+			// Handle TypeScript declaration merging if enabled
+			if (ignoreDeclarationMerge && identifiers.length > 1) {
```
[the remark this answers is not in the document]
snitin315: Hmm, that looks like a bug to me because we do r[eport error when a variable & namespace have same name](https://typescript-eslint.io/play/#ts=5.8.2&fileType=.ts&code=JYOwLgpgTgZghgYwgAgILIN4ChnIA5QD2eAjAFzIkDcWAvllqJLIiutrgcQEwXc30sCQiADOYNMgC8yAEQBrAFZLZNIA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQHYHsBaaRAE0QGN4BDE9KfAM0cgBpxtIABAFwE8ADigrQAlgJ6EU8Ubh4B6AsTKUadDAG0O2KImjR80NtuxYdnUQHMCJACKra1HqPy4AsnsuJ6POIhMQAL4mALocwYFAA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false).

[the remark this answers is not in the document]
snitin315: But the [following code also reports no error](https://eslint.org/play/#eyJ0ZXh0IjoiLyplc2xpbnQgbm8tcmVkZWNsYXJlOiBcImVycm9yXCIqLyBcblxuLyogZ2xvYmFsIEJhciAqL1xuXG52YXIgQmFyID0gMzsiLCJvcHRpb25zIjp7InJ1bGVzIjp7fSwibGFuZ3VhZ2VPcHRpb25zIjp7InBhcnNlck9wdGlvbnMiOnsiZWNtYUZlYXR1cmVzIjp7fX19fX0=):

```js
/*eslint no-redeclare: "error"*/ 

// should it be error? 'Bar' is already defined by a variable declaration.
/* global Bar */ 

var Bar = 3;
```

snitin315:
I think we can also remove `impliedStrict` from ecma features in case of typescriptEslintParser. [As per the docs](https://typescript-eslint.io/packages/parser/#configuration), only `jsx` & `globalReturn` seems to be valid.

Screenshot 2025-04-30 at 6 16 50 AM

nzakas:
@snitin315 please double-check the lint failure.

@JoshuaKGoldberg @mdjermanovic @fasttime this PR has been auto-closed for inactivity multiple times. What do you want to do here?

lib/rules/no-redeclare.js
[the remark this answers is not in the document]
snitin315: @mdjermanovic Fixed. The new behavior is as follows:

```js
/* eslint no-redeclare: ["error", { "ignoreDeclarationMerge": true }] */

namespace Foo { // <- error: 'Foo' is already defined as a built-in global variable.
	export const a = 1;
}

namespace Foo { // <- error: 'Foo' is already defined as a built-in global variable.
	export const b = 2;
}

/* global Bar */  // <- error: 'Bar' is already defined by a variable declaration.

namespace Bar {
	export const a = 1;
}

namespace Bar {
	export const b = 1;
}
```

fasttime:
Sorry for losing track of this. I will review it soon.

docs/src/rules/no-redeclare.md
fasttime, line 1: I think the rule description should be updated:

https://github.com/eslint/eslint/blob/c03442bbb2aef1021f28fb5ad344a65afd5ad551/docs/src/rules/no-redeclare.md?plain=1#L12-L16

Note that only _JavaScript_ and _variables_ are mentioned, but the rule would now report other types of declarations as well.

tests/lib/rules/no-redeclare.js
```diff
@@ -716,3 +716,975 @@ ruleTester.run("no-redeclare", rule, {
...
+			options: [{ ignoreDeclarationMerge: true }],
+		},
+		{
+			code: `
+  function A() {}
+  namespace A {}
+  namespace A {}
+		`,
+			options: [{ ignoreDeclarationMerge: true }],
+		},
```
fasttime: These test cases include the option `ignoreDeclarationMerge`, but if the option is removed, the tests will still pass. This is intentional, and it's how the typescript-eslint rule works ([Playground link](https://typescript-eslint.io/play/#ts=5.8.3&fileType=.ts&code=HYQwtgpgzgDiDGEAEBBJBvJBfAUKSsCyamWQA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQHYHsBaaRAE0QGN4BDE9KfAM0cgBpxtIABAFwE8ADigrQAlgJ6EU8Ubh4B6AsTKUadDAG0O2KImjR80SNrABdDgF8QFoA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false)):

https://github.com/eslint/eslint/blob/c03442bbb2aef1021f28fb5ad344a65afd5ad551/lib/rules/no-redeclare.js#L112-L118

If we are going to recreate the typescript-eslint functionality without re-evaluating it, I'd suggest removing the option from the test cases and the code examples where it isn't necessary, and clarifying the documentation.

fasttime (commented):
The typescript-eslint rule `no-redeclare`, upon which this PR is based, is not recommended for new projects, because it reports problems that are already reported by the TypeScript compiler, see https://typescript-eslint.io/rules/no-redeclare/. The built-in `no-redeclare` rule on the other hand _is_ recommended, and it reports some problems that are not reported by tsc, like `var` redeclarations.

I'm not sure if we want to add the unrecommended TypeScript extensions to this rule without an option.

nzakas:
If the typescript-eslint rule is not recommended for us, maybe we just shouldn't update this rule?

fasttime:
> If the typescript-eslint rule is not recommended for us, maybe we just shouldn't update this rule?

I'm not sure. It seems that all new problems the rule would find will be also reported by TypeScript, provided that the file is tested.

nzakas:
> I'm not sure. It seems that all new problems the rule would find will be also reported by TypeScript, provided that the file is tested.

Right. If updating this rule just adds noise that is already caught by TypeScript, I'm not sure it's worth it.

fasttime:
Then maybe we should update just the documentation and rule metadata, and keep only the new tests that ensure the rule works as intended with TypeScript files?

nzakas:
> Then maybe we should update just the documentation and rule metadata, and keep only the new tests that ensure the rule works as intended with TypeScript files?

Makes sense to me. 👍

tests/lib/rules/no-redeclare.js
```diff
@@ -716,3 +716,706 @@ ruleTester.run("no-redeclare", rule, {
...
+						id: "NodeListOf",
+					},
+					messageId: "redeclaredAsBuiltin",
+				},
+			],
+		},
+		{
+			code: `
+	  interface A {}
+	  interface A {}
```
fasttime: This is not considered an error in TypeScript and is not reported by `@typescript-eslint/no-redeclare`.

tests/lib/rules/no-redeclare.js
```diff
@@ -716,3 +716,706 @@ ruleTester.run("no-redeclare", rule, {
...
+					},
+					line: 3,
+					messageId: "redeclared",
+				},
+			],
+		},
+		{
+			code: `
+	  interface A {}
+	  class A {}
```
fasttime: This is also valid.

tests/lib/rules/no-redeclare.js
```diff
@@ -716,3 +716,706 @@ ruleTester.run("no-redeclare", rule, {
...
+					},
+					line: 3,
+					messageId: "redeclared",
+				},
+			],
+		},
+		{
+			code: `
+	  class A {}
+	  namespace A {}
```
fasttime: Also valid.

tests/lib/rules/no-redeclare.js
```diff
@@ -716,3 +716,706 @@ ruleTester.run("no-redeclare", rule, {
...
+					line: 3,
+					messageId: "redeclared",
+				},
+			],
+		},
+		{
+			code: `
+	  interface A {}
+	  class A {}
+	  namespace A {}
```
fasttime: Also valid.

tests/lib/rules/no-redeclare.js
```diff
@@ -716,3 +716,706 @@ ruleTester.run("no-redeclare", rule, {
...
+					},
+					line: 4,
+					messageId: "redeclared",
+				},
+			],
+		},
+		{
+			code: `
+	  function A() {}
+	  namespace A {}
```
fasttime: Also valid.

fasttime (commented):
I've commented on some invalid cases which are actually considered valid TypeScript code and are also treated as valid by the typescript-eslint rule. Initially, I didn't realize our rule would report those usages as invalid without modification. To better align with user expectations, I think it would be good to ignore TS type-only declarations (the ones that are stripped out from code during compilation) like `type`, `interface`, `namespace`, `declare function`, etc. These are already handled by TypeScript and are safe to redeclare if the declarations are compatible.

tests/lib/rules/no-redeclare.js
snitin315: Hmm, we do need some modifications to handle these.

snitin315:
@fasttime I've addressed the feedback. PTAL

tests/lib/rules/no-redeclare.js
```diff
@@ -716,3 +716,640 @@ ruleTester.run("no-redeclare", rule, {
...
+					messageId: "redeclaredBySyntax",
+					type: "Block",
+				},
+			],
+		},
+
+		{
+			code: `
+	  type T = 1;
+	  type T = 2;
```
fasttime: I think this case can be also considered valid because it's reported by the TypeScript compiler:

[TypeScript Playground](https://www.typescriptlang.org/play/?#code/C4TwDgpgBAglC8UCMBuAUKSsHJUA)

tests/lib/rules/no-redeclare.js
```diff
@@ -716,3 +716,640 @@ ruleTester.run("no-redeclare", rule, {
...
+		},
+		{
+			code: `
+	    var Foo;
+		namespace Foo {
+			export const a = 2;
+		}
+		namespace Foo {
+			export const b = 4;
+		}
```
fasttime: The combination `var` + `namespace` is also checked by TypeScript, so we don't need to report it:

[TypeScript Playground](https://www.typescriptlang.org/play/?ssl=4&ssc=2&pln=1&pc=1#code/G4QwTgBAYg9jDcAoAdiAtgUwM4AcQGMNo4IBvRASAwA8cYwAXCfGZLJkCAXggCYkAvkA)

tests/lib/rules/no-redeclare.js
```diff
@@ -716,3 +716,640 @@ ruleTester.run("no-redeclare", rule, {
...
+					},
+					line: 4,
+					messageId: "redeclared",
+				},
+			],
+		},
+		{
+			code: `
+	  type something = string;
+	  const something = 2;
```
fasttime: I'm not sure about this case. TypeScript considers it valid code since the two declarations don't collide: the first one is a type, the second one is value. `@typescript-eslint/no-redeclare` on the other hand flags this case as invalid because the declarations use the same identifier:

[typescript-eslint Playground](https://typescript-eslint.io/play/#ts=5.8.3&fileType=.ts&code=C4TwDgpgBAzg9gWwsAFgSwHYHMoF5bABOmWA3AFADGcGMwsiy62eUATKUA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6Jge1ukQBNEy8AIY90YANrhsORNGgdokKRAC6UgL4h1QA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false)

I'm not sure if the combination of a value and type with the same name should be forbidden.

tests/lib/rules/no-redeclare.js
```diff
@@ -716,3 +716,640 @@ ruleTester.run("no-redeclare", rule, {
...
+			namespace Bar {
+				export const a = 1;
+			}
+			`,
+			languageOptions: {
+				globals: {
+					Foo: "readonly",
+					Bar: "readonly",
+				},
+			},
```
fasttime: Similar to the previous case, I'm also not sure about this one. It seems that the `global` language option declares `Foo` and `Bar` as global _values_, but `namespace` doesn't redeclare these values, it declares types. I'm not sure if we want to forbid the combination of value and type where both have the same identifier.

tests/lib/rules/no-redeclare.js
```diff
@@ -716,3 +716,640 @@ ruleTester.run("no-redeclare", rule, {
...
+			code: `
+			/* global Foo */
+			namespace Foo {
+				export const a = 1;
+			}
+
+			/* global Bar */
+			namespace Bar {
+				export const a = 1;
+			}
```
fasttime: As above, I'm not sure if the combination of a value and type with the same name should be forbidden.

tests/lib/rules/no-redeclare.js
```diff
@@ -716,3 +716,640 @@ ruleTester.run("no-redeclare", rule, {
...
+			`,
+
+			languageOptions: {
+				globals: {
+					Foo: "readonly",
+				},
+			},
```
fasttime, line 1458: As above, I'm not sure if the combination of a value and type with the same name should be forbidden.

fasttime (commented):
Thanks for the update. There's a particular scenario I'm not sure about. What should happen when a value and a type share the same name? For example:

```ts
const a = 42; // value
type a = string; // type
```

or

```ts
let b: object; // value
interface b { foo: string } // type
```

or also

```ts
/* global c */
type c = number;
```

The TypeScript compiler considers these valid code because there is no collision between values and types even if they share the same name:

[TypeScript Playground](https://www.typescriptlang.org/play/?#code/MYewdgzgLgBAhjAvDALAJgNwCgoE8AOApvEjNAE4CWYA5tlgDaGwBGAXDCCwFaHBTZqUQuQBmcYMRYwA3jFEgQHCtRowAvliA)

fasttime:
@eslint/eslint-tsc I could use some feedback on https://github.com/eslint/eslint/pull/19563#pullrequestreview-3036151420.

nzakas:
I think we should follow the typescript-eslint behavior, which is to flag this as a redeclaration:
https://typescript-eslint.io/play/#ts=5.8.2&fileType=.ts&code=LAKAxg9gdgzgLgAgIYILwICwCYDcCD0%2BCAbkgDYCuApqHAJ4AOVyaC8ATgJZQDmehCekyA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQHYHsBaaRAE0QGN4BDE9KfAM0cgBpxtIABAFwE8ADigrQAlgJ6EU8Ubh4B6AsTKUadDJETRo%2BaJA4BfEAaA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false

tests/lib/rules/no-redeclare.js
fasttime, line 1: These cases should be valid:

```ts
declare function a(): void;
declare function a(): void;

declare function b(): void;
declare class b { }

enum c { }
namespace c { }

declare enum d { }
interface e { }
```

typescript-eslint doesn't report any violations: [typescript-eslint Playground](https://typescript-eslint.io/play/#ts=5.8.2&fileType=.ts&code=CYUwxgNghgTiAEAzArgOzAFwJYHtXygAoBKALngDcctgBuAKFElgRXWzwJPKpocfDQ4SNJlz4ARt0rU6A5sOYBnJfAnwA3vAC%2B9eiFTIAtvDCad9VFCMglAByhgEZrbvlCEB4-GDndWVAwQGERHTz8gA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQHYHsBaaRAE0QGN4BDE9KfAM0cgBpxtIABAFwE8ADigrQAlgJ6EU8Ubh4B6AsTKUadDAG0O2KImjR80NtrABdDgF8QFoA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false)

tests/lib/rules/no-redeclare.js
```diff
@@ -716,3 +716,640 @@ ruleTester.run("no-redeclare", rule, {
...
+						id: "NodeListOf",
+					},
+					messageId: "redeclaredAsBuiltin",
+				},
+			],
+		},
+		{
+			code: `
+	  function A() {}
+	  class A {}
```
fasttime: There should be no need to report this case since it's a syntax error in JavaScript and it doesn't compile in TypeScript: [typescript-eslint Playground](https://typescript-eslint.io/play/#ts=5.8.2&fileType=.ts&code=GYVwdgxgLglg9mABAQQBQEpEG8C%2BAoCAGwEMBnUlbHIA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQHYHsBaaRAE0QGN4BDE9KfAM0cgBpxtIABAFwE8ADigrQAlgJ6EU8Ubh4B6AsTKUadDAG0O2KImjR80SNrABdDgF8QFoA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false).

tests/lib/rules/no-redeclare.js
```diff
@@ -716,3 +716,640 @@ ruleTester.run("no-redeclare", rule, {
...
+					line: 3,
+					messageId: "redeclared",
+				},
+			],
+		},
+		{
+			code: `
+	  function A() {}
+	  class A {}
+	  namespace A {}
```
fasttime: As above, this is incorrect code, so the first error seems superfluous.

lib/rules/no-redeclare.js
```diff
@@ -84,6 +86,129 @@ module.exports = {
...
+
+			// Function + namespace (no interface or class)
+			if (
+				functions === 1 &&
+				namespaces >= 1 &&
+				interfaces === 0 &&
+				classes === 0
+			) {
+				return true;
+			}
```
fasttime: With this logic, the rule is making explicit exceptions for certain combinations of declaration types that should not be reported.

I think it would be easier to do the opposite: start by allowing all combinations, and report an error only for specific cases, for example when `counts.others > 0`, `counts.functions > 1`, and probably not much more.

nzakas:

kirkwaiblinger:
> The typescript-eslint rule `no-redeclare`, upon which this PR is based, is not recommended for new projects, because it reports problems that are already reported by the TypeScript compiler, see https://typescript-eslint.io/rules/no-redeclare/. The built-in `no-redeclare` rule on the other hand _is_ recommended, and it reports some problems that are not reported by tsc, like `var` redeclarations.

FYI (in case you didn't come across this already), this has been a topic of recent discussion on the typescript-eslint side... https://github.com/typescript-eslint/typescript-eslint/issues/11461.

nzakas:
@kirkwaiblinger it would be helpful if you could post a small summary here so we can know what it is you're referring us to.

kirkwaiblinger:
> @kirkwaiblinger it would be helpful if you could post a small summary here so we can know what it is you're referring us to.

This is fair! 

I meant to flag that 

- the docs at `@typescript-eslint/no-redeclare` have some flaws/contradictions and should not be taken too literally at the moment. 
- in particular, the recommendation against using `@typescript-eslint/no-redeclare` in new projects is disputed, and I think likely will be revised, though we don't have 100% alignment yet. 
- the extension rule's unconditional flagging of type-space vs value-space name collisions is controversial, and not using the extension rule at all is legitimate/recommended in order to disable that report. So, if that feature is brought to the base rule, likely an option makes sense.

I haven't followed too closely, but I think that this is the direction the PR was going already; just thought I'd flag that the typescript-eslint info that is referred to in the discussion here may evolve over time.

fasttime:
> * the extension rule's unconditional flagging of type-space vs value-space name collisions is controversial, and not using the extension rule at all is legitimate/recommended in order to disable that report. So, if that feature is brought to the base rule, likely an option makes sense.

This is an interesting suggestion. Our current stance is that a value and a type sharing the same name should be reported as a problem, because that is what typescript-eslint is doing ([Playground link](https://typescript-eslint.io/play/#ts=5.8.2&fileType=.ts&code=LAKAxg9gdgzgLgAgIYILwICwCYDcCD0%2BCAbkgDYCuApqHAJ4AOVyaC8ATgJZQDmehCekyA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQHYHsBaaRAE0QGN4BDE9KfAM0cgBpxtIABAFwE8ADigrQAlgJ6EU8Ubh4B6AsTKUadDJETRo%2BaJA4BfEAaA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false)). Introducing an option to control this behavior could be useful, though it would also add complexity. Thoughts on this, @eslint/eslint-tsc?

On a related note, I noticed that typescript-eslint/typescript-eslint#11461 is accepting PRs. Perhaps it might be worth coordinating efforts to avoid diverging implementations.

fasttime:
This issue was discussed in the 2025-11-13 TSC Meeting and the resolution was that name collisions between values and types should not be reported by the rule.

fasttime:
I was planning to resume reviewing this pull request after the last turbulent days, and I've now gone through the above discussions. To summarize, and to make sure we are aligned, my understanding of the current consensus for implementing the changes to this rule is as follows (with higher entries having higher priority):

1. The rule should maintain the current behavior for JS sources.
2. It should not report name collisions between values and types (https://github.com/eslint/eslint/pull/19563#issuecomment-3541769265).
3. It should not report redeclarations of types (since these are either already handled by the TypeScript compiler or are legitimate) (https://github.com/eslint/eslint/pull/19563#issuecomment-2984403842).
4. It should otherwise report problems reported by `@typescript-eslint/no-redeclare`.

Let me know in case I missed something. (And thanks for the last changes @snitin315).

tests/lib/rules/no-redeclare.js
```diff
@@ -739,3 +739,592 @@ ruleTester.run("no-redeclare", rule, {
...
+					},
+					messageId: "redeclaredBySyntax",
+				},
+			],
+		},
+
+		{
+			code: `
+	  type T = 1;
+	  type T = 2;
```
fasttime: This case should be valid. It's already reported by TypeScript ([Playground](https://www.typescriptlang.org/play/?#code/C4TwDgpgBAglC8UCMBuAUKSsHJUA)).

tests/lib/rules/no-redeclare.js
```diff
@@ -739,3 +739,592 @@ ruleTester.run("no-redeclare", rule, {
...
+		},
+		{
+			code: `
+	    var Foo;
+		namespace Foo {
+			export const a = 2;
+		}
+		namespace Foo {
+			export const b = 4;
+		}
```
fasttime: This case is also valid: `var` declares a value, whereas `namespace` declares a type, hence there is no name collision. The redeclaration of the type `Foo` is not a concern, since TypeScript allows it.

tests/lib/rules/no-redeclare.js
```diff
@@ -739,3 +739,592 @@ ruleTester.run("no-redeclare", rule, {
...
+					},
+					messageId: "redeclared",
+				},
+			],
+		},
+
+		// Notifications of readonly are moved from no-undef: https://github.com/eslint/eslint/issues/4504
+		{
+			code: "/*global b:false*/ var b = 1;",
+			options: [{ builtinGlobals: true }],
```
fasttime: This case isn't reported in JavaScript ([Playground](https://eslint.org/play/#eyJ0ZXh0IjoiLyplc2xpbnQgbm8tcmVkZWNsYXJlOiBbXCJlcnJvclwiLCB7IFwiYnVpbHRpbkdsb2JhbHNcIjogdHJ1ZSB9XSovXG5cbi8qZ2xvYmFsIGI6ZmFsc2UqLyB2YXIgYiA9IDE7Iiwib3B0aW9ucyI6eyJydWxlcyI6e30sImxhbmd1YWdlT3B0aW9ucyI6eyJwYXJzZXJPcHRpb25zIjp7ImVjbWFGZWF0dXJlcyI6e319fX19)). I don't see why it should be reported in TypeScript.

tests/lib/rules/no-redeclare.js
```diff
@@ -739,3 +739,592 @@ ruleTester.run("no-redeclare", rule, {
...
+					data: {
+						id: "a",
+					},
+					messageId: "redeclared",
+				},
+			],
+		},
+		{
+			code: "var Object = 0;",
+			options: [{ builtinGlobals: true }],
```
fasttime: This case is valid in JavaScript. I think it should remain valid in TypeScript ([Playground](https://eslint.org/play/#eyJ0ZXh0IjoiLyplc2xpbnQgbm8tcmVkZWNsYXJlOiBbXCJlcnJvclwiLCB7IFwiYnVpbHRpbkdsb2JhbHNcIjogdHJ1ZSB9XSovXG5cbnZhciBPYmplY3QgPSAwO1xuIiwib3B0aW9ucyI6eyJydWxlcyI6e30sImxhbmd1YWdlT3B0aW9ucyI6eyJwYXJzZXJPcHRpb25zIjp7ImVjbWFGZWF0dXJlcyI6e319fX19)).

lib/rules/no-redeclare.js
fasttime, line 1: Per https://github.com/eslint/eslint/pull/19563#issuecomment-3687365917, I think the current changes in this file aren't needed. The underlying issue is simply that `iterateDeclarations()` should only report value declarations.

Type-only declarations aren't expected to conflict with value declarations and they are already handled by the TypeScript compiler, so we don't need to address them here. A minimal fix would look something like this:

```diff
 			}
 
 			for (const id of variable.identifiers) {
-				yield { type: "syntax", node: id, loc: id.loc };
+				// Intentionally omit class declarations and import declarations because they can't
+				// be redeclared in JavaScript, and they are handled by the compiler in TypeScript.
+				if (
+					[
+						"ArrayPattern",
+						"ArrowFunctionExpression",
+						"AssignmentPattern",
+						"FunctionDeclaration",
+						"FunctionExpression",
+						"Property",
+						"RestElement",
+						"TSEnumDeclaration",
+						"VariableDeclarator",
+					].includes(id.parent.type) &&
+					!id.parent.declare
+				) {
+					yield { type: "syntax", node: id, loc: id.loc };
+				}
 			}
 
 			if (variable.eslintExplicitGlobalComments) {
```

tests/lib/rules/no-redeclare.js
```diff
@@ -739,3 +739,592 @@ ruleTester.run("no-redeclare", rule, {
...
+			/* global Foo */
+			namespace Foo {
+				export const a = 1;
+			}
+
+			/* global Bar */
+			namespace Bar {
+				export const a = 1;
+			}
+			`,
```
fasttime: Same here: this case should be valid, because a `global` comment declares values, whereas `namespace` declares a type.

tests/lib/rules/no-redeclare.js
```diff
@@ -739,3 +739,592 @@ ruleTester.run("no-redeclare", rule, {
...
+
+			languageOptions: {
+				globals: {
+					Foo: "readonly",
+				},
+			},
```
fasttime, line 1458: This case should be valid since `namespace` only declares types.

tests/lib/rules/no-redeclare.js
```diff
@@ -739,3 +739,592 @@ ruleTester.run("no-redeclare", rule, {
...
+			code: `
+	  type NodeListOf = 1;
+			`,
+			options: [{ builtinGlobals: true }],
+			languageOptions: {
+				parserOptions: {
+					lib: ["dom"],
+					sourceType: "script",
+				},
+			},
```
fasttime, line 1418: This case is already reported by TypeScript with the specified parser options, so it should be valid, I think. The TypeScript Playground has the `"dom"` library included by default: https://www.typescriptlang.org/play/?#code/C4TwDgpgBAcg9gEwgGQJYGdgHkBmUC8UAjANwBQQA.

fasttime (commented):
I've added some new comments. I hope we've addressed all outstanding questions and can now move this PR forward in the right direction.

tests/lib/rules/no-redeclare.js
kirkwaiblinger: Unless I misunderstand what is meant, `namespace` can also create values though too...

When only declaring types, no JS is emitted, eg
```ts
namespace Foo {
   export interface Bar { 
   }
}
```
but if at least one runtime value is declared, a JS object is emitted, eg
```ts
namespace Foo {
   export interface Bar { 
   }
   
   export const Baz = 1;
}
```


Wasn't able to find a better reference but here's the general docs on `namespace`: https://www.typescriptlang.org/docs/handbook/namespaces.html

Link to a demo in the TS playground: https://www.typescriptlang.org/play/?target=99#code/HYQwtgpgzgDiDGEAEBVYBLYUAuJjfRGwgBMkBvAKCSQgA8YB7AJ2yU2OYDMFkAxRowqVqSAL6UJlUJFi8kASSy58hYmSo16TVkniNlSAEIhmSALxIAjAG5JQA

fasttime: You are correct. This case should stay invalid, and in the general case, I guess the rule needs to distinguish between type-only and value namespaces. I checked the parser output, but there doesn't seem to be any indication of whether a namespace is actually instantiated as a value. That means the check would have to rely on recursively inspecting the namespace's contents, for example:

```ts
namespace Foo {
    namespace Bar {
        namespace Baz {
            // export const qux = 1; // Uncomment to make outer namespace a value
        }
    }
}
```

tests/lib/rules/no-redeclare.js
snitin315: Resolved.

lib/rules/no-redeclare.js
snitin315: Updated.

tests/lib/rules/no-redeclare.js
snitin315: Updated in https://github.com/eslint/eslint/pull/19563/changes/0973449745a9921d9a311c693f5346271e5d6cc4

fasttime (commented):
Based on https://github.com/eslint/eslint/pull/19563/changes/BASE..a730ed90506e0d95dad4477a1597c2b7a4cb4d3f#r2709907914, redeclarations of instantiated namespaces should be invalid:

```ts
namespace Foo {
	const a = 2;
}
namespace Foo {
	const b = 2;
}
```

**Update:** on second thought, I think this case doesn't match any of the rules we defined (https://github.com/eslint/eslint/pull/19563#issuecomment-3687365917). At this point, I don't feel strongly on reporting this or not, as long as we understand the behavior and are able to explain it.

lib/rules/no-redeclare.js
```diff
@@ -84,6 +106,272 @@ module.exports = {
...
+
+			const hasNonMergeable = validNodes.some(
+				node => !mergeableTypes.has(node.type),
+			);
+
+			// If there are non-mergeable declarations, not valid
+			if (hasNonMergeable) {
+				return false;
+			}
```
fasttime, line 255: Maybe I'm missing something, but this check seems unnecessary?

fasttime (commented):
We should update the logic to report on enum redeclarations, as these are flagged by typescript-eslint, e.g.:

```ts
enum Foo {}
enum Foo {}
```

[**Playground link**](https://typescript-eslint.io/play/#ts=5.9.2&fileType=.ts&code=KYOwrgtgBAYg9nKBvAvgKFJWDnqA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQHYHsBaaRAE0QGN4BDE9KfAM0cgBpxtIABAFwE8ADigrQAlgJ6EU8Ubh4B6AsTKUadDAG0O2KImjR80NtuxYdnUQHMCJACKra1HqPy4AsnsuJ6POIhMQAL4mALocwYFAA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false)

fasttime (commented):
We should probably not report declarations that use the `declare` keyword, e.g.:

```ts
declare var A: string;
declare var A: string;
```

Although typescript-eslint reports this case, TypeScript does not produce any values when the code is compiled.

[**TypeScript Playground**](https://www.typescriptlang.org/play/?#code/CYUwxgNghgTiAEA3W8CCAueBnALjAlgHYDmA3ALABQoksCyMamuBJpQA)

JoshuaKGoldberg:
👋 @snitin315 are you still working on this PR? This and #19812 are the last two open PRs for #19173, it'd be nice to close that out.

snitin315:
Yes, apologies for the delay. I'll push updates today

lib/rules/no-redeclare.js
snitin315: It's necessary,  it's the only thing catching destructuring redeclarations in plain JS. [Playground](https://eslint.org/play/#eyJ0ZXh0IjoiLyplc2xpbnQgbm8tcmVkZWNsYXJlOiBcImVycm9yXCIqL1xuXG52YXIgW2FdID0gZm9vOyB2YXIgW2FdID0gYmFyOyIsIm9wdGlvbnMiOnsicnVsZXMiOnt9LCJsYW5ndWFnZU9wdGlvbnMiOnsicGFyc2VyT3B0aW9ucyI6eyJlY21hRmVhdHVyZXMiOnsianN4Ijp0cnVlfX0sInBhcnNlciI6IkB0eXBlc2NyaXB0LWVzbGludC9wYXJzZXIiLCJzb3VyY2VUeXBlIjoibW9kdWxlIn19fQ==)

```js
/*eslint no-redeclare: "error"*/

var [a] = foo; var [a] = bar; // currently reported; without check, it doesn't report this.
```

snitin315: Added more tests in https://github.com/eslint/eslint/pull/19563/commits/f5664beaf333238a34a25f6d06d8b128688ba55c

snitin315:
>Update: on second thought, I think this case doesn't match any of the rules we defined (#19563 (comment)). At this point, I don't feel strongly on reporting this or not, as long as we understand the behavior and are able to explain it.

**On namespaces creating values:** agreed, and the rule already accounts for it — it checks whether a namespace is *instantiated* rather than treating `namespace` as type-only:

```ts
var Foo;
namespace Foo {
	const a = 1; // reported — instantiated
}

var Foo;
namespace Foo {
	export type T = string; // not reported — nothing emitted
}
```

**On `namespace` + `namespace`:** doesn't follow from rules 1–3, but I think rule 4 covers it. `@typescript-eslint/no-redeclare` defaults to `ignoreDeclarationMerge: true`, whose ignore list includes `namespace + namespace` (instantiated or not). That also seems right on the merits — merging two instantiated namespaces is the intended use of namespace merging and `tsc` accepts it. Left unreported.

fasttime:
I'm going to close and reopen this PR to verify that the ecosystem tests are passing.

lib/rules/no-redeclare.js
```diff
@@ -84,6 +124,284 @@ module.exports = {
...
+			).length;
+
+			if (typeAliasCount > 1) {
+				return false;
+			}
+
+			/*
+			 * Multiple TSEnumDeclarations are not allowed. An enum merges with a
+			 * namespace, but not with another enum.
+			 */
```
fasttime, line 278: Actually enum redeclarations do merge together. The rule disallows them only to maintain the typescript-eslint behavior.

[**TS Playground**](https://www.typescriptlang.org/play/?#code/KYOwrgtgBAYg9nKBvAglAvFAjAGigIQygCYBfAbgChLRJYFkBhI4vAESIGYLKg)

[**typescript-eslint Playground**](https://typescript-eslint.io/play/#ts=6.0.2&fileType=.tsx&code=PQKgBApgzgNglgOwC5gAJIJ4AdoGMBOcWSAtNPMsAgPYn4QAmEuMAhvQFyT77X5ghgAWABQoiAgCuAWzAAxatTABvMAEEwAXjABGADRgAQlrAAmMAF8A3KPFTZCpaoDCJ0wYAiJgMyWbIoA&eslintrc=N4KABGBEBOCuA2BTAzpAXGYBfEWg&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false)

lib/rules/no-redeclare.js
```diff
@@ -84,6 +124,284 @@ module.exports = {
...
+			}
+
+			// Multiple TSTypeAliasDeclarations are not allowed (type aliases don't merge)
+			const typeAliasCount = validNodes.filter(node =>
+				isNodeType(node, "TSTypeAliasDeclaration"),
+			).length;
+
+			if (typeAliasCount > 1) {
+				return false;
+			}
```
fasttime: This check isn't needed I think, since we decided not to report type-only redeclarations.

lib/rules/no-redeclare.js
```diff
@@ -84,6 +124,284 @@ module.exports = {
...
+			 */
+			const mergeableTypes = new Set([
+				"TSInterfaceDeclaration",
+				"ClassDeclaration",
+				"TSModuleDeclaration",
+				"FunctionDeclaration",
+				"TSDeclareFunction",
+				"TSEnumDeclaration",
+				"VariableDeclarator",
+				"TSTypeAliasDeclaration",
```
fasttime: This can be removed, I guess, since type-only-redeclarations aren't checked anyway.
```suggestion
```

lib/rules/no-redeclare.js
```diff
@@ -84,6 +124,284 @@ module.exports = {
...
+			 * If there are null nodes (builtins/comments), only allow if all
+			 * remaining nodes are type-only (interfaces, type aliases, or
+			 * non-instantiated namespaces) since they don't conflict with values.
+			 */
+			if (validNodes.length !== nodes.length) {
+				return (
+					validNodes.length > 0 &&
+					validNodes.every(
+						node =>
+							typeOnlyDeclarations.has(node.type) ||
```
fasttime: Same here, we aren't reporting type-only redeclarations.

```suggestion
```

lib/rules/no-redeclare.js
```diff
@@ -84,6 +124,284 @@ module.exports = {
...
+			).length;
+
+			if (enumCount > 1) {
+				return false;
+			}
+
+			// Separate nodes by namespace, treating non-instantiated namespaces as type-only
+			const typeOnlyNodes = validNodes.filter(
+				node =>
+					typeOnlyDeclarations.has(node.type) ||
```
fasttime: Same here.

```suggestion
```

lib/rules/no-redeclare.js
```diff
@@ -84,6 +124,284 @@ module.exports = {
...
+					(node.type === "TSModuleDeclaration" &&
+						!isNamespaceInstantiated(node)),
+			);
+			const valueOnlyNodes = validNodes.filter(node =>
+				valueOnlyDeclarations.has(node.type),
+			);
+			const bothNamespaceNodes = validNodes.filter(
+				node =>
+					!typeOnlyDeclarations.has(node.type) &&
```
fasttime: Same here.

```suggestion
```

fasttime (commented):
Thanks for the update @snitin315, happy to see this PR moving forward. I've left some comments. Please, have a look.

lib/rules/no-redeclare.js
snitin315: Updated comment

snitin315:
@fasttime Thanks for the review, I've addressed your feedback

lib/rules/no-redeclare.js
```diff
@@ -84,6 +124,263 @@ module.exports = {
...
+
+					if (!allAreTSDeclareFunction) {
+						return false;
+					}
+				}
+				/*
+				 * Single value-only + type-only declarations are OK
+				 * Multiple TSDeclareFunction + type-only declarations are OK
+				 */
+				return true;
```
fasttime: I think by the current logic this is just asking whether there is at most one value node:

```suggestion
				return valueOnlyNodes.length <= 1;
```

lib/rules/no-redeclare.js
```diff
@@ -84,6 +124,263 @@ module.exports = {
...
+		function isValidTSDeclarationCombination(nodes) {
+			if (nodes.length < 2) {
+				return false;
+			}
+
+			/**
+			 * Mergeable declaration types in TypeScript.
+			 */
+			const mergeableTypes = new Set([
+				"TSInterfaceDeclaration",
```
fasttime: Interfaces are type-only, so it should be fine to disregard them.

```suggestion
```

lib/rules/no-redeclare.js
```diff
@@ -84,6 +124,263 @@ module.exports = {
...
+				valueOnlyDeclarations.has(node.type),
+			);
+			const bothNamespaceNodes = validNodes.filter(
+				node =>
+					!valueOnlyDeclarations.has(node.type) &&
+					!(
+						node.type === "TSModuleDeclaration" &&
+						!isNamespaceInstantiated(node)
+					),
+			);
```
fasttime, line 300: It looks like these are just the instantiated namespaces?

```suggestion
			const bothNamespaceNodes = validNodes.filter(
				node =>
					node.type === "TSModuleDeclaration" &&
					isNamespaceInstantiated(node),
			);
```

fasttime (commented):
Thanks for the update @snitin315. A bunch of outdated code from previous revisions here that's taking time to review. I left a few more comments.

snitin315:
@fasttime Thanks! Just addressed your comments. PTAL

PR #19565, opened 2025-03-23
feat: support TS syntax in `no-shadow`

snitin315:
#### Prerequisites checklist

[ ] Documentation update
[ ] Add autofix to a rule
[ ] Add a CLI option
[ ] Add something to the core
[ ] Other, please explain:

### DIFF
--- a/lib/rules/no-shadow.js
+++ b/lib/rules/no-shadow.js
@@ -24,6 +24,24 @@ const FOR_IN_OF_TYPE = /^For(?:In|Of)Statement$/u;
 const SENTINEL_TYPE =
 	/^(?:(?:Function|Class)(?:Declaration|Expression)|ArrowFunctionExpression|CatchClause|ImportDeclaration|ExportNamedDeclaration)$/u;
 
+// TS-specific node types
+const FUNCTIONS_HOISTED_NODES = new Set(["FunctionDeclaration"]);
+const TYPES_HOISTED_NODES = new Set([
+	"TSInterfaceDeclaration",
+	"TSTypeAliasDeclaration",
+]);
+
+// TS-specific function variable def types
+const ALLOWED_FUNCTION_VARIABLE_DEF_TYPES = new Set([
+	"TSCallSignatureDeclaration",
+	"TSFunctionType",
+	"TSMethodSignature",
+	"TSEmptyBodyFunctionExpression",
+	"TSDeclareFunction",
+	"TSConstructSignatureDeclaration",
+	"TSConstructorType",
+]);
+
 //------------------------------------------------------------------------------
 // Rule Definition
 //------------------------------------------------------------------------------
@@ -32,6 +50,8 @@ const SENTINEL_TYPE =
 module.exports = {
 	meta: {
 		type: "suggestion",
+		dialects: ["typescript", "javascript"],
+		language: "javascript",
 
 		defaultOptions: [
 			{
@@ -39,6 +59,8 @@ module.exports = {
 				builtinGlobals: false,
 				hoist: "functions",
 				ignoreOnInitialization: false,
+				ignoreTypeValueShadow: true,
+				ignoreFunctionTypeParameterNameValueShadow: true,
 			},
 		],
 
@@ -54,7 +76,7 @@ module.exports = {
 				type: "object",
 				properties: {
 					builtinGlobals: { type: "boolean" },
-					hoist: { enum: ["all", "functions", "never"] },
+					hoist: { enum: ["all", "functions", "never", "types", "functions-and-types"] },
 					allow: {
 						type: "array",
 						items: {
@@ -62,6 +84,8 @@ module.exports = {
 						},
 					},
 					ignoreOnInitialization: { type: "boolean" },
+					ignoreTypeValueShadow: { type: "boolean" },
+					ignoreFunctionTypeParameterNameValueShadow: { type: "boolean" },
 				},
 				additionalProperties: false,
 			},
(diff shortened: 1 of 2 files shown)


JoshuaKGoldberg (changes requested):
Very exciting to see this nuanced rule's TS support shape up! I found a good collection of logical checks that don't impact unit tests when removed. Could you please take a look at them & the rest of the added logic in the PR? For each, either:

* It's a valid case a user might hit, so it should be unit tested
* It's not technically reachable, so it's "dead" code that can be removed safely

I left suggestions through around half the file but don't intend them as requests - just as notices of some of the code that doesn't fail tests when removed right now.

Thanks!

lib/rules/no-shadow.js
```diff
@@ -75,10 +109,196 @@ module.exports = {
...
+			variable,
+			shadowedVariable,
+		) {
+			if (ignoreFunctionTypeParameterNameValueShadow !== true) {
+				return false;
+			}
+
+			if (!("isValueVariable" in variable)) {
+				return false;
+			}
```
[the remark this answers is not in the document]
snitin315: Removing this is causing 6 test cases to fail

lib/rules/no-shadow.js
```diff
@@ -295,12 +539,107 @@ module.exports = {
...
 
+		/**
+		 * Checks if the initialization of a variable has the declare modifier in a
+		 * definition file.
+		 * @param {Object} variable The variable to check
+		 * @returns {boolean} Whether the variable is declared in a definition file
+		 */
+		function isDeclareInDTSFile(variable) {
+			const fileName = context.filename;
+			if (!fileName.endsWith(".d.ts")) {
```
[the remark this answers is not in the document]
snitin315: Ah! good catch. Fixed in https://github.com/eslint/eslint/pull/19565/commits/60fc8a3a479dce39c8fa74efc21aa397ba1c8b0a

lib/rules/no-shadow.js
```diff
@@ -75,10 +109,176 @@ module.exports = {
...
+		 * Checks if it's a function type parameter shadow
+		 * @param {Object} variable The variable to check
+		 * @returns {boolean} Whether it's a function type parameter shadow case to ignore
+		 */
+		function isFunctionTypeParameterNameValueShadow(variable) {
+			if (ignoreFunctionTypeParameterNameValueShadow !== true) {
+				return false;
+			}
+
+			return variable.defs.every(def =>
```
JoshuaKGoldberg, line 191: [Refactor] 🤔 Unit tests don't fail if we bail out early, and for the life of me I cannot figure out a test case to show why we shouldn't do this.

```suggestion
			return variable.defs.some(def =>
```

lib/rules/no-shadow.js
```diff
@@ -75,10 +109,176 @@ module.exports = {
...
+				functionExpr.type !== "FunctionExpression" &&
+				functionExpr.type !== "TSEmptyBodyFunctionExpression"
+			) {
+				return false;
+			}
+			const methodDefinition = functionExpr.parent;
+			if (methodDefinition.type !== "MethodDefinition") {
+				return false;
+			}
+			return methodDefinition.static;
```
JoshuaKGoldberg, line 233: [Refactor] Almost all of the `return`s in this function can be removed without failing unit tests 🤯.  

```suggestion
			const typeParameter = variable.identifiers[0].parent;
			const typeParameterDecl = typeParameter.parent;
			if (typeParameterDecl.type !== "TSTypeParameterDeclaration") {
				return false;
			}
			const functionExpr = typeParameterDecl.parent;
			const methodDefinition = functionExpr.parent;
			return methodDefinition.static;
```

Deep dive for fun: this comes from a couple long-standing pain in types representing scopes:
* Getters can't be used as type predicates (https://github.com/microsoft/TypeScript/issues/43368). Getters like [`get isTypeVariable(): boolean`](https://github.com/microsoft/TypeScript/issues/43368) therefore don't narrow down the types of variables scopes
* There aren't specific types like `TypeVariable extends Scope.Variable` to narrow down to anyway

I suspect the [`// this shouldn't happen...`](https://github.com/typescript-eslint/typescript-eslint/blob/884995d8da0249578d79e457705fc6cbc4984496/packages/eslint-plugin/src/rules/no-shadow.ts#L174) comment in tseslint is because the typescript-eslint codebase used to lean more into writing defensive, technically-never-hit code back in the day. That was at least for these pains, from TypeScript being generally a little less comprehensive, and from the test cases being less fleshed out than they are now. 

If this were sent in a PR today to typescript-eslint I'd probably:

* Post a similar comment to this one
* Suggest using `as` assertions to get around any type errors, along with:
* Adding TODO comments linking to existing or newly filed tracking issues to get the types able to know these things without assertions

lib/rules/no-shadow.js
```diff
@@ -75,10 +109,176 @@ module.exports = {
...
+		}
+
+		/**
+		 * Checks if the variable is a generic of a class
+		 * @param {Object} variable The variable to check
+		 * @returns {boolean} Whether the variable is a generic of a class
+		 */
+		function isGenericOfClass(variable) {
```
JoshuaKGoldberg, line 241: [Refactor] Similar comment on this as the previous function 🔪 

```ts
			const typeParameter = variable.identifiers[0].parent;
			const typeParameterDecl = typeParameter.parent;
			const classDecl = typeParameterDecl.parent;
			return (
				classDecl.type === "ClassDeclaration" ||
				classDecl.type === "ClassExpression"
			);
```

lib/rules/no-shadow.js
```diff
@@ -75,10 +109,176 @@ module.exports = {
...
+		/**
+		 * Checks for static method generic shadowing class generic
+		 * @param {Object} variable The variable to check
+		 * @param {Object} shadowedVariable The shadowed variable
+		 * @returns {boolean} Whether it's a static method generic shadowing class generic
+		 */
+		function isGenericOfAStaticMethodShadow(variable, shadowedVariable) {
+			return (
+				isGenericOfStaticMethod(variable) &&
+				isGenericOfClass(shadowedVariable)
```
JoshuaKGoldberg, line 278: [Refactor] Interesting: removing the `isGenericOfClass` check here doesn't fail any unit tests. I believe that's because if the variable is of a static method, it must also be of a class. Nothing else would receive `static: true`.

```suggestion
				isGenericOfStaticMethod(variable)
```

lib/rules/no-shadow.js
```diff
@@ -75,10 +109,176 @@ module.exports = {
...
+				ALLOWED_FUNCTION_VARIABLE_DEF_TYPES.has(def.node.type),
+			);
+		}
+
+		/**
+		 * Checks if the variable is a generic of a static method
+		 * @param {Object} variable The variable to check
+		 * @returns {boolean} Whether the variable is a generic of a static method
+		 */
+		function isGenericOfStaticMethod(variable) {
```
JoshuaKGoldberg, line 201: [Naming] Since this code was originally written, the general consensus on "generics" has shifted: https://www.totaltypescript.com/no-such-thing-as-a-generic. I think this would be a more precise name in line with what we'd normally call a type parameter:

```suggestion
		function isTypeParameterOfStaticMethod(variable) {
```

lib/rules/no-shadow.js
```diff
@@ -284,6 +484,30 @@ module.exports = {
 			const inner = getNameRange(variable);
 			const outer = getNameRange(scopeVar);
 
+			if (!inner || !outer || inner[1] >= outer[0]) {
+				return false;
+			}
+
+			if (!outerDef) {
+				return true;
+			}
```
JoshuaKGoldberg, line 493: [Refactor] I believe we're guaranteed `inner` and `outerDef` will always exist:

```suggestion
			if (!outer || inner[1] >= outer[0]) {
				return false;
			}
```

lib/rules/no-shadow.js
```diff
@@ -284,6 +484,30 @@ module.exports = {
...
+
+			if (hoist === "functions") {
+				return !FUNCTIONS_HOISTED_NODES.has(outerDef.node.type);
```
JoshuaKGoldberg, line 496: [Testing] I can remove both this line and the next `FUNCTIONS_HOISTED_NODES` one without failing any tests. Which feels like a test gap. 🤔

JoshuaKGoldberg (changes requested):
I am getting more and more bemused at how many existing pieces of dead code and/or gaps in code coverage are in old typescript-eslint extension rules like this one. Thanks for bearing with me on the refactor requests here. I think the rule is shaping up nicely to be a much more lean version of the current extension rule.

The strategy I've been taking is:

1. Run `npx mocha tests/lib/rules/no-shadow.js --watch` in a terminal
2. For each added line of code, comment it out and see if any tests fail

That's shown most of the `[Refactor]` and `[Testing]` change requests I've left so far.

I got through about 2/3 of the file and ran out of time this morning - but can continue through if re-requested. Cheers!

lib/rules/no-shadow.js
snitin315: I think this is unnecessary code; `hoist === "functions"` is already handled by the existing rule.

lib/rules/no-shadow.js
```diff
@@ -295,12 +519,111 @@ module.exports = {
...
+					(def.type === "TSModuleName" && def.node.declare),
+			);
+		}
+
+		/**
+		 * Checks if a variable is a duplicate of an enum name in the enum scope
+		 * @param {Object} variable The variable to check
+		 * @returns {boolean} Whether it's a duplicate enum name variable
+		 */
+		function isDuplicatedEnumNameVariable(variable) {
```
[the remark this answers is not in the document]
snitin315: Added. Should we add the corresponding FAQ in our docs as well?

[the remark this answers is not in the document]
JoshuaKGoldberg: That sounds good to me!

snitin315:
Not stale, Friendly ping @JoshuaKGoldberg for re-review.

JoshuaKGoldberg (dismissed):
Guy with mullet aggressively saying 'yes! that is awesome'

docs/src/rules/no-shadow.md
```diff
@@ -229,3 +275,91 @@ const y = (y => y)()
...
+/*eslint no-shadow: ["error", { "ignoreFunctionTypeParameterNameValueShadow": true }]*/
+
+const test = 1;
+type Func = (test: string) => typeof test;
+```
+
+:::
+
+## FAQ
+
```
snitin315, line 348:
```suggestion
```

snitin315:
Not stale, waiting for a review. @eslint/eslint-tsc

docs/src/rules/no-shadow.md
```diff
@@ -168,6 +176,44 @@ function b() {}
...
+// types
+type Bar<Foo> = 1;
+type Foo = 1;
+
+// functions
+if (true) {
+  const b = 6;
+}
+
+function b() {}
```
Tanujkanti4441, line 212: Screenshot 2025-05-11 132607

Function declarations are not reported here and seems it is intended.

https://github.com/eslint/eslint/blob/f60f2764971a33e252be13e560dccf21f554dbf1/lib/rules/no-shadow.js#L291-L294

PR #19592, opened 2025-04-01
feat: Add MCP server

### LINKED ISSUE #19588: Change Request: MCP Server
### ESLint version

HEAD

The [Model Context Protocol](https://modelcontextprotocol.io/) is the way that AIs interact with different tools. In order to allow AIs to interact with ESLint, we can provide an MCP server.

We can provide an MCP server as part of the `eslint` package. There's an [SDK](https://github.com/modelcontextprotocol/typescript-sdk) that provides the basics. To start, I think it would make sense to provide a tool to lint files (linting text is challenging because we won't have the configuration file or a directory to pull from).

This could either be an existing flag (`npx eslint --mcp`) or a new binary (`npx eslint-mcp`). 

### Participation

### Additional comments

_No response_

nzakas:
#### Prerequisites checklist

[ ] Documentation update
[ ] Add autofix to a rule
[x] Add a CLI option
[ ] Add something to the core
[ ] Other, please explain:

This change adds an MCP server that is triggered by `--mcp`. Because no other options are valid with this, I implemented it similar to `--init`.

The TypeScript SDK is a bit rough, so I had to work around a couple of things. Overall, everything works. You can test it out locally using the MCP inspector:
https://github.com/modelcontextprotocol/inspector

fixes #19588


nzakas:
I've got this working and it's kind of fun. :smile: Also added setup docs.

docs/src/use/mcp.md
```diff
@@ -0,0 +1,82 @@
...
+    parent: use eslint
+    title: MCP Server Setup
+    order: 5
+---
+
+Model Context Protocol (MCP) is an open standard that enables AI models to interact with external tools and services through a unified interface. When using GitHub Copilot in VS Code, you can connect ESLint as an MCP server to enhance your coding workflow with ESLint-specific capabilities.
+
+## Setting Up ESLint MCP Server in VS Code
+
+To add the ESLint MCP server to your VS Code workspace, follow these steps:
```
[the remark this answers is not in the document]
nzakas: Are you referring to the Copilot extension?

docs/src/use/mcp.md
```diff
@@ -0,0 +1,82 @@
...
+## Using the ESLint MCP Server with GitHub Copilot
+
+Once your MCP server is configured, you can use it with GitHub Copilot's agent mode:
+
+1. Open the Copilot Chat view in VS Code
+2. Ensure agent mode is enabled (look for the agent icon in the chat input)
+3. Toggle on the ESLint MCP server tools using the "Tools" button in the chat view
+4. Ask Copilot to perform ESLint tasks, such as:
+    - "Check this file for linting errors"
+    - "Fix all ESLint issues in the current file"
```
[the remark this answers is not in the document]
nzakas: Ooh, that is interesting. I'm not sure what the expected behavior there should be. I'll play around a bit.

bin/eslint.js
```diff
@@ -155,6 +155,26 @@ ${getErrorMessage(error)}`;
...
+		} = require("@modelcontextprotocol/sdk/server/stdio.js");
+
+		await mcpServer.connect(new StdioServerTransport());
+
+		// Note: do not use console.log() because stdout is part of the server transport
+		console.error("ESLint MCP server is running.");
+
+		process.on("SIGINT", () => {
+			mcpServer.close();
+			console.log("HI");
```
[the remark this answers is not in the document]
nzakas: Hehe, definitely a mistake.

docs/src/use/mcp.md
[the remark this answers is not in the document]
fasttime: Maybe we could say "Lint and fix the current file with ESLint". With that prompt, Copilot indicates when a problem isn't autofixable and asks before fixing it.

image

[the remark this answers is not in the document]
nzakas: It's running the same tool in both cases, so I'm not sure that wording is necessarily deterministic. I think there may be some way to provide more instructions for the tool. I'll look into it.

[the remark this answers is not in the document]
nzakas: I've tried changing the name of the tool a few different times to see if that helps, but it doesn't. Maybe we just don't provide a tool for fixing as it seems Copilot will just go and try to fix things on its own?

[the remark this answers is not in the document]
fasttime: Yes, the outcome is not deterministic. Sometimes Copilot doesn't even use the MCP server tool and instead it suggests to run `npx eslint --fix` directly.

nzakas:
Okay, I removed the "lint and fix" tool. I also removed the prompts as it's unclear how this is intended to be used and I'd rather start small and add things later as we find their helpful.

lib/mcp/mcp-server.js
```diff
@@ -0,0 +1,55 @@
...
+	"lint-files",
+	"Lint files using ESLint. This will not modify the files and will report any issues found.",
+	filePathsSchema,
+	async ({ filePaths }) => {
+		const eslint = new ESLint({
+			// enable lookup from file rather than from cwd
+			flags: ["unstable_config_lookup_from_file"],
+		});
+
+		const results = await eslint.lintFiles(filePaths);
```
fasttime, line 44: In VS Code (I haven't checked other IDEs yet) the MCP server runs with cwd set to my user's home folder. This is reflected in the `cwd` constructor option and it could affect how plugins work when files are linted. Are we comfortable with this? Alternatively we could set the current directory explicitly with `process.chdir` to avoid surprises.

docs/src/use/mcp.md
```diff
@@ -0,0 +1,152 @@
...
+### 4. Refresh Server List
+
+After adding the configuration, press the refresh button to update the list of available MCP servers.
+
+### 5. Using ESLint with Cascade
+
+Once configured, you can use ESLint tools with Cascade by asking it to:
+
+- Check files for linting errors
+- Fix ESLint issues in specific files
```
fasttime, line 131: Shall we keep this line also without the prompt to fix files?

```suggestion
```

lib/mcp/mcp-server.js
nzakas: I'm not sure there's anything we can do about this. I don't think we have a logical directory to pass to `process.chdr()`.

As a note, this is why we need the `unstable_config_lookup_from_file` flag passed to correctly find a config file for any files being linted.

lib/mcp/mcp-server.js
```diff
@@ -0,0 +1,55 @@
...
+// Server
+//-----------------------------------------------------------------------------
+
+const mcpServer = new McpServer({
+	name: "ESLint",
+	version: pkg.version,
+});
+
+const filePathsSchema = {
+	filePaths: z.array(z.string()).nonempty(),
```
caridy, line 27: try to be a lot more explicit here, otherwise the superagent/orchestrator/mcp client might give you the wrong path, something like this:

```suggestion
	filePaths: z.array(z.string().min(1).describe('Full filesystem path a file to be analyzed')).nonempty(),
```

I suspect some evolution is going to be needed around this at the MCP level.

lib/mcp/mcp-server.js
```diff
@@ -0,0 +1,55 @@
...
+			// enable lookup from file rather than from cwd
+			flags: ["unstable_config_lookup_from_file"],
+		});
+
+		const results = await eslint.lintFiles(filePaths);
+
+		return {
+			content: results.map(result => ({
+				type: "text",
+				text: JSON.stringify(result),
```
caridy, line 49: @nzakas this is where I think we have to put more effort. Most MCP servers will just return the result of the operation (calling into some tools or apis) without decorating that with some natural language explanation. In my experience, when you decorate it with some prefix and suffix in natural language, providing concrete instructions on what just happened and what to do next, that helps the super-agent to understand better the output, and the intent. Obviously, eslint is everywhere, and so the LLMs have a pretty good idea of what this output means, but I think some little instructions will help to control the LLM instead of being aggressive toward resolving all issues. Something like "... You must ask for confirmation before attempting to correct the issues identified above" can go a long way for the super-agent.

nzakas: Ah, gotcha. Thanks so much for the tip.

nzakas:
After updating the tool output with explicit instructions I'm no longer having the agent automatically fix the errors. Thanks @caridy!

lib/mcp/mcp-server.js
```diff
@@ -0,0 +1,72 @@
...
+const filePathsSchema = {
+	filePaths: z
+		.array(
+			z
+				.string()
+				.min(1)
+				.describe("Full filesystem path of a file to be analyzed"),
+		)
+		.nonempty(),
+};
```
fasttime, line 35: I'm trying to run the tool in Cursor and it's telling me that `filePaths` has a wrong type. I guess this is because something isn't right with the schema?

image

My server config in `.cursor/mcp.json` looks like this:

```json
{
	"mcpServers": {
		"eslint": {
			"command": "node",
			"args": ["/Users/fasttime/eslint/bin/eslint.js", "--mcp"],
			"env": {}
		}
	}
}
```

lib/mcp/mcp-server.js
fasttime: I thought that maybe we could set the current directory explicitly to `os.homedir()` or to `/` in case the server is started from a different location.

nzakas: I this calling `process.chdir()` is a bit dangerous. It's possible that they're still trying to figure out what the `cwd` for MCP servers and I don't think we want to override that default.

lib/mcp/mcp-server.js
nzakas: Okay, so it doesn't like the `describe()` in the file paths.

lib/mcp/mcp-server.js
nzakas: It looks like Cursor sets the cwd to the location of the Cursor executable. I'd say let's hold off on trying to change the directory. Maybe there's a way we can get the editor to give us a relevant directory.

fasttime: Windsurf also passes its own cwd to the MCP server process. We can start with the editor default then.

fasttime (approved):
LGTM, thanks! This is good enough for a start. I left some comments but I don't have any further suggestions at this time. I would like another review from a team member, if possible.

fasttime:
Oh, and we should probably add `.temp-eslintsuppressions` and `.cursor/` to gitignore.

nzakas:
I opted to fix the test so that `.temp-eslintsuppressions` is deleted after completion.

fasttime (approved):

nzakas:
Ping @mdjermanovic.

christian-kreuzberger-dtx:
While I think this change is awesome to have, it just introduced a ton of new dependencies for `eslint`.

Please consider reverting this PR and invest into #19682 .

RobinTail:
@nzakas 

If one doesn't use AI with ESLint, why do they have to install all that?
Could you make it an opt-in feature, an optional peer dependency?

image

SimenB:
There is a PR open for extracting it: #19699

RobinTail:
@SimenB , please release Jest 30

PR #19618, opened 2025-04-12
feat: add no-unassigned-vars rule

jtbandes:
#### Prerequisites checklist

See proposal by @snitin315 in https://github.com/eslint/eslint/issues/19581#issuecomment-2791481204 (and further discussion there).

New rule description:

> This rule flags `let` or `var` declarations that have no initializer and are never assigned any value, but are read/used in the code. These variables will always be `undefined` and are likely a programmer error.
> 
> This helps catch misleading patterns where a `let` appears mutable but is effectively constant and never written to.

Resolves https://github.com/eslint/eslint/issues/19581

Added `no-unassigned-vars` based on implementation from https://github.com/foxglove/eslint-plugin/pull/55 and discussion in #19581.


lib/rules/no-unassigned-vars.js
```diff
@@ -0,0 +1,76 @@
...
+		},
+
+		schema: [
+			{
+				type: "object",
+				properties: {
+					typeof: {
+						type: "boolean",
+					},
+				},
```
mdjermanovic, line 36: This option is unused; probably a copy & paste error?

lib/rules/no-unassigned-vars.js
```diff
@@ -0,0 +1,76 @@
...
+		defaultOptions: [
+			{
+				typeof: false,
+			},
+		],
+
+		docs: {
+			description:
+				"Disallow variable declarations that use `let` or `var` but have no intitial value and are never assigned",
+			recommended: true,
```
mdjermanovic, line 25:
```suggestion
			recommended: false,
```

Per our [semver policies](https://github.com/eslint/eslint?tab=readme-ov-file#semantic-versioning-policy), we can add rules to the recommended config only in major versions.

lib/rules/no-unassigned-vars.js
```diff
@@ -0,0 +1,76 @@
...
+
+	create(context) {
+		const sourceCode = context.sourceCode;
+
+		return {
+			[`VariableDeclaration:matches([kind=let], [kind=var]):not([declare=true]) > VariableDeclarator[id.type=Identifier]:not([init])`](
+				node,
+			) {
+				let hasRead = false;
+				for (const variable of sourceCode.getDeclaredVariables(node)) {
```
mdjermanovic, line 54: As in this case there's exactly one variable, maybe we can simplify this:

```js
const [variable] = sourceCode.getDeclaredVariables(node);
```

jtbandes: Maybe my ignorance about the APIs here, but how do you know there is exactly one? Is that true of every VariableDeclarator?

lib/rules/no-unassigned-vars.js
jtbandes: Thanks, is there some procedure to recommend that this rule be considered for the next major version?

kirkwaiblinger:
Question - should this supported `exported` as with `no-unused-vars` ([link](https://eslint.org/docs/latest/rules/no-unused-vars#exported))? 


In particular, thinking that something like this would be exempted from flagging
```js
// top level of a script file

/* exported x */
var x;

function foo() {
   if (x != null) {
      // ...
   }
}
```

Of course this style is quite rare now so I don't know if it's relevant to support anymore

lib/types/rules.d.ts
```diff
@@ -3687,6 +3687,13 @@ export interface ESLintRules extends Linter.RulesRecord {
 		]
 	>;
 
+	/**
+	 * Rule to disallow variable declarations that use `let` or `var` but have no intitial value and are never assigned.
```
kirkwaiblinger, line 3691:
```suggestion
	 * Rule to disallow variable declarations that use `let` or `var` but have no initial value and are never assigned.
```

lib/rules/no-unassigned-vars.js
```diff
@@ -0,0 +1,62 @@
...
+//------------------------------------------------------------------------------
+
+/** @type {import('../types').Rule.RuleModule} */
+module.exports = {
+	meta: {
+		type: "problem",
+
+		docs: {
+			description:
+				"Disallow variable declarations that use `let` or `var` but have no intitial value and are never assigned",
```
kirkwaiblinger, line 18:
```suggestion
				"Disallow variable declarations that use `let` or `var` but have no initial value and are never assigned",
```

lib/rules/no-unassigned-vars.js
kirkwaiblinger: I think it's because to declare multiple variables in a declarator you need an initializer, i.e. 
```js
var { x, y };
    ^^^^^^^^ declarator
```
is a syntax error, whereas 
```js
var { x, y } = /* ... */;
    ^^^^^^^^^^^^^^^^^^^^ declarator
```
is allowed and presumably has multiple variables as part of the declarator

docs/src/rules/no-unassigned-vars.md
```diff
@@ -0,0 +1,99 @@
+---
+title: no-unassigned-vars
+rule_type: problem
+related_rules:
+- prefer-const
+- init-declarations
+---
+
+
+This rule flags `let` or `var` declarations that have no initializer and are never assigned any value, but are read/used in the code. These variables will always be `undefined` and are likely a programmer error.
```
kirkwaiblinger, line 10: > These variables will always be `undefined` and are likely a programmer error.

Or can throw a [TDZ `ReferenceError`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/let#temporal_dead_zone_tdz)!

kirkwaiblinger:
I've left a few comments trying to be helpful since I happened to be thinking about this but I'm not an eslint member or anything, so obviously feel free to ignore! 🙂 Not trying to spam.

lib/rules/no-unassigned-vars.js
```diff
@@ -0,0 +1,62 @@
...
+"use strict";
+
+//------------------------------------------------------------------------------
+// Rule Definition
+//------------------------------------------------------------------------------
+
+/** @type {import('../types').Rule.RuleModule} */
+module.exports = {
+	meta: {
+		type: "problem",
```
Tanujkanti4441, line 14:
```suggestion
		type: "problem",
		dialects: ["typescript", "javascript"],
		language: "javascript",
```

If this rule is supporting TypeScript syntax then we must update the meta data as well?

docs/src/rules/no-unassigned-vars.md
```diff
@@ -0,0 +1,99 @@
...
+console.log(config);
+
+let one = undefined;
+if (one === two) {
+  // Noop
+}
+```
+
+:::
+
```
Tanujkanti4441, line 99: Can you add some TypeScript related examples also?

mdjermanovic:
> Question - should this support `exported` as with `no-unused-vars` ([link](https://eslint.org/docs/latest/rules/no-unused-vars#exported))?
> 
> In particular, thinking that something like this would be exempted from flagging
> 
> ```js
> // top level of a script file
> 
> /* exported x */
> var x;
> 
> function foo() {
>    if (x != null) {
>       // ...
>    }
> }
> ```
> 
> Of course this style is quite rare now so I don't know if it's relevant to support anymore

I think the first version of this rule doesn't need to support this feature, and then we can see if there's any feedback about it.  Seems like a rare case to declare and use a variable in a script but expect it to be assigned elsewhere. For `no-unused-vars`, the use case where a variable is declared and set for use elsewhere seems much more realistic.

The same for `export let x;`.

docs/src/rules/no-unassigned-vars.md
mdjermanovic: Yeah, I also don't think it's so relevant to this particular rule because forgetting to set the variable will not cause a TDZ error since it's implicitly initialized to `undefined`.

lib/rules/no-unassigned-vars.js
mdjermanovic: Yes, before each major release we open an issue to discuss all rules that have been added and decide which ones will be added to the recommended config. Example: https://github.com/eslint/eslint/issues/17596.

docs/src/rules/no-unassigned-vars.md
```diff
@@ -0,0 +1,107 @@
+---
+title: no-unassigned-vars
+rule_type: problem
+related_rules:
+- prefer-const
+- init-declarations
+---
+
+
+This rule flags `let` or `var` declarations that have no initializer and are never assigned any value, but are read/used in the code. These variables will always be `undefined` and are likely a programmer error.
```
nzakas, line 10:
```suggestion
This rule flags `let` and `var` declarations that have no initializer and are read without first being assigned a value. These variables will always be `undefined` and are likely a programmer error.
```

docs/src/rules/no-unassigned-vars.md
```diff
@@ -0,0 +1,107 @@
...
+
+This helps catch misleading patterns where a `let` appears mutable but is effectively constant and never written to.
```
nzakas, line 12: Let's remove this. I think this wording is confusing and doesn't add anything to the rule description.

docs/src/rules/no-unassigned-vars.md
nzakas: We should also show an example here. This intro is supposed to describe what the rule does with an example.

nzakas (commented):
There's an edge case I mentioned on the issue but wasn't addressed in this PR:

```js
let value;

// other code

// this should be okay
if (typeof value === "undefined") {
    // ...
}

// maybe this should be okay?
if (value === undefined) {
     // ...
}

// what about this?
if (!value) {
    // ...
}
```

kirkwaiblinger:
> There's an edge case I mentioned on the issue but wasn't addressed in this PR. What should we do when the value is compared against undefined?
> 
> ```js
> let value;
> 
> // other code
> 
> // should this be okay?
> if (typeof value === "undefined") {
>     // ...
> }
> 
> // maybe this should be okay?
> if (value === undefined) {
>      // ...
> }
> 
> // what about this?
> if (!value) {
>     // ...
> }
> ```

Why would this be relevant? I'm not seeing a mention of this on the issue?

docs/src/rules/no-unassigned-vars.md
jtbandes: The suggested wording would be inaccurate, since the rule is not checking the order of reads and writes, it is specifically checking for the case where there are some reads but there are no writes whatsoever (and no initializer).

jtbandes:
> What should we do when the value is compared against undefined?

I don’t see a reason for the rule to distinguish different kinds of reads. The point is that if the variable is being read, but it’s **never assigned** any value, then this is likely a programmer error. For your examples I’d say “it depends” — on whether `value` is ever assigned.

nzakas:
> Why would this be relevant? I'm not seeing a mention of this on the issue, maybe some context is lost?

Sorry, I'm apparently impersonating an LLM and hallucinated this. Please disregard.

docs/src/rules/no-unassigned-vars.md
```diff
@@ -0,0 +1,107 @@
...
+console.log(config);
+
+let one = undefined;
+if (one === two) {
+  // Noop
+}
+
+// In TypeScript:
+declare let value: number | undefined;
+console.log(value);
```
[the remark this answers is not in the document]
jtbandes: Actually it seems that this breaks the highlighting of lint errors in the examples (they are only highlighted in the first code block):

image


see deploy preview: https://deploy-preview-19618--docs-eslint.netlify.app/rules/no-unassigned-vars

I believe this is why I originally put all the examples in a single code block.

Given that limitation, what would you prefer to do here? IMO it's nicer if all the errors are highlighted so I would keep it in the same code block, but it's fine this way too.

docs/src/rules/no-unassigned-vars.md
```diff
@@ -0,0 +1,122 @@
...
+while (!flag) {
+  // Do something...
+}
+
+let config;
+function init() {
+  return config?.enabled;
+}
+```
+
```
Tanujkanti4441, line 59:
```suggestion

:::
```

we need to close this `incorrect` tag first.

docs/src/rules/no-unassigned-vars.md
```diff
@@ -0,0 +1,122 @@
...
+In TypeScript:
```
Tanujkanti4441, line 60:
```suggestion
In TypeScript:

::: incorrect
```

this way it will highlight the `ts` code as well

docs/src/rules/no-unassigned-vars.md
```diff
@@ -0,0 +1,122 @@
...
+}
+setup();
+console.log(config);
+
+let one = undefined;
+if (one === two) {
+  // Noop
+}
+```
+
```
Tanujkanti4441, line 113:
```suggestion

:::
```

Same for `correct` tag also.

docs/src/rules/no-unassigned-vars.md
```diff
@@ -0,0 +1,122 @@
...
+In TypeScript:
```
Tanujkanti4441, line 114:
```suggestion
In TypeScript:

::: correct
```

Tanujkanti4441 (changes requested):

Tanujkanti4441 (commented):
I am not seeing this rule in the rule list https://deploy-preview-19618--docs-eslint.netlify.app/rules/

@snitin315, shouldn't it get generated automatically?

jtbandes:
@Tanujkanti4441 I think it should be fixed now, looks like I had to run `npm run build:site`

jtbandes:
@snitin315 @nzakas @mdjermanovic @Tanujkanti4441 Thanks for all the feedback so far. Is there anything else you would like me to change at this point?

lib/rules/no-unassigned-vars.js
```diff
@@ -0,0 +1,64 @@
...
+			unassigned:
+				"'{{name}}' is not initialized and never assigned, so it will always be undefined.",
+		},
+	},
+
+	create(context) {
+		const sourceCode = context.sourceCode;
+
+		return {
+			[`VariableDeclaration:matches([kind=let], [kind=var]):not([declare=true]) > VariableDeclarator[id.type=Identifier]:not([init])`](
```
nzakas, line 36: This is a pretty long selector, which also means it will be fairly slow. Can we change this to just `VariableDeclarator` and then write the conditions in JavaScript? Something like this?

```js
const declaration = sourceCode.getParent(node);
const shouldCheck = !node.init && node.id.type === "Identifier" && declaration.kind !== "const" && !declaration.declare;
```

nzakas (dismissed):
LGTM. Would like @mdjermanovic and @snitin315 to review before merging.

docs/src/rules/no-unassigned-vars.md
```diff
@@ -0,0 +1,132 @@
+---
+title: no-unassigned-vars
+rule_type: problem
+related_rules:
+- prefer-const
```
kirkwaiblinger, line 5: consider adding `no-unused-vars` to the related rules?

jtbandes: I think I considered that, but since this rule is actually only checking _used_ vars now, I'm not really sure if I can see the argument for it being related?

Happy to change if anyone feels strongly

lib/rules/no-unassigned-vars.js
```diff
@@ -0,0 +1,72 @@
...
+/** @type {import('../types').Rule.RuleModule} */
+module.exports = {
+	meta: {
+		type: "problem",
+		dialects: ["typescript", "javascript"],
+		language: "javascript",
+
+		docs: {
+			description:
+				"Disallow variable declarations that use `let` or `var` but have no initial value and are never assigned",
```
snitin315, line 20: `no-unused-vars` will also [report vars](https://eslint.org/play/#eyJ0ZXh0IjoiLyogZXNsaW50IG5vLXVudXNlZC12YXJzOiBcImVycm9yXCIgKi9cblxubGV0IHg7XG4iLCJvcHRpb25zIjp7InJ1bGVzIjp7fSwibGFuZ3VhZ2VPcHRpb25zIjp7InBhcnNlck9wdGlvbnMiOnsiZWNtYUZlYXR1cmVzIjp7fX19fX0=) with this description.

So something like this is more accurate:

```suggestion
			description:
				"Disallow `let` or `var` variables that are read but never assigned.",
```

lib/rules/no-unassigned-vars.js
```diff
@@ -0,0 +1,72 @@
...
+			recommended: false,
+			url: "https://eslint.org/docs/latest/rules/no-unassigned-vars",
+		},
+
+		schema: [],
+		messages: {
+			unassigned:
+				"'{{name}}' is not initialized and never assigned, so it will always be undefined.",
```
snitin315, line 28: How about this?
```suggestion
				"'{{name}}' is always 'undefined' because it's never assigned.",
```

docs/src/rules/no-unassigned-vars.md
```diff
@@ -0,0 +1,132 @@
...
+::: correct
+
+```ts
+/*eslint no-unassigned-vars: "error"*/
+declare let value: number | undefined;
+console.log(value);
+```
+
+:::
+
```
snitin315, line 132: Can we also add the `When Not To Use It` section similar to all other rules?

jtbandes: What would you suggest to put in this section? I think the rule should always be used 😄 

Just looking at an example from `no-unused-vars`:

> **When Not To Use It**
> If you don’t want to be notified about unused variables or function arguments, you can safely turn this rule off.


IMO, this section is not very useful since it simply restates the rule description, but negated.

lib/rules/no-unassigned-vars.js
jtbandes: Doesn't look like there is a trailing `.` on most rules' descriptions

snitin315: Yeah, we can remove that. Added accidently

docs/src/rules/no-unassigned-vars.md
kirkwaiblinger: The clear connection in my head as a user is that no-unused-vars flags variables that are written-to-but-not-read and no-unassigned-vars flags variables that are read-but-not-written-to. But your call!

jtbandes:
@snitin315 friendly ping :)

docs/src/rules/init-declarations.md
```diff
@@ -1,6 +1,8 @@
 ---
 title: init-declarations
 rule_type: suggestion
+related_rules:
+- no-unused-vars
```
mdjermanovic, line 5:
```suggestion
- no-unassigned-vars
```

I'm guessing this was meant to be the new rule.

tests/lib/rules/no-unassigned-vars.js
```diff
@@ -0,0 +1,106 @@
...
+		"var y = undefined; log(y);",
+		"let a = x, b = y; log(a, b);",
+		"var a = x, b = y; log(a, b);",
+		"const foo = (two) => { let one; if (one !== two) one = two; }",
+	],
+	invalid: [
+		{
+			code: "let x; let a = x, b; log(x, a, b);",
+			errors: [
+				{ messageId: "unassigned", line: 1, column: 5 },
```
mdjermanovic, line 42: Can you also add `data` property to all errors in invalid test cases?

https://eslint.org/docs/latest/integrate/nodejs-api#testing-errors-with-messageid

docs/src/rules/no-unassigned-vars.md
```diff
@@ -0,0 +1,133 @@
...
+In TypeScript:
+
+::: incorrect
+
+```ts
+/*eslint no-unassigned-vars: "error"*/
+let value: number | undefined;
```
mdjermanovic, line 69:
```suggestion
/*eslint no-unassigned-vars: "error"*/

let value: number | undefined;
```

Just to add a blank line after the configuration comment.

docs/src/rules/no-unassigned-vars.md
```diff
@@ -0,0 +1,133 @@
...
+
+:::
+
+In TypeScript:
+
+::: correct
+
+```ts
+/*eslint no-unassigned-vars: "error"*/
+declare let value: number | undefined;
```
mdjermanovic, line 128:
```suggestion
/*eslint no-unassigned-vars: "error"*/

declare let value: number | undefined;
```

docs/src/rules/init-declarations.md
jtbandes: Thanks for catching that 🤦

jtbandes:
Seems like there were no errors raised by adding it to eslint-config-eslint 👍

docs/src/rules/no-unassigned-vars.md
snitin315: Something like below:

## When Not To Use It

You can disable this rule if your code intentionally uses variables that are declared but never assigned a value before being read. This might be the case in:

- Legacy codebases where uninitialized variables are used as placeholders or flags.
- Code patterns relying on default `undefined` values for control flow, though this is typically discouraged.
- Certain TypeScript use cases where variables are declared with a type and intentionally left unassigned (though using `declare` is preferred).

jtbandes: ok, added this with some modifications:
- "declared but never assigned a value before being read" → "declared and used, but are never assigned a value"
- deleted "relying on default undefined values for control flow" — this use case is fine and is not flagged by this rule assuming the variable actually gets assigned somewhere.

snitin315 (approved):
LGTM, thanks!

Leaving it open for @mdjermanovic

mdjermanovic (approved):

nzakas (approved):
LGTM. Nice work!

PR #19679, opened 2025-05-02
refactor: NodeEventGenerator -> SourceCodeTraverser

nzakas:
#### Prerequisites checklist

[ ] Documentation update
[ ] Add autofix to a rule
[ ] Add a CLI option
[ ] Add something to the core
[x] Other, please explain:

Refactor

This pull request replaces `NodeEventGenerator` with `SourceCodeTraverser` as a way to move towards allowing async rules. For right now, there's just a `traverseSync()` method, but eventually we can add a `traverse()` method that returns a promise. The `traverseSync()` method now contains the iteration logic that previously lived in the `runRules()` function as I continue to pull functionality out of `linter.js` and into more self-contained pieces.

The next step after this will be to replace `SafeEmitter` with a more general container for visitor methods that will allow for both synchronous and asynchronous calls.

refs #18787


nzakas:
I'm working on the test failure. Right now, running `npx mocha tests/lib/rules/no-sync.js` actually passes on my machine, but `npm test` fails, so that probably means there's some kind of state being left around in `SourceCodeTraverser` by another test. Figuring out which test(s) are causing the problem will be fun.

lib/linter/source-code-traverser.js
```diff
@@ -218,39 +196,115 @@ class NodeEventGenerator {
...
 	 * @returns {void}
+	 * @throws {Error} If an error occurs during traversal.
 	 */
-	leaveNode(node) {
-		this.currentAncestry.shift();
-		this.applySelectors(node, true);
+	traverseSync(sourceCode, emitter) {
+		const esquery = new ESQueryHelper(emitter, {
+			visitorKeys: sourceCode.visitorKeys ?? this.#language.visitorKeys,
+			fallback: vk,
```
mdjermanovic, line 263:
```suggestion
			fallback: vk.getKeys,
```

`fallback` should be a function.

tests/lib/linter/source-code-traverser.js
```diff
@@ -0,0 +1,387 @@
...
+				emit: {
+					value() {
+						throw new Error("Test error");
+					},
+				},
+			});
+
+			["Foo"].forEach(selector =>
+				emitterWithError.on(selector, () => {}),
+			);
```
mdjermanovic, line 241:
```suggestion
			const emitterWithError = createEmitter();

			emitterWithError.on("Foo", () => {
				throw new Error("Test error");
			});
```

To simplify the test, and I think it's more representative that a listener rather than the emitter throws an error.

tests/lib/linter/node-event-generator.js
```diff
@@ -1,565 +0,0 @@
...
-				nodeTypeKey: "kind",
-			});
-
-			const dummyNode = { kind: "Foo" };
-
-			generator.enterNode(dummyNode);
-
-			assert(emitter.emit.calledOnce);
-			assert(emitter.emit.calledWith("Foo", dummyNode));
-		});
```
mdjermanovic, line 106: I think an equivalent of this test is missing in the new test file. It's important to ensure we don't accidentally use hardcoded `node.type` without noticing, as was the case before https://github.com/eslint/eslint/commit/910bd13c4cb49001f2a9f172229360771b857585.

lib/linter/linter.js
```diff
@@ -1199,6 +1194,9 @@ function runRules(
 
 	const lintingProblems = [];
 
+	// TODO: Necessary to initialize the .parent properties of each node. Need to find a better way...
+	sourceCode.traverse();
```
mdjermanovic, line 1198: Now we'll be calling `traverse()` twice. It doesn't have much effect on our source code implementations (js, json, css, markdown), as they're all caching the steps and will just return from the cache. But for third-party implementations, this might be a breaking change or a perf hit. For example, `html-eslint` doesn't cache steps, as it probably expects  `traverse()` to be called only once.

https://github.com/yeonjuan/html-eslint/blob/91e2ce609118eb1d0b2e2078fe465776bc60bc51/packages/eslint-plugin/lib/languages/html-source-code.js#L214-L262

lib/linter/linter.js
```diff
@@ -1177,9 +1175,6 @@ function runRules(
 ) {
 	const emitter = createEmitter();
 
-	// must happen first to assign all node.parent properties
-	const eventQueue = sourceCode.traverse();
```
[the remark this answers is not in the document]
nzakas: Ok, need to think this through a bit more. Was really trying to keep the API clean by passing in just `SourceCode` to `SourceCodeTraverser`.

tests/lib/linter/source-code-traverser.js
```diff
@@ -0,0 +1,396 @@
...
+			assert.notStrictEqual(
+				instance1a,
+				instance2,
+				"Should return different instance for different language",
+			);
+		});
+	});
+
+	describe("traversing an actual AST", () => {
```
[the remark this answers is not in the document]
nzakas: Strange. I'll look into it.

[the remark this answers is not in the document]
nzakas: Okay, I know what happened now because I just reproduced it. I took the `NodeEventGenerator` tests and asked Copilot to refactor to use `SourceCodeTraverser`. In the process it removed a bunch of tests. Gotta keep an eye on these AIs.

mdjermanovic:
There are merge conflicts in `lib/linter/linter.js`.

mdjermanovic (commented):
Just one question about the tests, then LGTM.

mdjermanovic (approved):
LGTM, thanks!

PR #19695, opened 2025-05-06
feat: add `sourceCode.isGlobalReference(node)` method

snitin315:
#### Prerequisites checklist

[x] Documentation update
[ ] Add autofix to a rule
[ ] Add a CLI option
[x] Add something to the core
[ ] Other, please explain:

Fix https://github.com/eslint/eslint/issues/19651

https://github.com/eslint/eslint/pull/19695#discussion_r2083400319


lib/languages/js/source-code/source-code.js
```diff
@@ -868,6 +871,39 @@ class SourceCode extends TokenStore {
...
+		}
+
+		const scope = this.getScope(node);
+		const variable = findVariable(scope, node);
+
+		return (
+			variable !== null &&
+			variable.scope.type === "global" &&
+			variable.defs.length === 0
+		);
```
snitin315, line 904: Currently, this returns `false` in case of implicitly created globals because variable is `null` in that case for example:

```js
function test() {
  implicitGlobal = 123; // becomes global variable in non-strict mode.
}

sourceCode.isGlobalReference(implicitGlobalNode) // false;
```

Is this the correct behavior or should we return `true` in this case?

lib/languages/js/source-code/source-code.js
```diff
@@ -868,6 +871,39 @@ class SourceCode extends TokenStore {
...
+		}
+
+		// If the identifier is a property in a MemberExpression, it's not a global reference
+		if (
+			node.parent.type === "MemberExpression" &&
+			node.parent.property === node &&
+			!node.parent.computed
+		) {
+			return false;
+		}
```
fisker, line 895: Since we already add this shortcut, maybe worth to add more? 

https://github.com/sindresorhus/eslint-plugin-unicorn/blob/3646efbd14979ca3dba438325aec2ffac33fa081/rules/ast/is-reference-identifier.js#L2

lib/languages/js/source-code/source-code.js
nzakas: I think it should return `true`. @eslint/eslint-tsc thoughts?

lib/languages/js/source-code/source-code.js
nzakas: I'm not even sure it's necessary to check for a MemberExpression. Any `Identifier` node that doesn't exist in a scope will return false. The only reason to check before calling `findVariable` is if `findVariable` is very slow. 

We already are using code without a lot of additional checks, so my feeling is that they're not needed. @snitin315 what do you think?

mdjermanovic: To clarify, rules that currently use this logic are checking identifiers in specific positions, like e.g., new/call expression callees, where they can only be variable references. For a generic method that can take any node, we shouldn't assume that.

lib/languages/js/source-code/source-code.js
mdjermanovic: This is an edge case and a legacy feature of JS that should be avoided. It's reported by `no-undef` and `no-implicit-globals`.

I think `isGlobalReference()` should return `false`, and if there's ever a need to return `true` in this case, we could add an option to this method.

Scope manager doesn't track the usage of these variables, so the following might be confusing:

```js
function test() {
  implicitGlobal = 123; // if this returns `true` 
  implicitGlobal; // then it might be expected that this returns `true` as well, but we can't determine that from ScopeManager
}

```

Also, the reference doesn't have the corresponding variable in its `resolved` property, and the corresponding variable doesn't exist in global scope's `variables` (but in global scope's `implicit`), which might be another confusing thing for the consumer.

nzakas: Okay, I'm fine with it returning `false` for now. I think this is something we need to discuss more fully because, in your example, `implicitGlobal = 123` does create a global and so the next line should be a reference to that variable. Split off: https://github.com/eslint/js/issues/653

lib/languages/js/source-code/source-code.js
nzakas: Yes!

lib/languages/js/source-code/source-code.js
```diff
@@ -868,6 +868,40 @@ class SourceCode extends TokenStore {
...
+			throw new TypeError("Missing required argument: node.");
+		}
+
+		if (node.type !== "Identifier") {
+			return false;
+		}
+
+		const variable = this.scopeManager.scopes[0].set.get(node.name);
+
+		if (!variable) {
```
mdjermanovic, line 887:
```suggestion
		if (!variable || variable.defs.length > 0) {
```

Rules that use this consider variable to be overwritten/shadowed if there's a definition in the source code.

lib/languages/js/source-code/source-code.js
```diff
@@ -868,6 +868,40 @@ class SourceCode extends TokenStore {
...
+			return false;
+		}
+
+		const references = variable.references;
+
+		for (const reference of references) {
+			if (
+				reference.identifier === node &&
+				reference.resolved.scope.type === "global"
+			) {
```
mdjermanovic, line 897:
```suggestion
			if (reference.identifier === node) {
```

`reference.resolved` is `variable`, which is already certainly in the global scope.

lib/languages/js/source-code/source-code.js
snitin315: Thanks! Fixed.

nzakas (commented):

tests/lib/languages/js/source-code/source-code.js
```diff
@@ -2740,41 +2740,6 @@ describe("SourceCode", () => {
...
-			assert.strictEqual(scope.block, node);
-			assert.deepStrictEqual(
-				scope.variables.map(v => v.name),
-				[],
-			);
-			assert.deepStrictEqual(
-				scope.upper.variables.map(v => v.name),
-				["x"],
-			);
-		});
```
mdjermanovic, line 2776: I'm guessing these tests were accidentally removed?

tests/lib/languages/js/source-code/source-code.js
```diff
@@ -3606,4 +3571,680 @@ describe("SourceCode", () => {
...
+				create: context => ({
+					Program() {
+						context.sourceCode.getScope();
+					},
+				}),
+			});
+
+			assert.throws(() => {
+				linter.verify("foo", {
+					rules: { "get-scope": 2 },
```
mdjermanovic, line 3587:
```suggestion
			linter.defineRule("is-global-reference", {
				create: context => ({
					Program() {
						context.sourceCode.isGlobalReference();
					},
				}),
			});

			assert.throws(() => {
				linter.verify("foo", {
					rules: { "is-global-reference": 2 },
```

tests/lib/languages/js/source-code/source-code.js
```diff
@@ -3606,4 +3571,680 @@ describe("SourceCode", () => {
...
+							checker: {
+								create(context) {
+									const sourceCode = context.sourceCode;
+
+									spy = sinon.spy(node => {
+										if (
+											node.name === "Math" &&
+											node.parent.type ===
+												"MemberExpression" &&
+											!node.parent.object.type ===
```
mdjermanovic, line 4089:
```suggestion
											node.parent.object.type !==
```

Othwerise, this would never be `true`.

docs/src/extend/custom-rules.md
```diff
@@ -543,6 +543,7 @@ Once you have an instance of `SourceCode`, you can use the following methods on
 - `getCommentsAfter(nodeOrToken)`: Returns an array of comment tokens that occur directly after the given node or token (see the [dedicated section](#accessing-comments)).
 - `getCommentsInside(node)`: Returns an array of all comment tokens inside a given node (see the [dedicated section](#accessing-comments)).
 - `isSpaceBetween(nodeOrToken, nodeOrToken)`: Returns true if there is a whitespace character between the two tokens or, if given a node, the last token of the first node and the first token of the second node.
+- `isGlobalReference(node)`: Returns true if the given identifier node is a reference to a global variable.
```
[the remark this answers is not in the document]
nzakas: "that isn't declared in this source code" is unclear to me. I think what you're saying is that it's a reference to a configured global variable? (Either `languageOptions.globals` or the `/* global */` comment?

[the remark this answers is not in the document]
mdjermanovic: > (Either `languageOptions.globals` or the `/* global */` comment)

Plus ES globals that are automatically added for the configured `ecmaVersion`.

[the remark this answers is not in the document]
nzakas: Maybe just this?

> Returns true if the given identifier node is a reference to a previously configured global variable.

tests/lib/languages/js/source-code/source-code.js
```diff
@@ -3606,4 +3571,680 @@ describe("SourceCode", () => {
...
+			flatLinter.verify(code, config);
+			assert(identifierSpy.called, "Identifier spy was not called.");
+			assert(
+				identifierSpy.callCount > 10,
+				"Identifier spy was not called enough times.",
+			);
+		});
+
+		it("should handle function parameters and shadowed globals", () => {
+			const code = "function test(param, console) { param; console; }";
```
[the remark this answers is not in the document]
snitin315: Changed it to `NaN`.

docs/src/extend/custom-rules.md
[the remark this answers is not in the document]
nzakas: Works for me.

nzakas:
@snitin315 please check the merge conflicts.

docs/src/_data/further_reading_links.json
mdjermanovic, line 1: Looks like this file was accidentally included in this PR?

mdjermanovic (approved):
LGTM, thanks! Leaving open for @nzakas to verify changes after previous approval.

Issue #7941, opened 2017-01-18
func-style ignores function expressions not starting with var keyword


kjholla:
The func-style rule has been set to enforce use of function declarations.

```js
var eslintConfig = {
	'root': true,
	'env': {
		'browser': true
	},
	'extends': 'eslint:recommended',
	'rules': {
		/** Coding Style Standards **/
		'func-style': ['error', 'declaration', { 'allowArrowFunctions': true }]
	}
}

module.exports = eslintConfig;

```

I would expect that with this configuration, code such as the one below would lead to an error. But it doesn't.

```js
var fnVariable;

fnVariable = function() {
      return true;
}

```

eslintbot:
Hi @kjholla, thanks for the issue. It looks like there's not enough information for us to know how to help you.

If you're reporting a bug, please be sure to include:

1. The version of ESLint you are using (run `eslint -v`)
2. What you did (the source code and ESLint configuration)
3. The actual ESLint output complete with numbers
4. What you expected to happen instead

Requesting a new rule? Please see [Proposing a New Rule](http://eslint.org/docs/developer-guide/contributing/new-rules) for instructions.

Requesting a rule change? Please see [Proposing a Rule Change](http://eslint.org/docs/developer-guide/contributing/rule-changes) for instructions.

If it's something else, please just provide as much additional information as possible. Thanks!

platinumazure:
Can't reproduce on the ESLint demo. My guess is you have an old version of ESLint where this was a bug, but upgrading to latest should fix this for you.

In future, please include your ESLint version.

kjholla:
Hello,

Thank you for the response.

My setup is as below:

Node version: 6.2.2
NPM version: 3.9.5
ESLint version: 3.13.1

ESLint Config File (.eslintrc.js):
```
var eslintConfig = {
	'root': true,
	'env': {
		'browser': true
	},
	'extends': 'eslint:recommended',
	'rules': {
		/** Coding Style Standards **/
		'func-style': ['error', 'declaration', { 'allowArrowFunctions': true }]
	}
}

module.exports = eslintConfig;
```


My Source Code:
```
var fnVariable

fnVariable();

fnVariable = function() {
	var a, b;
	a = 2;
	b = 2;
	return a + b;
}
```

What I would have expected: An error as below
`  5:1  error  Expected a function declaration  func-style
`

What I get: No error at all.

ljharb:
[...]

Issue #2594, opened 2015-05-22
Multiple scoped shared configs require the prefix


MethodGrab:
ESLint v0.21.2
Relevant code: ([#config.js#84](https://github.com/eslint/eslint/blob/master/lib/config.js#L84]))

### Expected:
Multiple, **scoped**, shared configs in a single module should be extendable from a `.eslintrc` without the `eslint-config-` prefix (as non-scoped configs are).

Assuming a module called `eslint-config-test` with a default config `index.js` and a sub-config `modern.js` then all of these should work:  
```
{ "extends": "@scope/eslint-config-test" }
```
```
{ "extends": "@scope/eslint-config-test/modern" }
```
```
{ "extends": "@scope/test" }
```
```
{ "extends": "@scope/test/modern" }
```

### Actual:
The 4th example, extending a sub-config, will fail with the error `Cannot find module '@scope/test/eslint-config-modern'` unless the root module name includes the `eslint-config-` prefix.

These will work:
```
{ "extends": "@scope/test" }
```
```
{ "extends": "@scope/eslint-config-test/modern" }
```

nzakas:
Looks like a bug.

gcochard:
Should it be inserting the `eslint-config-` after the *first* slash, rather than *last*?

nzakas:
Yup

gcochard:
I'll take this.

MethodGrab:
Thanks for the confirmation and quick work on a fix! :+1:

PR #19714, opened 2025-05-08
feat: allow global type declaration in `no-var`

remcohaszing:
When declaring a global variable in TypeScript, using `var` vs `let`/`const` yields different behaviour. A declaration using `var` yields the expected result.

See [TypeScript playground](https://www.typescriptlang.org/play/?#code/PQgEB4CcFMDNpgOwMbVAGwJYCMC8AiAEwHsBbfUYAPgFgAoew6ZdAQxlAHN1jtX1QAb3qhRGaABdQAGUkAuUAGcJkTIk4ixAN3agAauwXLV6+ptFqJCWK1SgA6mpIB3IebGjHiIyrUa6HgC+9MEMdGDcvPygtqiKivSyEvQGkPReZuHAoM5OxK6ckvS5iC4AdEnFec5lqVWl+WUZYWAlLkpFdG2NSaC4oADkA-XlqX2Dw13VTWrjQ5kRPHzoACoAFpiKXJ2Ry+ubFTtL-PuKtez0uycbZ830i1GrNx3JdFdPB73982-HH2djb6Td6nGaIOaTe7ZRTQdCwbavGFww6I2Gwc5pOhI9F3LIdOEvejYlEQolojGkrHkryU+jQAAeAAdiJApIJAkA)

Refs #19173  
Refs https://github.com/microsoft/vscode/pull/248075  
Refs https://github.com/typescript-eslint/typescript-eslint/issues/2594  
Refs https://github.com/typescript-eslint/typescript-eslint/issues/7941

#### Prerequisites checklist

[ ] Documentation update
[ ] Add autofix to a rule
[ ] Add a CLI option
[ ] Add something to the core
[ ] Other, please explain:

- **Node version:** v22.15.0
- **npm version:** v10.9.2
- **Local ESLint version:** v9.26.0
- **Operating System:** linux 6.12.10-76061203-generic

[ ] `Default (Espree)`
[x] `@typescript-eslint/parser`
[ ] `@babel/eslint-parser`
[ ] `vue-eslint-parser`
[ ] `Other`

<details>

https://github.com/remcohaszing/eslint

</details>

```ts
declare global {
  var bar: 'car'
}
```

Nothing

N/A

The `no-var` rule no longer reports the direct use of `var` within a TypeScript `declare global` block.

### DIFF
--- a/docs/src/_data/further_reading_links.json
+++ b/docs/src/_data/further_reading_links.json
@@ -1,800 +1,800 @@
 {
-    "https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/set": {
-        "domain": "developer.mozilla.org",
-        "url": "https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/set",
-        "logo": "https://developer.mozilla.org/favicon-48x48.cbbd161b.png",
-        "title": "setter - JavaScript | MDN",
-        "description": "The set syntax binds an object property to a function to be called when there is an attempt to set that property."
-    },
-    "https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/get": {
-        "domain": "developer.mozilla.org",
-        "url": "https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/get",
-        "logo": "https://developer.mozilla.org/favicon-48x48.cbbd161b.png",
-        "title": "getter - JavaScript | MDN",
-        "description": "The get syntax binds an object property to a function that will be called when that property is looked up."
-    },
-    "https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Working_with_Objects": {
-        "domain": "developer.mozilla.org",
-        "url": "https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Working_with_Objects",
-        "logo": "https://developer.mozilla.org/favicon-48x48.cbbd161b.png",
-        "title": "Working with objects - JavaScript | MDN",
-        "description": "JavaScript is designed on a simple object-based paradigm. An object is a collection of properties, and a property is an association between a name (or key) and a value. A property’s value can be a function, in which case the property is known as a method. In addition to objects that are predefined i…"
-    },
-    "https://github.com/airbnb/javascript#arrows--one-arg-parens": {
-        "domain": "github.com",
-        "url": "https://github.com/airbnb/javascript#arrows--one-arg-parens",
-        "logo": "https://github.com/fluidicon.png",
-        "title": "GitHub - airbnb/javascript: JavaScript Style Guide",
-        "description": "JavaScript Style Guide. Contribute to airbnb/javascript development by creating an account on GitHub."
-    },
-    "https://www.adequatelygood.com/JavaScript-Scoping-and-Hoisting.html": {
-        "domain": "www.adequatelygood.com",
-        "url": "https://www.adequatelygood.com/JavaScript-Scoping-and-Hoisting.html",
-        "logo": "https://www.adequatelygood.com/favicon.ico",
-        "title": "JavaScript Scoping and Hoisting",
-        "description": null
-    },
-    "https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/var#var_hoisting": {
-        "domain": "developer.mozilla.org",
-        "url": "https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/var#var_hoisting",
-        "logo": "https://developer.mozilla.org/favicon-48x48.cbbd161b.png",
-        "title": "var - JavaScript | MDN",
-        "description": "The var statement declares a function-scoped or globally-scoped variable, optionally initializing it to a value."
-    },
-    "https://en.wikipedia.org/wiki/Indent_style": {
-        "domain": "en.wikipedia.org",
-        "url": "https://en.wikipedia.org/wiki/Indent_style",
-        "logo": "https://en.wikipedia.org/static/apple-touch/wikipedia.png",
-        "title": "Indentation style - Wikipedia",
-        "description": null
-    },
-    "https://github.com/maxogden/art-of-node#callbacks": {
-        "domain": "github.com",
-        "url": "https://github.com/maxogden/art-of-node#callbacks",
-        "logo": "https://github.com/fluidicon.png",
-        "title": "GitHub - maxogden/art-of-node: a short introduction to node.js",
-        "description": ":snowflake: a short introduction to node.js. Contribute to maxogden/art-of-node development by creating an account on GitHub."
-    },
-    "https://web.archive.org/web/20171224042620/https://docs.nodejitsu.com/articles/errors/what-are-the-error-conventions/": {
-        "domain": "web.archive.org",
-        "url": "https://web.archive.org/web/20171224042620/https://docs.nodejitsu.com/articles/errors/what-are-the-error-conventions/",
-        "logo": "https://archive.org/favicon.ico",
-        "title": "What are the error conventions? - docs.nodejitsu.com",
-        "description": "docs.nodejitsu.com is a growing collection of how-to articles for node.js, written by the community and curated by Nodejitsu and friends. These articles range from basic to advanced, and provide relevant code samples and insights into the design and philosophy of node itself."
-    },
-    "https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes": {
-        "domain": "developer.mozilla.org",
-        "url": "https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes",
-        "logo": "https://developer.mozilla.org/favicon-48x48.cbbd161b.png",
-        "title": "Classes - JavaScript | MDN",
-        "description": "Classes are a template for creating objects. They encapsulate data with code to work on that data. Classes in JS are built on prototypes but also have some syntax and semantics that are not shared with ES5 class-like semantics."
-    },
-    "https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/static": {
-        "domain": "developer.mozilla.org",
-        "url": "https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/static",
-        "logo": "https://developer.mozilla.org/favicon-48x48.cbbd161b.png",
-        "title": "static - JavaScript | MDN",
-        "description": "The static keyword defines a static method or property for a class, or a class static initialization block (see the link for more information about this usage). Neither static methods nor static properties can be called on instances of the class. Instead, they’re called on the class itself."
-    },
-    "https://www.crockford.com/code.html": {
-        "domain": "www.crockford.com",
-        "url": "https://www.crockford.com/code.html",
-        "logo": "https://www.crockford.com/favicon.png",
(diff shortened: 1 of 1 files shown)


lib/rules/no-var.js
```diff
@@ -346,9 +346,19 @@ module.exports = {
 
 		return {
 			"VariableDeclaration:exit"(node) {
-				if (node.kind === "var") {
```
snitin315, line 349: Currently, this rule doesn't support TS syntax. If we add TS support to this rule, we need to add the following meta properties in the schema.

https://github.com/eslint/eslint/blob/58a171e8f0dcc1e599ac22bf8c386abacdbee424/lib/rules/no-loss-of-precision.js#L16-L17

tests/lib/rules/no-var.js
```diff
@@ -36,6 +36,12 @@ ruleTester.run("no-var", rule, {
 				parserOptions: { ecmaFeatures: { globalReturn: true } },
 			},
 		},
+		{
+			code: "declare global { var bar: 'car' }",
+			languageOptions: {
+				parser: require("@typescript-eslint/parser"),
+			},
+		},
```
snitin315, line 44: What we have been following up untill now for TS cases is creating a separate `ruleTesterTypescript` instance. You can refer this PR https://github.com/eslint/eslint/pull/19560/

https://github.com/eslint/eslint/blob/58a171e8f0dcc1e599ac22bf8c386abacdbee424/tests/lib/rules/no-loss-of-precision.js#L322-L327

snitin315 (commented):
Thanks for picking this up, few comments.

Tanujkanti4441 (commented):
Could you also add a section in `no-var` docs with `correct`, `incorrect` examples that says it also support Typescript syntax?

for reference https://eslint.org/docs/latest/rules/no-invalid-this

remcohaszing:
Good catch! Done

docs/src/_data/further_reading_links.json
Tanujkanti4441, line 1: I think this file is formatted automatically. Can you revert the changes in this file?

docs/src/rules/no-var.md
```diff
@@ -54,6 +54,42 @@ const CONFIG = {};
 
 :::
 
+This rule additionally supports TypeScript type syntax. There are multiple ways to declare global variables in TypeScript. Only using `var` works for all cases. See this [TypeScript playground](https://www.typescriptlang.org/play/?#code/PQgEB4CcFMDNpgOwMbVAGwJYCMC8AiAEwHsBbfUYAPgFgAoew6ZdAQxlAHN1jtX1QAb3qhRGaABdQAGUkAuUAGcJkTIk4ixAN3agAauwXLV6+ptFqJCWK1SgA6mpIB3IebGjHiIyrUa6HgC+9MEMdGDcvPygtqiKivSyEvQGkPReZuHAoM5OxK6ckvS5iC4AdEnFec5lqVWl+WUZYWAlLkpFdG2NSaC4oADkA-XlqX2Dw13VTWrjQ5kRPHzoACoAFpiKXJ2Ry+ubFTtL-PuKtez0uycbZ830i1GrNx3JdFdPB73982-HH2djb6Td6nGaIOaTe7ZRTQdCwbavGFww6I2Gwc5pOhI9F3LIdOEvejYlEQolojGkrHkryU+jQAAeAAdiJApIJAkA) for reference.
```
Tanujkanti4441, line 57:
```suggestion
This rule additionally supports TypeScript type syntax. There are multiple ways to declare global variables in TypeScript. Only using `var` works for all cases.
```

I think we can avoid this playground link as we already have examples below and we are working on typescript syntax support in ESLint playground.

remcohaszing: > we are working on typescript syntax support in ESLint playground.

Neat!

Still, I think the TypeScript playground does add context. The code example alone doesn’t show why it’s ok (better even) to use `var` instead of `let` inside a `declare global` block. I’ll gladly remove it if you’re still not convinced.

Tanujkanti4441: In that case, instead of the playground link, how about if we add a code block in the docs page to explain why we need `var` in TypeScript with appropriate code comments in block?

you can see [this](https://eslint.org/docs/latest/rules/no-await-in-loop) page for reference.

remcohaszing:
I’m having some trouble to properly explain this without going into too much detail and showing code samples. This is why I really feel the need for the TypeScript playground link.

Maybe it would be best to
1. First remove the playground link as @Tanujkanti4441 suggested;
2. Implement https://github.com/typescript-eslint/typescript-eslint/issues/2594;
3. In a follow-up, link to that rule for a deeper explanation.

Tanujkanti4441:
> I’m having some trouble to properly explain this without going into too much detail and showing code samples. This is why I really feel the need for the TypeScript playground link.
> 
> Maybe it would be best to
> 
>1. First remove the playground link as @Tanujkanti4441 suggested;
> 
>2. Implement [Rule proposal: prefer `declare var` for global declarations typescript-eslint/typescript-eslint#2594](https://github.com/typescript-eslint/typescript-eslint/issues/2594);
> 
>3. In a follow-up, link to that rule for a deeper explanation.

Okay! lets do it this way, but can you address the suggestion in https://github.com/eslint/eslint/pull/19714#issuecomment-2872951727?

docs/src/rules/no-var.md
nzakas: I think it's fine to link to the TypeScript playground for demonstration purposes. Our own playground won't have the type checking so not as solid of an example.

lib/rules/no-var.js
```diff
@@ -346,9 +348,19 @@ module.exports = {
...
-					report(node);
+				if (node.kind !== "var") {
+					return;
```
kirkwaiblinger, line 352: In https://github.com/typescript-eslint/typescript-eslint/issues/7941, the proposal is to _force_ `var` for global ambient declarations rather than just to permit it. Is the discrepancy in behavior here, where non-`var`s are simply ignored, intentional?

lib/rules/no-var.js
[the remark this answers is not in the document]
kirkwaiblinger: Here's a little minimal repository demonstrating this behavior: https://github.com/kirkwaiblinger/tsglobalvardemo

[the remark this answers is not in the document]
kirkwaiblinger: Come to think of it, it actually doesn't matter whether it's a `.d.ts` file or not. The same behavior applies with `declare`d variables in ordinary `.ts` files.

```ts
// foo.ts

declare var globalVar: string;
declare let globalLet: string;
declare const globalConst: string;


// read unqualified
console.log(globalConst); // works
console.log(globalLet); // works
console.log(globalVar); // works

// write unqualified
globalConst = 'doesnt work';
globalLet = 'works';
globalVar = 'works';

// read on globalThis
console.log(globalThis.globalConst); // doesnt work
console.log(globalThis.globalLet); // doesnt work
console.log(globalThis.globalVar); // works

// write on globalThis
globalThis.globalConst = 'doesnt work';
globalThis.globalLet = 'doesnt work';
globalThis.globalVar = 'works';

// read on self
console.log(self.globalConst); // doesnt work
console.log(self.globalLet); // doesnt work
console.log(self.globalVar); // works

// write on self
self.globalConst = 'doesnt work';
self.globalLet = 'doesnt work';
self.globalVar = 'works';
```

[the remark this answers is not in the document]
remcohaszing: What matters is whether TypeScript considers file to be a script or a module. If a module contains imports or exports, it’s a module. Otherwise it’s a script. In a script everything you declare is global. In a module you can declare globals with `declare global`.

So this module:

```ts
declare global {
  var x: number;
}

export {}
```

is equivalent to this script:

```ts
declare var x: number;
```

However, TypeScript also has the option `moduleDetection`. If this is set to `force`, everything source file considered a module. This makes it ambiguous whether the latter is a local variable declaration or a global. We could assume default options. The `moduleDetection` option doesn’t apply to declaration files.

lib/rules/no-var.js
remcohaszing: That’s #2594, though there is discussion about it in #7941. I have been fiddling a bit to implement #2594. I don’t think it should have been closed as a duplicate. They’re two different things.

kirkwaiblinger: Hmm, I see what you're saying... 

However, I think it's not such a good idea to have two separate rules, especially in two separate plugins, handle overlapping and possibly-conflicting logic. It's like if [return-await](https://typescript-eslint.io/rules/return-await/) were actually two separate rules for when `return await` is required and when it's prohibited, and those two rules were in separate plugins, and one of them didn't even have access to the TS API [required for correctness in some cases](https://github.com/eslint/eslint/pull/19714#discussion_r2090566074).

What do you think of the following proposal instead? What if we draw the line such that this rule, `no-var`, ignores _all_ TS ambient declarations, even those that _shouldn't_ be `var`s, and then the typescript-eslint rule (maybe called something like `declare-var`) would have the job of properly enforcing _or_ prohibiting `var`s in all cases for `declare`d variables. This way detailed TS knowledge stays out of `no-var` and the dividing line between what's handled by typescript-eslint and what's handled by `no-var` is much more clear.

Concretely, that would mean I propose that all of the following, currently documented as incorrect, would be permitted by `no-var`...
```ts
declare var x: number

declare namespace ns {
	var x: number
}

declare module 'module' {
	var x: number
}
```
and instead would be banned by a new typescript-eslint rule.

nzakas: I think the most correct implementation is as you suggested in https://github.com/eslint/eslint/pull/19714#discussion_r2091481695. `no-var` should not be *requiring* `var` anywhere, it should be *allowing* `var` where another option doesn't make sense.

remcohaszing: Actually this even applies to regular code, not just declarations. For example, the following file is also a global script as far as TypeScript is concerned. It doesn’t even matter if it’s JavaScript or TypeScript syntax.

```js
var a = 1
```

How far should the `no-var` exceptions go? Also how many users will actually run into this? I think most code is written as a module today, or at least that’s often the intention.

There are only 2 cases where the user should definitely use `var`:

1. Within `declare global`, which is handled by this PR.
2. Within a TypeScript declaration file that doesn’t contain ESM syntax. This can be detected based on the file name (`*.d.ts` / `*.d.cts` / `*.d.mts` / `*.d.*.ts`). Should I add support for these?

nzakas: The original intent of this rule is to ban the use of `var` completely. The goal of this PR is to mark as an exception those places in TypeScript where changing from `var` to something else either isn't possible or changes the behavior in an undesirable way.

> 2\. Within a TypeScript declaration file that doesn’t contain ESM syntax. This can be detected based on the file name (`*.d.ts` / `*.d.cts` / `*.d.mts` / `*.d.*.ts`). Should I add support for these?

At this point I'd say no. This feels like a step too far. A better approach for a user is to disable this rule just in those files.

remcohaszing:
This PR sparked quite some discussion. What still needs to happen for this PR to be ready?

The summary as I see it now:

- People should prefer `var` over `let` or `const` when declaring globals — Out of scope for this rule. Makes sense elsewhere. No action needed (in this PR).
- There are more false positives that can be detected without a TypeScript context in declaration files — Edge case. People can disable `no-var` if they have to. No action needed.
- The link to the TypeScript playground should be removed — Keep it. It’s useful context for demonstration purposes. No action needed.
- Explanation with an example — This is already explained. The example code requires type-checking to make sense and is on the TypeScript playground. No action needed.

nzakas (approved):

snitin315 (approved):
LGTM, thanks!

PR #19727, opened 2025-05-13
docs: upfront recommend against using the no-return-await rule

mikedidomizio:
#### Prerequisites checklist

Upfront recommendation in the docs against not using this rule anymore.

[x] Documentation update
[ ] Add autofix to a rule
[ ] Add a CLI option
[ ] Add something to the core
[ ] Other, please explain:

A lot of rules have valid reasons to not use them, but in this case I don't see any reason anymore to even entertain this rule. Since this was a rule that made sense previously, it would be nice if now referencing it or glancing back at it, that it would upfront tell you that it is now not recommended.

chrome_ZOfXN8mWw8

### DIFF
--- a/docs/src/rules/no-return-await.md
+++ b/docs/src/rules/no-return-await.md
@@ -14,17 +14,9 @@ It is NOT recommended to use the `no-return-await` rule anymore because:
 
 Historical context: When promises were first introduced, calling `return await` introduced an additional microtask, one for the `await` and one for the return value of the async function. Each extra microtask delays the computation of a result and so this rule was added to help avoid this performance trap. Later, [V8 changed the way](https://v8.dev/blog/fast-async) `return await` worked so it would create a single microtask, which means this rule is no longer necessary.
 
-***
-
-This rule warns on any usage of `return await`.
-
-Using `return await` inside an `async function` keeps the current function in the call stack until the Promise that is being awaited has resolved. `return await` can also be used in a try/catch statement to catch errors from another function that returns a Promise.
-
-You can avoid the extra microtask by not awaiting the return value, with the trade off of the function no longer being a part of the stack trace if an error is thrown asynchronously from the Promise being returned. This can make debugging more difficult.
-
 ## Rule Details
 
-This rule aims to prevent a likely common performance hazard due to a lack of understanding of the semantics of `async function`.
+This rule warns on any usage of `return await`.
 
 Examples of **incorrect** code for this rule:
 
@@ -74,8 +66,4 @@ async function foo4() {
 
 ## When Not To Use It
 
-There are a few reasons you might want to turn this rule off:
-
-* If you want to use `await` to denote a value that is a thenable
-* If you do not want the performance benefit of avoiding `return await`
-* If you want the functions to show up in stack traces (useful for debugging purposes)
+You should not use this rule. There is no reason to avoid `return await`.
--- a/docs/src/rules/no-return-await.md
+++ b/docs/src/rules/no-return-await.md
@@ -16,7 +16,7 @@ Historical context: When promises were first introduced, calling `return await` 
 
 ## Rule Details
 
-This rule warns on any usage of `return await`.
+This rule warns on any usage of `return await` except in `try` blocks.
 
 Examples of **incorrect** code for this rule:
 
--- a/docs/src/rules/no-return-await.md
+++ b/docs/src/rules/no-return-await.md
@@ -9,8 +9,8 @@ further_reading:
 
 It is NOT recommended to use the `no-return-await` rule anymore because:
 
-- `return await` on a promise will not result in an extra microtask.
-- `return await` yields a better stack trace for debugging.
+* `return await` on a promise will not result in an extra microtask.
+* `return await` yields a better stack trace for debugging.
 
 The following documentation is kept for historical context, and is not necessarily accurate anymore.
 
--- a/docs/src/rules/no-return-await.md
+++ b/docs/src/rules/no-return-await.md
@@ -12,7 +12,7 @@ It is NOT recommended to use the `no-return-await` rule anymore because:
 * `return await` on a promise will not result in an extra microtask.
 * `return await` yields a better stack trace for debugging.
 
-Historical context: When promises were first introduced, calling `return await` introduced an additional microtask, one for the `await` and one for the return value of the async function. Each extra microtask delays the computation of a result and so this rule was added to help avoid this performance trap. Later, [V8 changed the way](https://v8.dev/blog/fast-async) `return await` worked so it would create a single microtask, which means this rule is no longer necessary.
+Historical context: When promises were first introduced, calling `return await` introduced an additional microtask, one for the `await` and one for the return value of the async function. Each extra microtask delays the computation of a result and so this rule was added to help avoid this performance trap. Later, [ECMA-262 changed the way](https://github.com/tc39/ecma262/pull/1250) `return await` worked so it would create a single microtask, which means this rule is no longer necessary.
 
 ## Rule Details
 
--- a/docs/src/rules/no-return-await.md
+++ b/docs/src/rules/no-return-await.md
@@ -18,7 +18,7 @@ Historical context: When promises were first introduced, calling `return await` 
 
 This rule warns on any usage of `return await`.
 
-Using `return await` inside an `async function` keeps the current function in the call stack until the Promise that is being awaited has resolved, at the cost of an extra microtask before resolving the outer Promise. `return await` can also be used in a try/catch statement to catch errors from another function that returns a Promise.
+Using `return await` inside an `async function` keeps the current function in the call stack until the Promise that is being awaited has resolved. `return await` can also be used in a try/catch statement to catch errors from another function that returns a Promise.
 
 You can avoid the extra microtask by not awaiting the return value, with the trade off of the function no longer being a part of the stack trace if an error is thrown asynchronously from the Promise being returned. This can make debugging more difficult.
 
--- a/docs/src/rules/no-return-await.md
+++ b/docs/src/rules/no-return-await.md
@@ -12,9 +12,11 @@ It is NOT recommended to use the `no-return-await` rule anymore because:
 * `return await` on a promise will not result in an extra microtask.
 * `return await` yields a better stack trace for debugging.
 
-The following documentation is kept for historical context, and is not necessarily accurate anymore.
+Historical context: When promises were first introduced, calling `return await` introduced an additional microtask, one for the `await` and one for the return value of the async function. Each extra microtask delays the computation of a result and so this rule was added to help avoid this performance trap. Later, [V8 changed the way](https://v8.dev/blog/fast-async) `return await` worked so it would create a single microtask, which means this rule is no longer necessary.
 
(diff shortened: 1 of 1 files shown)


docs/src/rules/no-return-await.md
```diff
@@ -7,6 +7,13 @@ further_reading:
 - https://jakearchibald.com/2017/await-vs-return-vs-return-await/
 ---
 
+It is NOT recommended to use the `no-return-await` rule anymore because:
+
+* `return await` on a promise will not result in an extra microtask.
+* `return await` yields a better stack trace for debugging.
+
+The following documentation is kept for historical context, and is not necessarily accurate anymore.
+
```
JoshuaKGoldberg, line 16: @kirkwaiblinger's done a ton of work on the tseslint side. I think it would be good to get Kirk's thoughts here if possible.

JoshuaKGoldberg, line 16: @kirkwaiblinger's done a ton of work on the tseslint side. I think it would be good to get Kirk's thoughts here if possible.

docs/src/rules/no-return-await.md
```diff
@@ -7,6 +7,13 @@ further_reading:
 - https://jakearchibald.com/2017/await-vs-return-vs-return-await/
 ---
 
+It is NOT recommended to use the `no-return-await` rule anymore because:
+
+* `return await` on a promise will not result in an extra microtask.
+* `return await` yields a better stack trace for debugging.
+
+The following documentation is kept for historical context, and is not necessarily accurate anymore.
```
kirkwaiblinger, line 15: (praise) I think the callout that the rest of the docs page contains inaccurate information is a very good addition. It's annoyed me for a while that this page contained out of date information without such a disclaimer.

(suggestion) In order to be clear where the inaccurate information starts, what do you think about adding a horizontal line after this paragraph (a `***` in markdown)?

docs/src/rules/no-return-await.md
```diff
@@ -7,6 +7,13 @@ further_reading:
 - https://jakearchibald.com/2017/await-vs-return-vs-return-await/
 ---
 
+It is NOT recommended to use the `no-return-await` rule anymore because:
+
+* `return await` on a promise will not result in an extra microtask.
+* `return await` yields a better stack trace for debugging.
```
kirkwaiblinger, line 13: Much of this feels redundant with the paragraph 

> The original intent of this rule was to discourage the use of `return await`, to avoid an extra microtask. However, due to the fact that JavaScript now handles native `Promise`s differently, there is no longer an extra microtask. More technical information can be found in [this V8 blog entry](https://v8.dev/blog/fast-async).

Can you think of a way to combine them / remove the redundancy?

kirkwaiblinger (commented):
Overall I like this change! 

Here's an idea of what my suggested tweaks might look like:

suggested changes

docs/src/rules/no-return-await.md
kirkwaiblinger, line 15: (optional) Important components about the docs are certainly not accurate, so you could go stronger on this, rather than hedging.

```suggestion
The following documentation is kept for historical context, and is not accurate anymore.
```

docs/src/rules/no-return-await.md
nzakas: I agree. Let's just remove the paragraph. The bulleted list is all we need.

docs/src/rules/no-return-await.md
nzakas: I think we can just be plain here.

```suggestion
Historical context: When promises were first introduced, calling `return await` introduced an additional microtask, one for the `await` and one for the return value of the async function. Each extra microtask delays the computation of a result and so this rule was added to help avoid this performance trap. Later, V8 changed the way `return await` worked so it would create a single microtask, which means this rule is no longer necessary.
```

nzakas (commented):
For clarity, I'd remove the paragraphs prior to "Rule Details". I'd also update the first paragraph under "Rule Details" to say:

> This rule warns on any usage of `return await`. 

That makes it clear that there's no performance issue any longer.

docs/src/rules/no-return-await.md
mikedidomizio: I like the paragraph that explains things for anyone curious. I've updated the PR to include the paragraph. Since the paragraph following this change was felt to be redundant (which I agree), I've re-included the link to the V8 blog entry into the suggested paragraph for anyone that wants to dive deeper.

docs/src/rules/no-return-await.md
mikedidomizio: Thanks Josh for help pinging the right people. Appreciate it.

docs/src/rules/no-return-await.md
mikedidomizio: Paragraph removed, [commented here](https://github.com/eslint/eslint/pull/19727#discussion_r2089604493).  Going to mark this thread as resolved.

docs/src/rules/no-return-await.md
mikedidomizio: I've went with the [suggested paragraph](https://github.com/eslint/eslint/pull/19727#discussion_r2089125889) which does not include that everything after is not accurate. If you feel there's value in having it, let me know.

mikedidomizio:
Thanks @nzakas and @kirkwaiblinger for the comments. All great. I've consolidated the suggestions and here is where we stand now.

chrome_oiaTB0VyZQ

docs/src/rules/no-return-await.md
```diff
@@ -7,7 +7,16 @@ further_reading:
...
+
+Historical context: When promises were first introduced, calling `return await` introduced an additional microtask, one for the `await` and one for the return value of the async function. Each extra microtask delays the computation of a result and so this rule was added to help avoid this performance trap. Later, [V8 changed the way](https://v8.dev/blog/fast-async) `return await` worked so it would create a single microtask, which means this rule is no longer necessary.
+
+***
+
+This rule warns on any usage of `return await`.
 
 Using `return await` inside an `async function` keeps the current function in the call stack until the Promise that is being awaited has resolved, at the cost of an extra microtask before resolving the outer Promise. `return await` can also be used in a try/catch statement to catch errors from another function that returns a Promise.
```
nzakas, line 21: One last fix

```suggestion
Using `return await` inside an `async function` keeps the current function in the call stack until the Promise that is being awaited has resolved. `return await` can also be used in a try/catch statement to catch errors from another function that returns a Promise.
```

mikedidomizio: Updated, removed mention of the extra microtask https://github.com/eslint/eslint/pull/19727/commits/d71bf2e7f725c62d2d29cda35555e79e99671d77

docs/src/rules/no-return-await.md
```diff
@@ -7,9 +7,18 @@ further_reading:
...
+
+Historical context: When promises were first introduced, calling `return await` introduced an additional microtask, one for the `await` and one for the return value of the async function. Each extra microtask delays the computation of a result and so this rule was added to help avoid this performance trap. Later, [V8 changed the way](https://v8.dev/blog/fast-async) `return await` worked so it would create a single microtask, which means this rule is no longer necessary.
+
+***
+
+This rule warns on any usage of `return await`.
+
+Using `return await` inside an `async function` keeps the current function in the call stack until the Promise that is being awaited has resolved. `return await` can also be used in a try/catch statement to catch errors from another function that returns a Promise.
 
 You can avoid the extra microtask by not awaiting the return value, with the trade off of the function no longer being a part of the stack trace if an error is thrown asynchronously from the Promise being returned. This can make debugging more difficult.
```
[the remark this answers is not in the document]
mikedidomizio: Yeah, I do feel like maybe we just leave the inaccuracies.

- I think if you end up rewriting it to be accurate, you might end up with a very confusing doc.
- I think also because the rule is deprecated, and maybe will be removed in the future, that rewriting the doc is a poor use of time.
- I think to keep within the scope of the intention of this PR, of just warning people up front, what a rewrite consists of would warrant a much bigger discussion.

I think preserving the old, and calling it a day is what I'm leaning towards 👍.

If everyone is in agreement, my actions would be:

- Revert the previous commit https://github.com/eslint/eslint/commit/d71bf2e7f725c62d2d29cda35555e79e99671d77
- Add `The following documentation is kept for historical context, and is not accurate anymore.` before the horizontal line.

At that point, I think we're in a good place to wrap this up.

How does that sound @nzakas?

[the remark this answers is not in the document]
nzakas: @mikedidomizio I disagree. I think it's easy to miss that one line saying the following isn't accurate. If you'd prefer not to continue working on the PR, that's okay. I can take it over.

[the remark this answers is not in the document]
mikedidomizio: I'd hate to push work onto anyone, but you probably know what you want and how it'll be structured, to avoid the back and forth, it's probably best that you take over for the time being. 

I'm fine with you pushing commits onto this PR, or just sharing your proposed changes first for feedback. I'd love to see what you're thinking in how it'll look.

nzakas:
Added a commit. Let me know what you think.

mikedidomizio:
> Added a commit. Let me know what you think.

@nzakas I like the changes. I like how you've reworded what the rule aims to prevent to now simply just what it does, and also summarized it at the end with "You should not use this rule...".  

I think the page is concise and clear enough now that anyone stumbling on it should get the message to not use it.  The only thing better would be when/if it eventually all gets removed.

LGTM 👍

kirkwaiblinger (approved):

snitin315 (dismissed):
LGTM, thanks! Leaving it open for others to review

docs/src/rules/no-return-await.md
```diff
@@ -7,15 +7,16 @@ further_reading:
...
+* `return await` on a promise will not result in an extra microtask.
+* `return await` yields a better stack trace for debugging.
 
-You can avoid the extra microtask by not awaiting the return value, with the trade off of the function no longer being a part of the stack trace if an error is thrown asynchronously from the Promise being returned. This can make debugging more difficult.
+Historical context: When promises were first introduced, calling `return await` introduced an additional microtask, one for the `await` and one for the return value of the async function. Each extra microtask delays the computation of a result and so this rule was added to help avoid this performance trap. Later, [V8 changed the way](https://v8.dev/blog/fast-async) `return await` worked so it would create a single microtask, which means this rule is no longer necessary.
```
mdjermanovic, line 15: "V8 changed the way" might imply that in other engines there can still be an extra microtask, depending on the implementation. This wasn't just an arbitrary decision or optimization in V8, but a change in the ECMAScript specification (https://github.com/tc39/ecma262/pull/1250), so I think we should keep "JavaScript now handles native Promises differently" or something like that.

docs/src/rules/no-return-await.md
```diff
@@ -7,15 +7,16 @@ further_reading:
...
 
 ## Rule Details
 
-This rule aims to prevent a likely common performance hazard due to a lack of understanding of the semantics of `async function`.
+This rule warns on any usage of `return await`.
```
mdjermanovic, line 19:
```suggestion
This rule warns on any usage of `return await` except in `try` blocks.
```

docs/src/rules/no-return-await.md
nzakas, line 15:
```suggestion
Historical context: When promises were first introduced, calling `return await` introduced an additional microtask, one for the `await` and one for the return value of the async function. Each extra microtask delays the computation of a result and so this rule was added to help avoid this performance trap. Later, [ECMA-262 changed the way](https://github.com/tc39/ecma262/pull/1250) `return await` worked so it would create a single microtask, which means this rule is no longer necessary.
```

PR #19812, opened 2025-06-02
feat: support TypeScript syntax in `no-unused-vars`

Tanujkanti4441:
#### Prerequisites checklist

[ ] Documentation update
[ ] Add autofix to a rule
[ ] Add a CLI option
[ ] Add something to the core
[ ] Other, please explain:

Added typescript syntax support in `no-unused-vars` rule.

Refs #19173

Moving forward https://github.com/eslint/eslint/pull/19580


lib/rules/no-unused-vars.js
```diff
@@ -39,6 +43,37 @@ const astUtils = require("./utils/ast-utils");
...
 //------------------------------------------------------------------------------
 
+/**
+ * Checks if the current file is a TypeScript definition file
+ * @param {string|undefined} filename The filename to check
+ * @returns {boolean} `true` if the file is a .d.ts file
+ * @private
+ */
+function isDefinitionFile(filename) {
+	return filename?.endsWith(".d.ts") ?? false;
```
lumirlumir: Can we add a logic to the `isDefinitionFile` helper function to also check for `.d.mts` and `.d.cts` files?

Here is a reference for it:

https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-7.html#new-file-extensions

Image

lib/rules/no-unused-vars.js
```diff
@@ -104,6 +139,8 @@ module.exports = {
 			usedIgnoredVar:
 				"'{{varName}}' is marked as ignored but is used{{additional}}.",
 			removeVar: "Remove unused variable '{{varName}}'.",
+			usedOnlyAsType:
+				"'{{varName}}' is {{action}} but only used as a type{{additional}}.",
 		},
 	},
```
lumirlumir, line 170:
```suggestion
                },
                
                dialects: ["typescript", "javascript"],
		language: "javascript",
        },
```

Could you add `dialects` and `language` fields to the `meta` property? As far as I know, the newly updated rules that support TypeScript use these fields.

https://github.com/eslint/eslint/blob/28cc7abbb72b29b1cac6fc4253646a7839586064/lib/rules/no-var.js#L202-L203

lib/rules/no-unused-vars.js
```diff
@@ -39,6 +43,37 @@ const astUtils = require("./utils/ast-utils");
...
+	return filename?.endsWith(".d.ts") ?? false;
+}
+
+/**
+ * Checks if a reference is a type-only usage (used in TypeScript type position)
+ * @param {Object} node The node to check
+ * @returns {boolean} `true` if the identifier is only used in a type position
+ * @private
+ */
+function isTypeReference(node) {
```
lumirlumir: Can we move the `isDefinitionFile` and `isTypeReference` helper functions under the `Helpers` comment?

Here is a reference for it:

https://github.com/eslint/eslint/blob/28cc7abbb72b29b1cac6fc4253646a7839586064/lib/rules/accessor-pairs.js#L31-L60

lumirlumir (changes requested):
Nice work! I’ve left a few comments about some missing parts 😄

lib/rules/no-unused-vars.js

lib/rules/no-unused-vars.js
```diff
@@ -1633,14 +1854,349 @@ module.exports = {
...
+		}
+
+		/**
+		 * Throws an error if the provided value is null or undefined.
+		 * @param {ASTNode} value The parent node to check for null or undefined.
+		 * @param {string} message The error message to use if the assertion fails.
+		 * @throws {Error} If `value` is null or undefined.
+		 * @returns {ASTNode} The provided node, if it is not null or undefined.
+		 */
+		function nullThrows(value, message) {
```
JoshuaKGoldberg: This is an interesting piece to copy over. The typescript-eslint project style is to very aggressively handle edge cases exposed by types. But it does sometimes go overboard and seem to require extra logic in places that don't need it (https://github.com/typescript-eslint/typescript-eslint/issues/10077).

Some of the places that `nullThrows` are also unnecessary. For example, the `node.parent` passed to `nullThrows` in the `Program` selector will be unnecessary once https://github.com/typescript-eslint/typescript-eslint/issues/11334 is resolved.

The ESLint project doesn't generally go that extra step. So, raising for other reviewers: should we just omit `nullThrows` & `NullThrowsReasons` in this PR?

lib/rules/no-unused-vars.js
```diff
@@ -1463,6 +1669,21 @@ module.exports = {
 						parent.range[1],
 					]);
 				}
+
+				if (
+					parent.parent === "FunctionExpression" &&
+					parent.parent.decarators &&
+					parent.parent.decarators.length > 0
```
JoshuaKGoldberg: [Bug] `decorators`, not `decarators`. The fact that this isn't failing any tests indicates a lack of coverage.

lib/rules/no-unused-vars.js
```diff
@@ -334,16 +344,48 @@ module.exports = {
...
+		 * @param {string|undefined} filename The filename to check
+		 * @returns {boolean} `true` if the file is a .d.ts file
+		 * @private
+		 */
+		function isDefinitionFile(filename) {
+			// Matches .d.ts, .d.mts, .d.cts at the end of the filename
+			return typeof filename === "string"
+				? /\.d\.(ts|mts|cts)$/u.test(filename)
+				: false;
+		}
```
JoshuaKGoldberg: [Refactor] This is very similar to the `isDeclareInDTSFile` logic in `no-shadow.js`, added in #19565. Which also had a review comment about matching all definition file extensions: https://github.com/eslint/eslint/pull/19565#discussion_r2010458934

I think that that's a sign there needs to be a shared utility function.

lib/rules/no-unused-vars.js
```diff
@@ -412,7 +453,10 @@ module.exports = {
 			let scope = ref.from;
 
 			while (scope) {
-				if (nodes.includes(scope.block)) {
+				if (
+					(Array.isArray(nodes) && nodes.includes(scope.block)) ||
+					(nodes instanceof Set && nodes.has(scope.block))
```
JoshuaKGoldberg, line 519: [Docs] The JSDoc for this function still says `nodes` can only ever be an array.

lib/rules/no-unused-vars.js
```diff
@@ -755,7 +926,7 @@ module.exports = {
 
 					// skip variables marked with markVariableAsUsed()
 					if (
-						!config.reportUsedIgnorePattern &&
+						// !config.reportUsedIgnorePattern &&
```
JoshuaKGoldberg: ?

tests/lib/rules/no-unused-vars.js
```diff
@@ -4611,3 +4611,3312 @@ try {
...
+					line: 2,
+					messageId: "usedOnlyAsType",
+				},
+			],
+		},
+		{
+			code: `
+ declare const foo: number;
+
+export type Foo = typeof foo;
```
JoshuaKGoldberg, line 7611: [Bug] `foo` is reported as unused in the tseslint rule, and I believe that's the right behavior. Reporting on values only used as types has been a bit ... controversial in the tseslint rule. Here's a summary of where the discussion has gone:

1. https://github.com/typescript-eslint/typescript-eslint/issues/3325 was a long-standing feature request to report on them
2. https://github.com/typescript-eslint/typescript-eslint/pull/9330 added that in about a year ago
3. https://github.com/typescript-eslint/typescript-eslint/issues/10266 asked for an option to allow them again, but most-to-all of the provided cases were non-idiomatic
    * The thread was long and got heated, so a summary is here: https://github.com/typescript-eslint/typescript-eslint/issues/10266#issuecomment-2581080684
4. https://github.com/typescript-eslint/typescript-eslint/pull/10690: was merged instead to add https://typescript-eslint.io/rules/no-unused-vars/#why-does-this-rule-report-variables-used-only-for-types

tests/lib/rules/no-unused-vars.js
```diff
@@ -4611,3 +4611,3312 @@ try {
...
+							messageId: "removeVar",
+							data: { varName: "Foo" },
+						},
+					],
+				},
+			],
+		},
+		{
+			filename: "foo.d.ts",
+			code: "class Foo {}\ndeclare class Bar {}\nexport {};",
```
JoshuaKGoldberg: [Question] Is there a reason this uses a single line string instead of a multi-line template string? The original tseslint tests use the latter.

JoshuaKGoldberg (changes requested):
Very exciting!

lib/rules/no-unused-vars.js
Tanujkanti4441: Forgot to remove this one, but with this line following test is falling:
```js
{
   code: `
     export enum Foo {
  		_A,
     }
    `,
   options: [
	{ reportUsedIgnorePattern: true, varsIgnorePattern: "_" },
    ],
},
```

And without this line, I think `reportUsedIgnorePattern: true` won't report the markedAsUsedVariables with ignored pattern when they are used, so should we report them in this condition?

tests/lib/rules/no-unused-vars.js
Tanujkanti4441: Thanks for the brief, just want to confirm this once, you mean this code should be reported as unused with message 

`'foo' is defined but never used.`

Instead of `'foo' is defined but only used as a type.` which is current behavior?

tests/lib/rules/no-unused-vars.js
Tanujkanti4441: Actually, while writing suggestions, the column number is not matching properly for me, so I tried this. Please suggest if I can do something better here?

lib/rules/no-unused-vars.js
Tanujkanti4441: Done!

tests/lib/rules/no-unused-vars.js
JoshuaKGoldberg: I don't follow, what do you mean by the column number not matching properly?

tests/lib/rules/no-unused-vars.js
JoshuaKGoldberg: Oh! I'm sorry, somehow I thought this was listed in `valid`. But it's clearly not. Ignore me here 😅

tests/lib/rules/no-unused-vars.js
Tanujkanti4441: Like here `column: 10` must be same for both code and output but somehow I was getting error on multiline code.

```js
{
	code: "import { Nullable } from 'nullable';\nconst a: string = 'hello';\nconsole.log(a);",
	errors: [
		{
			column: 10,
			data: {
				action: "defined",
				additional: "",
				varName: "Nullable",
			},
			line: 1,
			messageId: "unusedVar",
			suggestions: [
				{
					output: "\nconst a: string = 'hello';\nconsole.log(a);",
					messageId: "removeVar",
					data: { varName: "Nullable" },
				},
			   ],
		  },
	],
},
```

JoshuaKGoldberg: It passes on my machine. Maybe there's something with indentation that's misaligned?

Reference commit: https://github.com/eslint/eslint/commit/59ba196b703e38d67ed06f6d886047f25027859e

Tanujkanti4441: Done!

package.json
```diff
@@ -117,6 +117,7 @@
     "@humanwhocodes/retry": "^0.4.2",
     "@types/estree": "^1.0.6",
     "@types/json-schema": "^7.0.15",
+    "@typescript-eslint/scope-manager": "^8.34.1",
```
nzakas: I think this was added accidentally?

nzakas (commented):
This adds over 500 lines to the current rule, and I'm not sure this is the right place to add all of that logic. I think ideally, reference tracking would end up in eslint-scope and not be implemented inside of a single rule.

I'd like to propose we hold off on this and think about whether or not we want to add type reference tracking into eslint-scope. @eslint/eslint-tsc

mdjermanovic:
> This adds over 500 lines to the current rule, and I'm not sure this is the right place to add all of that logic. I think ideally, reference tracking would end up in eslint-scope and not be implemented inside of a single rule.
> 
> I'd like to propose we hold off on this and think about whether or not we want to add type reference tracking into eslint-scope.

Doesn't `@typescript-eslint/scope-manager` already track type references? So, when this rule is used with `@typescript-eslint/parser`, which uses `@typescript-eslint/scope-manager` as scope manager, type references should already be available to this rule in scopes?

nzakas:
@mdjermanovic Perhaps? 🤔 Do we want to assume that `@typescript-eslint/scope-manager` is always being used when we see TypeScript syntax? That seems like a shaky assumption moving forward.

mdjermanovic:
Based on [this playground example](https://typescript-eslint.io/play/#ts=5.8.2&showAST=scope&fileType=.tsx&code=LAKA9GAEBqCGBOBLWAjANgU0gMQPa8kQGdCA7SAWVwBMBXTAZQGNcAHLACgHM1cVY0zNhgB0TABaI01IeyIBtAAwBdAJQjQAFwCe7HPkgBeSKVoBbFBngBuUKAiRNkkvAwAzKxlJMsmgngIMAA9iTRJEcio6RhZ2ABpIWFJqSABiVw9XbwwSXDdHcSw4JFRMfVxQTE1IIIAuctsQIA&eslintrc=N4KABGBEBOCuA2BTAzpAXGYBfEWg&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false), it seems that `@typescript-eslint/scope-manager` does track type references.

I'd say it's reasonable to expect that a TypeScript-aware parser provides a TypeScript-aware scope manager.

nzakas:
I suppose --- I'm just still uncomfortable with the added 500 lines of code to this rule. Is that really needed?

mdjermanovic:
I haven't reviewed the changes in detail, but I see that `@typescript-eslint/scope-manager` is used directly in the rule: 

https://github.com/Tanujkanti4441/eslint/blob/421354008ddfbd023f3d9d038b0277dc19445af7/lib/rules/no-unused-vars.js#L13-L16

@Tanujkanti4441 if we would assume that, for TypeScript code, scopes have already been analyzed by `@typescript-eslint/scope-manager` or a compatible scope manager, would that, and how much, reduce the complexity and the number of lines added to this rule?

Tanujkanti4441:
> I haven't reviewed the changes in detail, but I see that `@typescript-eslint/scope-manager` is used directly in the rule:
> 
> https://github.com/Tanujkanti4441/eslint/blob/421354008ddfbd023f3d9d038b0277dc19445af7/lib/rules/no-unused-vars.js#L13-L16
> 
> @Tanujkanti4441 if we would assume that, for TypeScript code, scopes have already been analyzed by `@typescript-eslint/scope-manager` or a compatible scope manager, would that, and how much, reduce the complexity and the number of lines added to this rule?

@mdjermanovic, I mostly tried to replicate the code of typescript-eslint's `no-unused-vars` rule and using of `@typescript-eslint/scope-manager` is also a part of that process and in most of the code we are marking the `identifiers` in `ts` specific `nodes` as used, so can we assume that we may not simplify this further with the use of `@typescript-eslint/scope-manager`? Or am I missing any detail here?

package.json
Tanujkanti4441: No, it is an intentional change. It is used in the rule.

nzakas:
@Tanujkanti4441 the overall question is about the amount of code this adds to the rule. It seems like if variable reference tracking is already implemented in the typescript-eslint scope manager, that it shouldn't take 500 lines of code to track TypeScript identifiers.

package.json
JoshuaKGoldberg: I don't think this is a direction ESLint can take. The TSESLint scope manager is a pretty large dependency that pulls in a lot of TypeScript-specific concepts and code. For example, it has direct dependencies on `@typescript-eslint/types` and `@typescript-eslint/visitor-keys`.

lib/rules/no-unused-vars.js
```diff
@@ -671,6 +702,116 @@ module.exports = {
...
 
+		/**
+		 * Determine whether a variable definition is a type import. e.g. "import type { Foo } from 'foo';".
+		 * @param {Object} definition The variable definition to check.
+		 * @returns {boolean} `true` if the definition is a type import, otherwise `false`.
+		 * @private
+		 */
+		function isTypeImport(definition) {
+			return (
+				definition?.type === DefinitionType.ImportBinding &&
```
JoshuaKGoldberg: To remove one of the dependencies on `@typescript-eslint/scope-manager`, this can be inlined to:


```suggestion
				definition?.type === 'ImportBinding' &&
```

This isn't completely idiomatic (tseslint enums are meant to be opaque) but it's better than a full runtime dependency on the package.

lib/rules/no-unused-vars.js
```diff
@@ -1633,14 +1841,349 @@ module.exports = {
...
+		/**
+		 * Visits all identifier nodes within a given pattern node and applies a callback to each.
+		 *
+		 * This function uses PatternVisitor to traverse complex patterns and invokes the provided callback for every identifier found within the pattern.
+		 * @param {ASTNode} node The pattern node to traverse.
+		 * @param {Function} cb The callback function to call for each identifier found in the pattern.
+		 * @returns {void}
+		 */
+		function visitPattern(node, cb) {
+			const visitor = new PatternVisitor({}, node, cb);
```
JoshuaKGoldberg: For removing this usage of the `@typescript-eslint/scope-manager` package: hmm! This was a fun deep dive. There's no ideal API I found, but maybe a good workaround?

As I understand it, the backing need here is to be able to parse out all the identifiers from the declaration, right? As in, supporting direct variable IDs, destructures, etc.? `PatternVisitor` 

From this rule's perspective, I think the "easiest" solution would be for `context.sourceCode.scopeManager` to provide some API like this. As in, perhaps it could make `PatternVisitor` or some wrapping API/function around it public (which it currently isn't, right?). That would probably be an issue -> RFC to file on ESLint/rewrite, but given how narrow the use case, I'm not sold that it would make sense.

To keep logic within this rule, I think it would make sense to instead write a small pattern visitor function. It just needs to handle the known kinds of syntax.

```ts
function collectPatternIdentifiers(node, identifiers) {
	switch (node.type) {
		case "ArrayPattern":
			for (const element of node.elements) {
				if (element) {
					collectPatternIdentifiers(element, identifiers);
				}
			}
			break;

		case "AssignmentPattern":
			collectPatternIdentifiers(node.left, identifiers);
			break;

		case "Identifier":
			identifiers.push(node);
			break;

		case "ObjectPattern":
			for (const property of node.properties) {
				collectPatternIdentifiers(
					property.type === "Property" ? property.value : property.argument,
					identifiers,
				);
			}
			break;

		case "RestElement":
			collectPatternIdentifiers(node.argument, identifiers);
			break;

		case "TSAsExpression":
		case "TSNonNullExpression":
		case "TSSatisfiesExpression":
			collectPatternIdentifiers(node.expression, identifiers);
			break;
	}
}
```

I got all unit tests to pass locally by swapping this in.

JoshuaKGoldberg:
@Tanujkanti4441 is this something you still have time & interest for? I poked around and think I have a working solution to remove the `@typescript-eslint/scope-manager` dependency.

Tanujkanti4441:
> @Tanujkanti4441 is this something you still have time & interest for? I poked around and think I have a working solution to remove the `@typescript-eslint/scope-manager` dependency.

Hey, due to time constraints not able to dedicate the necessary attention to this PR right now, so feel free to move this forward. Thanks.

JoshuaKGoldberg:
OK, I applied the suggested changes as much as I could. The main things I did were:

* (most important) Replace the usage of `@typescript-eslint/scope-manager` with a `collectPatternIdentifiers` function per https://github.com/eslint/eslint/pull/19812#discussion_r2637143164
* Moved `isDefinitionFile` from `ast-utils.js` to a new `file-utils.js` file, as it's not really an AST util

I didn't work on the suggestions around https://github.com/eslint/eslint/pull/19812#issuecomment-3144193010 for re-using analyses from the provided scope-manager. I don't know what APIs are exported that would be usable. So, I suppose this is still waiting for input from @eslint/eslint-tsc.

nzakas:
@JoshuaKGoldberg FYI, what you want is "tsc agenda", not "tsc waiting" to make sure it gets on the meeting agenda. https://eslint.org/docs/latest/maintain/manage-issues#when-to-send-to-tsc

sam3k:
The TSC agreed to review the code and run performance testing to confirm the extra code doesn't cause a significant slowdown before deciding whether to add TypeScript support to `no-unused-vars`.

PR #19831, opened 2025-06-08
feat: throw error when column is negative in `getIndexFromLoc`

lumirlumir:
#### Prerequisites checklist

[ ] Documentation update
[ ] Add autofix to a rule
[ ] Add a CLI option
[x] Add something to the core
[ ] Other, please explain:

Hello,

In this PR, I’ve fixed the handling of negative column indices in `getIndexFromLoc` and added corresponding tests.

The JavaScript `SourceCode` class is based on 0-indexed columns, so passing a negative column index should throw an error — but currently, it doesn’t.

This could lead to unintended side effects or bugs, so I’ve addressed the issue accordingly.


mdjermanovic:
I think this change is fine for a semver-minor release. @eslint/eslint-tsc thoughts?

lib/languages/js/source-code/source-code.js
```diff
@@ -832,6 +832,7 @@ class SourceCode extends TokenStore {
 		 * use getIndexFromLoc consistently without worrying about edge cases at the end of a file.
 		 */
 		if (
+			loc.column < 0 ||
```
mdjermanovic: The error message might be misleading because the line length is not relevant for this check. Perhaps we could check for the negative column separately, e.g., after we check the loc.line (lines 807-811 above) and throw an error with a more specific message?

lumirlumir: Actually, I found this edge case while implementing https://github.com/eslint/rewrite/pull/212.

In that PR, I used a more detailed error message like below:

image

So, if it’s okay, can I update the error messages more robustly, as I did in https://github.com/eslint/rewrite/pull/212?

nzakas: I'd prefer keeping error messages shorter. The longer the message, the less likely people will read it.

In this case, I'd say let's check if the column is less than zero and throw an error saying, "Invalid column number ${number}." I think we can assume people understand that columns must be positive numbers.

nzakas:
I think it's fine for semver-minor. Using -1 now would cause a difficult-to-track-down error right now, so throwing a more useful one is definitely preferable.

fasttime:
> I think this change is fine for a semver-minor release.

I agree. It's possible that this change could break some plugins, but since the current behavior for negative columns in locations was not intended or documented, this can be a fix.

lumirlumir:
Thanks for the suggestions. I’ve added new commits and marked this PR as accepted.

nzakas:
@mdjermanovic just a reminder: don't forget to move PRs into the "Second Review Needed" column on the Triage board when waiting for a second review.

fasttime (approved):
LGTM, thanks!

